PR #259, opened 2021-08-02
Kafka protocol and Cloudevents support

axsaucedo:
Exploration of kafka support and cloud events support


docs/examples/kafka-sklearn/README.md
```diff
@@ -0,0 +1,248 @@
...
+     'outputs': [{'name': 'predict',
+       'shape': [1],
+       'datatype': 'INT64',
+       'parameters': None,
+       'data': [8]}]}
+
+
+
+
+```python
```
adriangonz, line 203: Should we add a few words here about how we are sending / returning payloads to / from Kafka?

mlserver/cloudevents.py
```diff
@@ -0,0 +1,61 @@
...
+CLOUDEVENTS_HEADER_MODEL_ID = "Ce-Modelid"
+CLOUDEVENTS_HEADER_INFERENCE_SERVICE = "Ce-Inferenceservicename"
+CLOUDEVENTS_HEADER_NAMESPACE = "Ce-Namespace"
+CLOUDEVENTS_HEADER_ENDPOINT = "Ce-Endpoint"
+
+ENV_SDEP_NAME = "SELDON_DEPLOYMENT_ID"
+ENV_PREDICTOR_NAME = "PREDICTOR_ID"
+ENV_MODEL_NAME = "PREDICTIVE_UNIT_ID"
+
+NOT_IMPLEMENTED_STR = "NOTIMPLEMENTED"
```
adriangonz, line 20: Would it make sense to just use `None` instead of the `"NOTIMPLEMENTED"` string?

mlserver/cloudevents.py
```diff
@@ -0,0 +1,61 @@
...
+CLOUDEVENTS_HEADER_SPECVERSION_DEFAULT = "0.3"
+
+CLOUDEVENTS_HEADER_REQUEST_ID = "Ce-Requestid"
+CLOUDEVENTS_HEADER_MODEL_ID = "Ce-Modelid"
+CLOUDEVENTS_HEADER_INFERENCE_SERVICE = "Ce-Inferenceservicename"
+CLOUDEVENTS_HEADER_NAMESPACE = "Ce-Namespace"
+CLOUDEVENTS_HEADER_ENDPOINT = "Ce-Endpoint"
+
+ENV_SDEP_NAME = "SELDON_DEPLOYMENT_ID"
+ENV_PREDICTOR_NAME = "PREDICTOR_ID"
```
adriangonz, line 17: This may fall out of the scope of this PR, but wouldn't these variables fail with multi-model servers running multiple models? They would also be specific to Seldon Core, would it be possible to make them agnostic from the underlying orchestrator?

mlserver/kafka/server.py
```diff
@@ -0,0 +1,110 @@
...
+        # TODO: Implement
+        pass
+
+    async def start(self):
+        self._create_server()
+
+        # TODO: Move logic to KafkaRouter (Also to implement cutom handlers modularly)
+        await self._consumer.start()
+        await self._producer.start()
+        # TODO: Catch or ensure parent function catches and stops server
```
adriangonz, line 49: This is an interesting point. What's the recommended way to deal with errors with Kafka? In REST / gRPC is easy as they're sync protocols, but is there any standard practice for Kafka?

mlserver/utils.py
```diff
@@ -1,11 +1,16 @@
...
+import logging
 
 from typing import List
 
 from .settings import ModelSettings
 from .errors import InvalidModelURI
 
 
+# In order for logging to be captured we need uvicorn logger
+logger = logging.getLogger("uvicorn.error")
```
adriangonz, line 11: Interesting! Do you know why is this the case? Are other loggers getting hidden?

adriangonz (commented):
Nice one! I've added a few comments. Besides those, should we add some tests to ensure that the cloudevent headers get added correctly? I'm thinking that these could be added on the `tests/rest/test_endpoints.py` file, which already has a few test examples interacting with the REST server.

mlserver/cloudevents.py
```diff
@@ -0,0 +1,106 @@
+from enum import Enum
+from typing import Dict, Optional, Union
+
+from .settings import Settings, ModelSettings
+from .types import InferenceRequest, InferenceResponse, Parameters
+from .middleware import InferenceMiddleware
+
+CLOUDEVENTS_HEADER_ID = "Ce-Id"
```
axsaucedo, line 8: May be worth validating if the headers are still indeed correct (as I know SCV2 may be using different ones)

axsaucedo (commented):
Looks great! Really cool as well to see the unit tests with full end to end kafka, will make it much easier to test any change - LGTM

Btw I can't seem to approve the PR as I opened it, so may actually require you to approve it manually 


image

mlserver/cloudevents.py
```diff
@@ -0,0 +1,106 @@
...
+    Request = "io.seldon.serving.inference.request"
+    Response = "io.seldon.serving.inference.response"
+
+
+def get_namespace() -> Optional[str]:
+    try:
+        # Namespace can be fetched from loaded file vars from k8s 1.15.3+
+        with open("/var/run/secrets/kubernetes.io/serviceaccount/namespace", "r") as f:
+            return f.read()
+    except Exception:
```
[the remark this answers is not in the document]
adriangonz: Great point! I'll change that :+1:

PR #268, opened 2021-08-06
Tweaks to get things working properly if environment.tar.gz is provided

sakoush:
If we found a tarball  (`environment.tar.gz`) for the conda environment used during model training, we need to install `mlserver` and deps after unpacking.

We also pin the version to be installed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,8 @@ SHELL ["/bin/bash", "-c"]
 
 ENV MLSERVER_MODELS_DIR=/mnt/models \
     MLSERVER_ENV_TARBALL=/mnt/models/environment.tar.gz \
-    PATH=/home/default/.local/bin:$PATH
+    PATH=/home/default/.local/bin:$PATH \
+    MLSERVER_VERSION=0.4.0
 
 RUN apt-get update && \
     apt-get -y --no-install-recommends install \
@@ -44,5 +45,5 @@ RUN useradd -u 1000 -s /bin/bash mlserver && \
 USER 1000
 
 # Need to source `activate-env.sh` so that env changes get persisted
-CMD . ./hack/activate-env.sh $MLSERVER_ENV_TARBALL \
-    && mlserver start $MLSERVER_MODELS_DIR
+CMD . ./hack/activate-env.sh $MLSERVER_ENV_TARBALL $MLSERVER_VERSION \
+    && python -m mlserver.cli.main start $MLSERVER_MODELS_DIR
--- a/hack/activate-env.sh
+++ b/hack/activate-env.sh
@@ -5,10 +5,10 @@ set -o errexit
 set -o pipefail
 
 _printUsage() {
-  echo "Usage: ./run-in-env.sh <envTarball>"
+  echo "Usage: ./run-in-env.sh <envTarball> <version>"
 }
 
-if [ "$#" -ne 1 ]; then
+if [ "$#" -ne 2 ]; then
   echo 'Invalid number of arguments'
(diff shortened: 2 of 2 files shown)


Dockerfile
```diff
@@ -4,7 +4,8 @@ SHELL ["/bin/bash", "-c"]
 
 ENV MLSERVER_MODELS_DIR=/mnt/models \
     MLSERVER_ENV_TARBALL=/mnt/models/environment.tar.gz \
-    PATH=/home/default/.local/bin:$PATH
+    PATH=/home/default/.local/bin:$PATH \
+    MLSERVER_VERSION=0.4.0
```
sakoush, line 8: not sure if there is a better way of defining the version to be used. i.e. to fetch it from where the version is actually defined?

Dockerfile
```diff
@@ -44,5 +45,5 @@ RUN useradd -u 1000 -s /bin/bash mlserver && \
 USER 1000
 
 # Need to source `activate-env.sh` so that env changes get persisted
-CMD . ./hack/activate-env.sh $MLSERVER_ENV_TARBALL \
-    && mlserver start $MLSERVER_MODELS_DIR
+CMD . ./hack/activate-env.sh $MLSERVER_ENV_TARBALL $MLSERVER_VERSION \
+    && python -m mlserver.cli.main start $MLSERVER_MODELS_DIR
```
sakoush, line 49: this works but I am not sure if there are any caveats?

hack/activate-env.sh
```diff
@@ -34,10 +34,15 @@ _unpackEnv() {
   echo "--> Disabling user-installed packages..."
   # https://github.com/conda/conda/issues/448#issuecomment-195848539
   export PYTHONNOUSERSITE=True
+
+  echo "--> Installing mlserver..."
+  pip install mlserver==$_version
+  pip install mlserver-mlflow==$_version
```
sakoush, line 40: I assume that `mlserver-mlflow` might be at a different version in the future? for now they look to be released together with `mlserver`?

Dockerfile
adriangonz: Good point. I'd install the version that it's copied locally to the Docker image rather than fetching one from PyPi. That way we ensure that the Docker image reflects whatever is on the repo.

Dockerfile
adriangonz: Did you see any differences in using `python -m mlserver.cli.main` versus just doing `mlserver start`?

hack/activate-env.sh
adriangonz: Rather than just `mlserver-mlflow`, should we re-install every runtime? I guess this is where it starts to get confusing, but from the `activate-env.sh` script point of view, there isn't anything special about the MLflow runtime vs the others.

adriangonz: Put another way, users could fall into the same issue with SKLearn models, XGBoost, etc. It should be possible to leverage the same workaround for those cases.

adriangonz (commented):
Nice one! What would be the different scenarios where the user's environment could clash with the base image? E.g. what would happen if they specify a different version of `numpy` which clashes with what gets installed with `pip install mlserver`?

hack/activate-env.sh
sakoush: Yes I was thinking about this as well. we probably should install all deps if we go down this path.

sakoush:
not required as this stage, we decided that we will pack `mlserver` and `mlserver-mlfow` in the conda packed enviornment.

PR #271, opened 2021-08-16
MLServer Runtime for Alibi Detect Models

SachinVarghese:
This PR creates an alibi-detect runtime for MLserver.


runtimes/alibi-detect/mlserver_alibi_detect/cd/runtime.py
```diff
@@ -0,0 +1,42 @@
...
+class AlibiDriftDetectRuntime(AlibiDetectRuntime):
+    """
+    Implementation of the MLModel interface to load and serve Alibi drift models.
+    """
+
+    async def load(self) -> bool:
+
+        model_uri = await get_model_uri(self._settings)
+
+        if not bool(self.alibi_detect_settings.init_detector):
```
adriangonz, line 21: This one should always be a boolean now, right? In that case, we could get rid of the explicit `bool()` casting.

runtimes/alibi-detect/mlserver_alibi_detect/cd/runtime.py
```diff
@@ -0,0 +1,42 @@
...
+            self._model = load_detector(model_uri)
+        else:
+            detector_data = pickle.load(open(model_uri, "rb"))
+            try:
+                detector_type = self.alibi_detect_settings.detector_type
+
+                detect_module = import_module(ALIBI_DETECT_DRIFT_MODULE)
```
adriangonz, line 28: Instead of doing the import manually, we could let Pydantic handle that by changing the `detector_type` field type to [`PyObject`](https://pydantic-docs.helpmanual.io/usage/types/#pydantic-types). 

We leverage this already for the `implementation` field:

https://github.com/SeldonIO/MLServer/blob/f088620d67cd366ad19857db38d5f89c7586a427/mlserver/settings.py#L65

runtimes/alibi-detect/mlserver_alibi_detect/cd/runtime.py
```diff
@@ -0,0 +1,42 @@
...
+                drift_detector = getattr(detect_module, detector_type)
+            except Exception as e:
+                raise InvalidAlibiDetector(detector_type, self._settings.name, e)
```
adriangonz, line 31: Instead of passing the underlying exception manually, we could "chain" the exceptions `e` and `InvalidAlibiDetector()` as:

```
raise InvalidAlibiDetector(detector_type, self._settings.name) from e
```

That way we preserve all the information (e.g. stacktrace, etc.). First time I hear about this BTW... full credit goes to this [SO answer](https://stackoverflow.com/a/792163/5015573).

runtimes/alibi-detect/mlserver_alibi_detect/protocols/seldon_http.py
```diff
@@ -0,0 +1,64 @@
...
+        return SeldonPayload.TENSOR
+    elif "ndarray" in data_def:
+        return SeldonPayload.NDARRAY
+    elif "tftensor" in data_def:
+        return SeldonPayload.TFTENSOR
+    else:
+        raise Exception("Unknown Seldon payload %s" % data_def)
+
+
+class SeldonRequestHandler(RequestHandler):
```
adriangonz, line 45: Probably out of scope of this PR, but do you think we could use `Codecs` to implement this logic? The main benefit would be that we can then re-use this for other runtimes (e.g. to provide v1 support to the `mlserver-sklearn` runtime).

runtimes/alibi-detect/mlserver_alibi_detect/protocols/v2.py
```diff
@@ -0,0 +1,62 @@
...
+        npty = np.float64
+    else:
+        raise InferenceError(f"V2 unknown type or type that can't be coerced {ty}")
+
+    arr = np.array(data, dtype=npty)
+    arr.shape = tuple(shape)
+    return arr
+
+
+class KFservingV2RequestHandler(RequestHandler):
```
adriangonz, line 40: For this one, we could re-use the existing `NumpyCodec` that we have in the `codecs.numpy` module:

https://github.com/SeldonIO/MLServer/blob/f088620d67cd366ad19857db38d5f89c7586a427/mlserver/codecs/numpy.py#L42-L64

It should be just a question of calling `NumpyCodec().decode(request.inputs[0])`, and that should take care of unwrapping the request into an array.

runtimes/alibi-detect/mlserver_alibi_detect/runtime.py
```diff
@@ -0,0 +1,102 @@
...
+
+# ENV_PREFIX_ALIBI_DETECT_SETTINGS = "MLSERVER_MODEL_ALIBI_DETECT_"
+
+
+class AlibiDetectSettings(dict):
+    """
+    Parameters that apply only to alibi detect models
+    """
+
+    def __init__(self, dictionary):
```
adriangonz, line 27: Isn't this one the default behaviour for Pydantic models?

runtimes/alibi-detect/mlserver_alibi_detect/runtime.py
```diff
@@ -0,0 +1,102 @@
...
+            raise InferenceError("Unrecognized request format: %s" % e)
+
+        request_handler = get_request_handler(
+            Protocol(self.alibi_detect_settings.protocol), body
+        )
+        request_handler.validate()
+        input_data = request_handler.extract_request()
+
+        y = await self.predict_fn(input_data, {})
+        output_data = json.dumps(y, cls=NumpyEncoder)
```
adriangonz, line 71: Instead of the default `json` encoder here, we could leverage `orjson` which should be slightly faster. The interface should be similar to the `json` module, so hopefully it should be a drop-in replacement.

adriangonz (changes requested):
Nice one @SachinVarghese! Amazing work! 

I've added a few comments mainly pointing out to other parts of MLServer which we could re-use here, but it looks great!

runtimes/alibi-detect/mlserver_alibi_detect/protocols/seldon_http.py
SachinVarghese: Yeah, that could be a good idea like the numpy Codec and also probably useful for other projects also.

mlserver/types/dataplane.py
```diff
@@ -39,6 +39,7 @@ class Config:
         extra = Extra.allow
 
     content_type: Optional[str] = None
+    predict_parameters: Optional[dict] = {}
```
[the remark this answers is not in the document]
SachinVarghese: I think the reason was that I didn't want the content type to be used in the alibi detect predict function and only alibi specific params to be used in predict method where I just spread the params like `**predict_parameters`

runtimes/alibi-detect/mlserver_alibi_detect/cd/runtime.py
SachinVarghese: Yeah will update

runtimes/alibi-detect/mlserver_alibi_detect/cd/runtime.py
SachinVarghese: Sure will update

runtimes/alibi-detect/mlserver_alibi_detect/runtime.py
SachinVarghese: I was getting some issues with the `extra` dict being used in pydantic `expected str, bytes or os.PathLike object, not dict`

mlserver/types/dataplane.py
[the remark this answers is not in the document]
SachinVarghese: Have it at runtime level

adriangonz (approved):
Looks great! Let's wait for @arnaudvl 's thoughts, but I think we're good to merge this one.

docs/examples/alibi-detect/.gitignore
```diff
@@ -0,0 +1 @@
+alibi-detector-artifacts
```
michaelcheah, line 1: Maybe do this
```suggestion
alibi-detector-artifacts/*
```

and keep the folder so that whoever's running the notebook doesn't need to create the folder and can run it out of the box

runtimes/alibi-detect/mlserver_alibi_detect/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .cd.runtime import AlibiDriftDetectRuntime
+
+__all__ = ["AlibiDriftDetectRuntime"]
```
michaelcheah, line 3: Had some `ImportError`s while trying to install. I believe `__init__.py` files are required in both `.cd` and `.protocol` for them to be imported as submodules

runtimes/alibi-detect/setup.py
```diff
@@ -0,0 +1,40 @@
...
+from typing import Dict
+from setuptools import setup, find_packages
+
+ROOT_PATH = os.path.dirname(__file__)
+PKG_NAME = "mlserver-alibi-detect"
+PKG_PATH = os.path.join(ROOT_PATH, PKG_NAME.replace("-", "_"))
+
+
+def _load_version() -> str:
+    version = ""
```
michaelcheah, line 12: nit: unused

runtimes/alibi-detect/mlserver_alibi_detect/runtime.py
```diff
@@ -0,0 +1,106 @@
...
+
+        return types.InferenceResponse(
+            model_name=self.name,
+            model_version=self.version,
+            parameters=y["meta"],
+            outputs=[default_codec.encode(name="detect", payload=output_data)],
+        )
+
+    async def predict_fn(
+        self, input_data: Any, predictParameters: Optional[dict] = {}
```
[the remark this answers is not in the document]
michaelcheah: Also, the optional `predictParameters` don't seem to be used anyway

michaelcheah (commented):
Mainly code comments

Tested out the notebook. It looks good. The only thing I would point out is that the `p_val` used to demonstrate that the `init_parameters` can be changed does not demonstrate anything atm because none of the p values of the test data is below 0.05.

runtimes/alibi-detect/LICENSE
```diff
@@ -0,0 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "{}"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright 2020 Seldon Technologies Ltd.
```
axsaucedo, line 189: probably 2021

runtimes/alibi-detect/mlserver_alibi_detect/protocols/seldon_http.py
```diff
@@ -0,0 +1,64 @@
...
+from enum import Enum
+
+
+class SeldonPayload(Enum):
+    TENSOR = 1
+    NDARRAY = 2
+    TFTENSOR = 3
+
+
+def _extract_list(body: dict) -> np.array:
```
axsaucedo, line 13: I wonder if these seldon v1 protocol extraction utils can be made available upstraem, as the `if statement` extracting from data / ndarray /tfttensor is probably relevant in tempo, req logger, etc (definitely out of this PR but perhaps could be part of the exploration of the Handler classes

axsaucedo (commented):
Looks good, it seems most of the comments were about how to best address the handlers, which is something that we can address as a followup. Only thing I think would be good is to add documentation that is linked from the main docs

docs/examples/alibi-detect/.gitignore
jklaise: Might need to create the directory with an empty `.keep` file in it so git actually controls it if we want the folder to be available out-of-the-box.

runtimes/alibi-detect/mlserver_alibi_detect/cd/runtime.py
```diff
@@ -0,0 +1,33 @@
...
+
+        model_uri = await get_model_uri(self._settings)
+
+        if not self.alibi_detect_settings.init_detector:
+            self._model = load_detector(model_uri)
+        else:
+            try:
+                detector_data = pickle.load(open(model_uri, "rb"))
+                drift_detector = self.alibi_detect_settings.detector_type
+            except Exception as e:
```
jklaise, line 23: Should we catch such a generic exception? Is it just exceptions from `pickle` that we care about here?

runtimes/alibi-detect/mlserver_alibi_detect/protocols/seldon_http.py
SachinVarghese: Yeah good point. we also discussed this. I think a good idea that Adrian mentioned was to have them as codecs in MLserver that can be reused across runtimes and projects

runtimes/alibi-detect/mlserver_alibi_detect/protocols/seldon_http.py
```diff
@@ -0,0 +1,64 @@
...
+
+        ty = _get_request_ty(self.request)
+        if not (
+            ty == SeldonPayload.TENSOR
+            or ty == SeldonPayload.NDARRAY
+            or ty == SeldonPayload.TFTENSOR
+        ):
+            raise InferenceError(
+                "`data` key should contain either `tensor`,`ndarray`, or `tftensor`"
+            )
```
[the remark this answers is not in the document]
SachinVarghese: Good point, will update

runtimes/alibi-detect/mlserver_alibi_detect/cd/runtime.py
```diff
@@ -0,0 +1,33 @@
...
+                raise InvalidAlibiDetector(
+                    str(drift_detector), self._settings.name
```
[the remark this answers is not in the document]
SachinVarghese: Good point, will update

runtimes/alibi-detect/mlserver_alibi_detect/runtime.py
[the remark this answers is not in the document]
SachinVarghese: Yeah this one was for a feature to update detector params during predict-time. Not used in this PR though, so will remove.

docs/examples/alibi-detect/.gitignore
SachinVarghese: Yeah on this one I can add the empty folder but just checking since I think other runtimes artifacts are also in git @adriangonz ?

runtimes/alibi-detect/LICENSE
SachinVarghese: Fixed now

mlserver/settings.py
```diff
@@ -44,6 +44,7 @@ class Config:
     uri: Optional[str] = None
     version: Optional[str] = None
     format: Optional[str] = None
+    extra: Optional[dict] = {}
```
michaelcheah, line 47: I think I missed this earlier. I think this is also another case of this https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html

I'm not sure if pydantic allows this though?

SachinVarghese: Yeah that's true. I will look into this

michaelcheah: This didn't seem to cause any issues so I'm going to assume this is okay

image

michaelcheah (approved):

mlserver/settings.py
SachinVarghese: Would be to know if @adriangonz or @jklaise has better ideas on this.

jklaise: Yea, it appears that pydantic handles mutable defaults properly, alternatively one can use `default_factory`: https://stackoverflow.com/a/63808835

docs/examples/alibi-detect/.gitignore
adriangonz: I think that the artifacts for some examples are there actually (not for tests though).

runtimes/alibi-detect/mlserver_alibi_detect/cd/runtime.py
SachinVarghese: Yeah good point. I think I'll update it to (FileNotFoundError, EOFError, pickle.UnpicklingError) exceptions. I guess it covers the errors we are interested in here. Thoughts @jklaise ?

jklaise: Hmm guess that would cover the obvious cases, not an expert tho on catching errors when unpickling!

docs/examples/alibi-detect/.gitignore
SachinVarghese: okay added empty folder now

runtimes/alibi-detect/mlserver_alibi_detect/cd/runtime.py
SachinVarghese: Okay I have added these more specific exceptions now.

runtimes/alibi-detect/mlserver_alibi_detect/runtime.py
```diff
@@ -0,0 +1,104 @@
...
+
+    async def predict(self, payload: types.InferenceRequest) -> types.InferenceResponse:
+        payload = self._check_request(payload)
+        model_input = payload.inputs[0]
+        default_codec = NumpyCodec()
+        input_data = self.decode(model_input, default_codec=default_codec)
+        y = await self.predict_fn(input_data)
+
+        # TODO: Convert alibi-detect output to v2 protocol
+        output_data = np.array(y["data"]["is_drift"])
```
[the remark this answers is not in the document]
SachinVarghese: Yeah good point will update.

runtimes/alibi-detect/setup.py
```diff
@@ -0,0 +1,40 @@
...
+
+setup(
+    name=PKG_NAME,
+    version=_load_version(),
+    url="https://github.com/SeldonIO/MLServer.git",
+    author="Seldon Technologies Ltd.",
+    author_email="hello@seldon.io",
+    description="Alibi-Detect runtime for MLServer",
+    packages=find_packages(),
+    install_requires=["mlserver", "alibi-detect==0.7.2"],
```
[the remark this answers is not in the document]
SachinVarghese: That's a good qustion, I guess we dont have to pin it here and can work with latest just like other runtimes. We can map alibi versions to the mlserver image built later and that could be more flexible. Will update

runtimes/alibi-detect/mlserver_alibi_detect/runtime.py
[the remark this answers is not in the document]
SachinVarghese: Updated now.

runtimes/alibi-detect/mlserver_alibi_detect/runtime.py
```diff
@@ -0,0 +1,104 @@
...
+
+        return types.InferenceResponse(
+            model_name=self.name,
+            model_version=self.version,
+            parameters=y["meta"],
+            outputs=[default_codec.encode(name="detect", payload=output_data)],
```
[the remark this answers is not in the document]
SachinVarghese: Updated now.

runtimes/alibi-detect/mlserver_alibi_detect/cd/runtime.py
```diff
@@ -0,0 +1,39 @@
...
+from alibi_detect.utils.saving import load_detector
+import pickle
+
+
+class AlibiDriftDetectRuntime(AlibiDetectRuntime):
+    """
+    Implementation of the MLModel interface to load and serve Alibi drift models.
+    """
+
+    async def load(self) -> bool:
```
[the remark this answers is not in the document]
SachinVarghese: Really good point, will do this in a followup PR with more examples soon

Issue #162, opened 2021-04-12
Convert input types based on MLflow's model signature


adriangonz:
MLflow models provide a typing mechanism through their [model signatures](https://www.mlflow.org/docs/latest/models.html#model-signature). These types are used to validate the input payloads, therefore is important that the `mlserver-mlflow` runtime takes them into account when converting from the V2 Protocol to an MLflow-valid input (see #160 and #161).

There is currently a mismatch between the types supported by the [V2 Dataplane](https://github.com/kubeflow/kfserving/blob/master/docs/predict-api/v2/required_api.md#tensor-data-1) and [MLflow native types](https://www.mlflow.org/docs/latest/python_api/mlflow.types.html#mlflow.types.DataType). For example, 

- MLflow allows you to set an input column as a `string`, whereas `BYTES` is the closest type in the V2 Dataplane.
- MLflow allows you to set an input column as `binary`, which is expected to be a base64 string. This is different to the `BYTES` field in the V2 Dataplane (which doesn't specify explicitly how these field needs to be transmitted). 

This conversion will also need to take into account how to convert between the V2 "lower-level" types and MLflow's native types.

PR #285, opened 2021-08-25
Sync metadata with MLflow's model signature

### LINKED ISSUE #164: Expose MLflow's model signature as metadata
MLflow models can define a [model signature](https://www.mlflow.org/docs/latest/models.html#model-signature). The information contained in this model signature has a few parallelisms to the [V2 Dataplane's model metadata](https://github.com/kubeflow/kfserving/blob/master/docs/predict-api/v2/required_api.md#model-metadata). Therefore, it would be good to explore how we could convert from the model signature (which should be present in a MLflow's model artifact) to the model metadata "in-the-fly" so that the same information can still be exposed.

Note that there is currently a mismatch between [MLflow's native types](https://www.mlflow.org/docs/latest/python_api/mlflow.types.html#mlflow.types.DataType) and the [V2 Dataplane](https://github.com/kubeflow/kfserving/blob/master/docs/predict-api/v2/required_api.md#tensor-data-1) accepted data types (e.g. `string` vs `BYTES`). Therefore, there may be some loss of information when converting from the former to the latter format. We can explore further down the line how this loss of information can be minimised.

adriangonz:
Fixes #164 #162

## Changelog

- Sync input and output metadata from MLflow's model signature
- Add Base64 and Datetime codecs (to account for MLflow's `binary` and `datetime` types)
- Refactor packing / unpacking code outside of `StringCodec`


runtimes/mlflow/mlserver_mlflow/metadata.py
```diff
@@ -0,0 +1,81 @@
...
+
+    # TODO: Check if new type, which may not exist
+    return _MLflowToContentType[input_spec.type]
+
+
+def _get_shape(input_spec: InputSpec) -> List[int]:
+    if isinstance(input_spec, TensorSpec):
+        return list(input_spec.shape)
+
+    return [-1]
```
sakoush, line 42: what would -1 mean? would downstream treat it as a special case? Should we set it as `None` instead?

runtimes/mlflow/mlserver_mlflow/runtime.py
```diff
@@ -94,11 +95,26 @@ async def load(self) -> bool:
...
         self._model = mlflow.pyfunc.load_model(model_uri)
+
         self._input_schema = self._model.metadata.get_input_schema()
+        self._signature = self._model.metadata.signature
+        self._sync_metadata()
 
         self.ready = True
         return self.ready
 
+    def _sync_metadata(self):
```
sakoush, line 106: I generally like to have a return type `None` in these cases just to make it explicit that we didnt forget about return types, but just a slight preference and I think your still is also valid.

runtimes/mlflow/tests/test_metadata.py
```diff
@@ -0,0 +1,217 @@
...
+                    name="input-1",
+                    datatype="INT32",
+                    shape=[-1],
+                    tags=Tags(content_type=NumpyCodec.ContentType),
+                ),
+            ],
+        ),
+    ],
+)
+def test_to_metadata_tensors(schema: Schema, expected: List[MetadataTensor]):
```
sakoush, line 128: should we have a case for multiple `TensorSpec` as input

runtimes/mlflow/tests/test_metadata.py
```diff
@@ -0,0 +1,217 @@
...
+            ),
+        ),
+        (
+            ColSpec(name="foo", type=DataType.datetime),
+            RequestInput(
+                name="foo",
+                datatype="BYTES",
+                parameters=Parameters(content_type=DatetimeCodec.ContentType),
+                shape=[1, 19],
+                data=b"2021-08-24T15:01:19",
```
sakoush, line 174: should we have a case with timezone aware datetime?

tests/codecs/test_base64.py
```diff
@@ -0,0 +1,109 @@
...
+            [b"Python is fun"],
+            ResponseOutput(
+                name="foo",
+                shape=[1, 20],
+                datatype="BYTES",
+                data=b"UHl0aG9uIGlzIGZ1bg==",
+            ),
+        ),
+        (
+            ["Python is fun"],
```
sakoush, line 20: same as about test case?

tests/codecs/test_base64.py
```diff
@@ -0,0 +1,109 @@
...
+        (
+            RequestInput(
+                name="foo",
+                shape=[1, 20],
+                datatype="BYTES",
+                data=b"UHl0aG9uIGlzIGZ1bg==",
+            ),
+            [b"Python is fun"],
+        ),
+        (
```
sakoush, line 58: repeated test case?

tests/codecs/test_base64.py
```diff
@@ -0,0 +1,109 @@
...
+            RequestInput(
+                name="foo",
+                shape=[1, 13],
+                datatype="BYTES",
+                data=b"Python is fun",
+            ),
+            [b"Python is fun"],
+        ),
+        (
+            RequestInput(
```
sakoush, line 77: repeated test case?

tests/codecs/test_base64.py
```diff
@@ -0,0 +1,109 @@
...
+                name="foo",
+                shape=[2, 20],
+                datatype="BYTES",
+                data=b"UHl0aG9uIGlzIGZ1bg==UHl0aG9uIGlzIGZ1bg==",
+            ),
+            [b"Python is fun", b"Python is fun"],
+        ),
+        (
+            RequestInput(
```
sakoush, line 95: repeated tests case?

tests/codecs/test_datetime.py
```diff
@@ -0,0 +1,99 @@
...
+            RequestInput(
+                name="foo",
+                shape=[1, 19],
+                datatype="BYTES",
+                data=TestDatetimeIsoB,
+            ),
+            [TestDatetime],
+        ),
+        (
+            RequestInput(
```
sakoush, line 67: repeated test case?

tests/codecs/test_datetime.py
```diff
@@ -0,0 +1,99 @@
...
+                name="foo",
+                shape=[1, 19],
+                datatype="BYTES",
+                data=TestDatetimeIsoB + TestDatetimeIsoB,
+            ),
+            [TestDatetime, TestDatetime],
+        ),
+        (
```
sakoush, line 84: repeated test case?

tests/codecs/test_datetime.py
```diff
@@ -0,0 +1,99 @@
+import pytest
+
+from datetime import datetime
+
+from mlserver.codecs import DatetimeCodec
+from mlserver.types import RequestInput, ResponseOutput
+
+TestDatetimeIso = "2021-08-24T15:01:19"
```
sakoush, line 8: do we care about timezone aware testing as well?

sakoush (commented):
Looks great!

all minor comments really.

The only comment to double check is why `to_metadata` is not being used apart from tests?

mlserver/codecs/base64.py
```diff
@@ -0,0 +1,60 @@
...
+        return ResponseOutput(
+            name=name,
+            datatype="BYTES",
+            shape=shape,
+            data=packed,
+        )
+
+    @classmethod
+    def decode(cls, request_input: RequestInput) -> List[bytes]:
+        packed = request_input.data.__root__
```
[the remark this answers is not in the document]
adriangonz: Totally, that's a great shout. I'll try to add one next time I need to access `__root__`.

mlserver/codecs/base64.py
```diff
@@ -0,0 +1,60 @@
...
+
+    ContentType = "base64"
+
+    @classmethod
+    def encode(cls, name: str, payload: List[bytes]) -> ResponseOutput:
+        # Assume that payload is already in b64, so we only need to pack it
+        packed, shape = pack(map(_encode_base64, payload))
+        return ResponseOutput(
+            name=name,
+            datatype="BYTES",
```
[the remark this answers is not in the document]
adriangonz: Yuup, that'd be another improvement. Actually, I think that we may be able to specify that on the OpenAPI spec so that the generated types reflect it (and generate the enum as well).

mlserver/codecs/pack.py
```diff
@@ -0,0 +1,46 @@
...
+    common_length = -1
+    N = 0
+    for elem in unpacked:
+        # TODO: Should we use the length of the UTF8 string or the bytes
+        # array?
+        elem_length = len(elem)
+        if common_length == -1:
+            common_length = elem_length
+
+        if common_length != elem_length:
```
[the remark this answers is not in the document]
adriangonz: Good spot. That is partly a limitation of the v2 protocol (and / or gRPC and protobuf). Since we pass everything as a tensor, we only have a flattened list of elements and a shape. Thus, is not possible to have variable lengths. 

Having said that I do think that, at the moment, it will accept variable inputs if you use JSON (the codec will also accept inputs looking like `["asd", "hello world", "etc"]`). However, that won't work on the way back when we serialise the response.

runtimes/mlflow/mlserver_mlflow/metadata.py
```diff
@@ -0,0 +1,81 @@
...
+                datatype=datatype,
+                shape=shape,
+                tags=Tags(content_type=content_type),
+            )
+        )
+
+    return metadata_tensors
+
+
+def to_metadata(
```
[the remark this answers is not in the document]
adriangonz: I think I was using it at some point, but then reverted back to the other methods. I'll remove it.

runtimes/mlflow/mlserver_mlflow/metadata.py
adriangonz: I see `-1` as the "batch dimension". I think Triton also uses the same convention. So in this case it would just mean a list of "shapeless things". We could use `[None]`, but I'm not sure whether / how that would get serialised on the metadata endpoints. 

For this case, do you think that maybe we should just return an empty list?

runtimes/mlflow/mlserver_mlflow/runtime.py
adriangonz: Do you mean as part of the type hints? Or on the `return` line?

runtimes/mlflow/tests/test_metadata.py
adriangonz: Sure! I'll add one.

runtimes/mlflow/tests/test_metadata.py
adriangonz: Yuuup, good point. I'll add one.

tests/codecs/test_base64.py
adriangonz: There should be some small variations on each test case. E.g. one uses a "binary string" whereas the other uses a regular one, or one returns multiple elements, etc.

tests/codecs/test_datetime.py
adriangonz: We don't care that much, in the sense that `datetime` should be able to parse the timezone if it's present on the string. But I'll add a test to make sure that this actually happens.

runtimes/mlflow/mlserver_mlflow/runtime.py
sakoush: as a type hint ie. `def _sync_metadata(self) -> None:`

tests/codecs/test_base64.py
sakoush: oh yes! should we add a note perhaps or run the test twice on each usecase for example if the testcase is "Python is fun" -> we run once with [b"Python is fun"] and with ["Python is fun"] to reduce redundancy?

sakoush (approved):
LGTM

PR #288, opened 2021-08-25
Improve support for binary numpy arrays

### LINKED ISSUE #253: Consider using `tobytes()` instead of `.tolist()` in numpy codec for performance improvement
Our use-case is to make multiple requests to MLServer and all of them will have a 640x640 image data in numpy format as input.

Given that `tobytes()` is way faster than `tolist()` does it make sense to support it by default in the numpy codec here: https://github.com/SeldonIO/MLServer/blob/61d9f078d1f886d14083551f33db6a3146f12745/mlserver/codecs/numpy.py#L54

----

Performance comparison of the two methods:

Screenshot 2021-07-31 at 9 19 32 PM

adriangonz:
Fixes #253

## Changelog

- Use `tobytes()` to serialise Numpy `bytes` arrays
- Allow to deserialise `BYTES` inputs into Numpy `bytes` arrays


so2bin:
@adriangonz Hello, encoding numpy to bytes will result to JSON encoder failure of InferenceResponse,  because bytes data type cannot be encoded to json.

adriangonz:
Hey @so2bin , 

That's a great point, and one which should have been already tackled in #409. That PR has already been merged to `master` but not released yet. Could you try out the `master` version of MLServer to confirm whether this bug has been resolved?

so2bin:
@adriangonz Hello, i have read you commited codes in #409: https://github.com/adriangonz/MLServer/blob/f2e02ff1604fb80b8f12b5a530e6b90025f3bb48/mlserver/codecs/numpy.py#L71
The error is still exists, as `[data.tobytes()]` cannot serialise to json too. Bytes used in `ResponseOutput` may worked in GRPC, but will failed in HTTP transimission. The only way to fix it may be  encoding bytes to string with BASE64.

adriangonz:
Hey @so2bin , this should be taken care of on the `mlserver.rest.responses.Response` class:

https://github.com/SeldonIO/MLServer/blob/ef8e83ba614f597f5015816d72aacb7f2b631f96/mlserver/rest/responses.py#L33-L52

The current workaround is a bit rough, but it should unblock the use case of sending back bytes for now. Have you tested out the latest `master` with your use case?

PR #301, opened 2021-09-07
Use spawn instead of fork for inference workers

### LINKED ISSUE #300: Tempo example fails when parallel inference is enabled
When parallel inference is enabled, the [outlier example using the Tempo runtime](https://tempo.readthedocs.io/en/latest/examples/outlier/README.html)seems to fail. In particular, it seems that either the `cifar10-service` or the `outlier` containers block the request path and never return a response.

adriangonz:
Fixes #300 

## Changelog

- Use `spawn` instead of `fork` to avoid issues like https://github.com/tensorflow/tensorflow/issues/8220

### DIFF
--- a/mlserver/parallel.py
+++ b/mlserver/parallel.py
@@ -1,4 +1,5 @@
 import asyncio
+import multiprocessing as mp
 
 from functools import wraps
 from concurrent.futures import ProcessPoolExecutor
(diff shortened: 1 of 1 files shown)


sakoush (approved):
LGTM

Issue #257, opened 2021-08-02
Support for headers in InferenceRequest and InferenceResponse


axsaucedo:
Currently there is no way to access the headers of the request, nor set the headers of the response. This is a key requirement for features such as protocol-specific cloudevents support, as well as reducing code duplication for headers setting logic across rest, grpc and kafka (When supported https://github.com/SeldonIO/MLServer/issues/256). There are some complexities on this including:

* Currently swagger schema is created from InferenceRequest/response so adding `headers` attribute would introduce inconsistency
* Custom protocols may require different considerations to ensure this can be set through the current infer middlewares

PR #305, opened 2021-09-09
Adaptive Batching

### LINKED ISSUE #239: Adaptive batching
Explore whether MLServer could include support for adaptive batching. That is, dynamically grouping incoming requests together into a single batch, so that we can leverage any vectorisation / optimisation in the underlying model.

adriangonz:
Fixes #239 

## Changelog

- Add support for AsyncIO-based adaptive batching
- Add `max_batch_timeout` and `max_batch_size` settings to `model-settings.json`

## TODO

- [x] Add docs around adaptive batching
- [x] Add support for multiple-input payloads
- [x] Properly handling cases where request ID is not set (is this an issue?)


mlserver/batching/requests.py
```diff
@@ -0,0 +1,121 @@
...
+        self.merged_request = self._merge_requests()
+        self._prediction_ids: List[str] = [  # type: ignore
+            req.id for req in self._inference_requests  # type: ignore
+        ]
+
+    def _merge_requests(self) -> InferenceRequest:
+        inputs_index: Dict[str, List[RequestInput]] = defaultdict(list)
+
+        for inference_request in self._inference_requests:
+            # TODO: What should happen if UID is empty?
```
axsaucedo, line 62: I guess it should always be populated, as it's either provided or should be automaticallyy generated (that's what happens in executor)

mlserver/batching/requests.py
```diff
@@ -0,0 +1,121 @@
...
+            for request_input in inference_request.inputs:
```
axsaucedo, line 63: Would we also want to propagate other metadata like parameters? Or even headers?

mlserver/batching/requests.py
```diff
@@ -0,0 +1,121 @@
...
+                inputs_index[request_input.name].append(request_input)
+
+        # TODO: What should happen if input cardinality is different? (e.g.
+        # inputs with 2 inputs and others with 1)
```
axsaucedo, line 67: Good point, I think we could put a constraint that it should always have the same number of inputs, or have a default... but the latter could be harder - I can't think of a situation where a model could take an input as optional (unless it's some custom python code). Maybe if you just add it with an error and pick it up when a usecase comes up

mlserver/batching/requests.py
```diff
@@ -0,0 +1,121 @@
...
+    if isinstance(sampled_datum, str):
+        return "".join(all_data)
+
+    if isinstance(sampled_datum, bytes):
+        return b"".join(all_data)
+
+    if isinstance(sampled_datum, list):
+        return sum(all_data, [])
+
+    # TODO: Should we raise an error if we couldn't merge the data?
```
axsaucedo, line 37: Hmm good question - I think we should be explicit on how we deal with errors, namely whether we'd rather ensure "it still gets processed" even if has to be sent without adaptive batching...

cliveseldon: I'd say raise an error - batches need to be same shape - so better to be explicit than try to predict.

mlserver/batching/requests.py
cliveseldon: Definitely an error I'd say

cliveseldon:
Interesting implementation in Triton. https://github.com/triton-inference-server/server/blob/main/docs/model_configuration.md#dynamic-batcher
A few more tunable settings which might be worth considering if useful in our non-GPU case

mlserver/batching/requests.py
adriangonz: Yuuup, good point. It should be safe to assume that it's going to be set up. In fact, I think that the layer above this method actually sets it if it's unset.

mlserver/batching/requests.py
adriangonz: Great catch. I think we should. We don't have the headers accessible there yet (should be part of #257 ), but once we do, we should propagate them as well. 

For now, I'll add the `parameters` (and probably `outputs` as well?). We can check whether they're all the same, and if not we can always raise an exception for explicitness.

mlserver/batching/requests.py
adriangonz: Just to clarify, my impression is that there are two different axes here. On one hand, we could have different inputs with a different number of "input heads". This is the tricky case, involving optional parameters, where we should totally raise an error. 

On the other hand, we have the case where a payload could already have multiple entries within the same input head (i.e. the request has a "mini batch" within it already). In this case, we could just append the `n` elements contained within that request, and just keep the count of how many elements correspond to each request ID. 

Before adding that in though, it would be great to hear your thoughts about the latter case.

mlserver/batching/requests.py
adriangonz: I was thinking of falling back to non-batched prediction, which shouldn't be too tricky. However, this is more of an argument between being explicit or not. 

I guess there's a strong argument for raising that to the user so that they're aware that the server couldn't batch a particular request. Although we could also make that clear through logging maybe?

mlserver/batching/requests.py
cliveseldon: Is there an example of the former case?

For the latter it should certainly add to mini-batches together I think. But if hard could be a later change.

adriangonz: The latter (i.e. merging variable size mini-batches) should now be also supported :+1:

mlserver/batching/requests.py
cliveseldon: But is it valid for this to happen if so its not an error.

docs/user-guide/adaptive-batching.md
```diff
@@ -0,0 +1,73 @@
...
+your model](#usage).
+Since these values are usually found through experimentation, **MLServer won't
+enable by default adaptive batching on newly loaded models**.
+
+## Usage
+
+MLServer lets you configure adaptive batching independently for each model
+through two main parameters:
+
+- **Maximum batch size**, that is how many requests you want to group together.
```
cliveseldon, line 45: Interesting that Triton has max_batch_size as a core settings outside adaptive batching. I suppose this is for memory reasons to not allow too big a request irrespective of if its in a single request or in an adaptive batch

cliveseldon: So can users set `max_batch_size` outside adaptive batching?

cliveseldon:
Looks good to me. Just added a few small comments.

docs/user-guide/adaptive-batching.md
adriangonz: No, not really. This would only apply for adaptive batching. Outside of it, the model accepts whatever it gets. 

In the case of Triton, do you mean that it will split a request into smaller chunks if it's too large? That's interesting.

mlserver/batching/requests.py
axsaucedo: I guess perhaps the issue would be more if the user sends user defined IDs that happen to not be unique... I guess the issue here would be the assumption that the IDs are unique... perhaps here the comment to add is whether we'd want to add a separate internal unique identifier to avoid clashes?

mlserver/batching/requests.py
axsaucedo: Ok that makes sense

adriangonz: Just added the `parameters`, by merging them all together. The `outputs` may be a bit of a tricky one, but I'll create that as a separate issue.

mlserver/batching/requests.py
```diff
@@ -0,0 +1,145 @@
...
+        shape.pop(batch_axis)
+        shape.insert(batch_axis, batch_size)
+
+        return RequestInput(
+            name=sampled.name, datatype=sampled.datatype, shape=shape, data=data
+        )
+
+    def _merge_data(self, request_inputs: List[RequestInput]) -> Tuple[int, Any]:
+        # We assume all inputs will have the same minibatch size
+        self._minibatch_sizes = []
```
sakoush, line 92: is that when the inputs are already coming batched from the client?

sakoush (commented):
Looks great!
I have a comment on the term adaptive. I am not sure how adaptive is the current implementation with regards that it is adapting to the underlying hardware resources (as you mention in the docs)? the parameter  `max_batch_time` is not taking account of the the time it takes to do the actual prediction? i.e. the decision to add a prediction request to a batch depends on arrival time and perhaps the load on the request queue but not taking into account how long in the prediction task? or maybe I havent understood your implementation?

mlserver/batching/adaptive.py
```diff
@@ -0,0 +1,94 @@
...
+    async def _queue_request(
+        self,
+        req: InferenceRequest,
+    ) -> Awaitable[InferenceResponse]:
+        await self._requests.put(req)
+
+        loop = asyncio.get_running_loop()
+        async_response = loop.create_future()
+
+        # TODO: What happens if ID is None?
```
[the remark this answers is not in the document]
sakoush: ok it could be none if we construct `InferenceRequest` manual and not passing an id as when we do the batching.

mlserver/batching/requests.py
```diff
@@ -0,0 +1,165 @@
...
+            datatype=sampled.datatype,
+            shape=shape,
+            data=data,
+            parameters=parameters,
+        )
+
+    def _merge_data(
+        self, request_inputs: List[RequestInput]
+    ) -> Tuple[int, Any, Optional[Parameters]]:
+        # We assume all inputs will have the same minibatch size
```
sakoush, line 91: is this comment still true?

tests/batching/test_adaptive.py
```diff
@@ -0,0 +1,92 @@
...
+    )
+    batched_requests = [
+        batched_req async for batched_req in adaptive_batcher._batch_requests()
+    ]
+
+    assert len(batched_requests) == 1
+    assert batched_requests[0]._inference_requests == sent_requests
+
+
+async def test_batch_requests_timeout(
```
sakoush, line 26: maybe add a note why this is testing timeout and perhaps send another request after timeout and then assert that you end up with two batches?

adriangonz:
That's a good question @sakoush. I think the term "adaptive" may be confusing here, although it seems that's what other projects also use for this feature (alongside "predictive batching", which can be even more confusing).

As far as I understand, the goal here is to batching requests together on-the-fly based on a fixed set of parameters (usually max batch size, and max time to wait). That is, not dynamically adjusting the params.

docs/user-guide/adaptive-batching.md
```diff
@@ -0,0 +1,73 @@
...
+The `maximum_batch_size` field of the `model-settings.json` file (or
+alternatively, the `MLSERVER_MODEL_MAXIMUM_BATCH_SIZE` global environment
+variable) controls the maximum number of requests that should be grouped
+together on each batch.
+The expected values are:
+
+- `N`, where `N > 1`, will create batches of up to `N` elements.
+- `0` or `1`, will disable adaptive batching.
+
+### `maximum_batch_time`
```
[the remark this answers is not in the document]
adriangonz: It's pretty much a case-by-case approach. Depending on how inference is implemented on each model, the underlying hardware and how long inference takes, some values will get a better performance than others.

We suggest benchmarking your model to find the right parameters, but maybe we could also add something about how they interact together? We could also point to Triton's tool which (as far as I understand) will benchmark your model to find out the best combination of parameters. Although I'm not sure if that one would play well with MLServer.

tests/batching/test_adaptive.py
```diff
@@ -0,0 +1,92 @@
...
+        assert async_response.done()
+
+        response = await async_response
+        assert sent_request.id == response.id
+
+        expected = await sum_model.predict(sent_request)
+        assert expected == response
+
+
+async def test_predict(adaptive_batcher: AdaptiveBatcher, sum_model: MLModel):
```
sakoush, line 65: what will happen if the first minibatch is greater than Maximum batch size? any edge cases we need to consider in this case? perhaps add a test for that?

mlserver/batching/adaptive.py
```diff
@@ -0,0 +1,94 @@
...
+        self._async_responses[req.id] = async_response  # type: ignore
+
+        return async_response
```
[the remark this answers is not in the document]
adriangonz: Yes, we await it later on the `_wait_response` method.

mlserver/batching/adaptive.py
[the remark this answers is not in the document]
adriangonz: Yes, exactly. The handlers should set if it's empty, although in SC it would be the executor the one ensuring that there's one set.

However, as @axsaucedo raised, it could be the case that the IDs are not unique. We were chatting of whether it would make sense to generate our own UIDs in the batching code, rather than relying on the request ones.

mlserver/batching/adaptive.py
```diff
@@ -0,0 +1,94 @@
...
+        if self._batching_task is not None:
+            if not self._batching_task.done():
+                # If task hasn't finished yet, let it keep running
+                return
+
+        # TODO: If _batching_task crashes, cancel all async responses
+        self._batching_task = asyncio.create_task(self._batcher())
+
+    async def _batcher(self):
+        async for batched in self._batch_requests():
```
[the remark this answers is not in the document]
adriangonz: Good point. That's my impression, but I'm not 100% sure. We should write a quick test to ensure that.

mlserver/batching/requests.py
adriangonz: Yes, exactly. I'll add some comments to clarify what do we mean by minibatches.

tests/batching/test_hooks.py
```diff
@@ -0,0 +1,39 @@
...
+    [
+        (0, 10),
+        (-1, 10),
+        (1, 10),
+        (10, 0),
+        (10, -1),
+        (0, 0),
+    ],
+)
+async def test_load_batching_disabled(
```
sakoush, line 30: is batching disabled because either max_batch_size or max_batch_time is not?

how are you making sure that actually batching is disabled and not exercised?

tests/batching/test_requests.py
```diff
@@ -0,0 +1,372 @@
...
+    "request_inputs,expected",
+    [
+        (
+            [
+                RequestInput(
+                    name="foo",
+                    datatype="INT32",
+                    shape=[1, 3],
+                    data=[1, 2, 3],
+                    parameters=Parameters(content_type="np"),
```
sakoush, line 25: hmm, what will happen if the content_type is different for different RequestInput, i.e. if one request is sent for some reason as numpy representation and one as dataframe representation. I guess batch happens before we apply the codec? I dont know actually what will happen.

sakoush (commented):
comments on testing.

mlserver/batching/requests.py
```diff
@@ -0,0 +1,165 @@
...
+    request_input: Union[RequestInput, ResponseOutput]  # type: ignore
+) -> int:
+    # TODO: Support other batch dimensions
+    return 0
+
+
+class BatchedRequests:
+    def __init__(self, inference_requests: List[InferenceRequest] = []):
+        self._inference_requests = inference_requests
+        self._prediction_ids: List[str] = [  # type: ignore
```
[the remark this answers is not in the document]
adriangonz: This is so that it doesn't complain about the request IDs, which can be `None` (although shouldn't).

mlserver/batching/requests.py
```diff
@@ -0,0 +1,165 @@
...
+    ) -> Tuple[int, Any, Optional[Parameters]]:
+        # We assume all inputs will have the same minibatch size
+        self._minibatch_sizes = []
+        batch_size = 0
+        all_data = []
+        all_params = {}
+        for request_input in request_inputs:
+            if request_input.parameters:
+                input_params = request_input.parameters.dict(exclude_unset=True)
+                all_params.update(input_params)
```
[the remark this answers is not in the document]
adriangonz: The first one is for the request-level parameters, whereas the second one is for the input-level parameters. We should move it out to a separate function TBH - I'll do that :+1:

mlserver/batching/requests.py
adriangonz: It is, across different input heads. That would a 2nd-level type of variability, e.g. we've got 2 input heads (e.g. `words` and `images`), but we get 1 image for each 3 words. So we'd have to workout how to map each cardinality (e.g. 3 and 1) with the number of output tensors sent back. 

To simplify things, for now we assume that cardinality across inputs will always be the same. That is, there will be the same number of elements on each input head.

tests/batching/test_adaptive.py
adriangonz: Good point, I'll add that :+1:

tests/batching/test_adaptive.py
adriangonz: Interesting. I'd imagine it would go through together as a single batch (even if it's larger), but let me add a test for that to ensure nothing breaks :+1:

docs/user-guide/adaptive-batching.md
cliveseldon: I think in Triton its a head limit on max batch a server can consume. I assume it returns an error if you send a batch too large.

tests/batching/test_hooks.py
adriangonz: Depending on the values of `max_batch_size` and `max_batch_time`, we disable adaptive batching. On all the cases covered by that `pytest.mark.parametrize` above, the expectation is that adaptive batching should get disabled.

mlserver/batching/requests.py
sakoush: should there be an assertion / exception to be the case, if not already done?

mlserver/batching/adaptive.py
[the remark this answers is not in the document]
adriangonz: Right, it should now be changed to use an internal set of IDs (i.e. without relying on the external request IDs).

mlserver/batching/requests.py
adriangonz: Right, it should now be changed to use an internal set of IDs (i.e. without relying on the external request IDs).

mlserver/batching/adaptive.py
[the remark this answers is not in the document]
adriangonz: I had a look, and turns out that it doesn't run different iterations concurrently. Even if all the for-loop body is encapsuled in a function, it will still run each iteration sequentially. One potential fix would be to use `asyncio.gather`, however we won't necessarily have all batches at the same time.

Instead, the PR should now leverage `asyncio.create_task()` and callbacks to ensure that the batches get scheduled for prediction immediately and can run concurrently.

PR #320, opened 2021-09-23
Sherif akoush/mlsv 42/add alibi explain runtime

sakoush:
Very early draft


mlserver/codecs/numpy.py
```diff
@@ -93,11 +93,11 @@ def encode(cls, name: str, payload: np.ndarray) -> ResponseOutput:
         )
 
     @classmethod
-    def decode(cls, request_input: RequestInput) -> np.ndarray:
-        model_data = _to_ndarray(request_input)
+    def decode(cls, v2_data: Union[RequestInput, ResponseOutput]) -> np.ndarray:
```
adriangonz, line 96: Interesting! Do you think we should make this the new interface for the `InputCodec` class? I was thinking that we could also include the opposite conversion, although that would probably require splitting the existing `encode` method into `encode_response` and `encode_request`.

runtimes/alibi-explain/mlserver_alibi_explain/common.py
```diff
@@ -0,0 +1,84 @@
...
+    if ty == str:
+        return bytearray(output.data).decode("UTF-8")
+    else:
+        py_str = bytearray(output.data).decode("UTF-8")
+        from ast import literal_eval
+        return literal_eval(py_str)
+
+
+def remote_predict(v2_payload: InferenceRequest, predictor_url: str) -> InferenceResponse:
+    response_raw = requests.post(predictor_url, json=v2_payload.dict())
```
adriangonz, line 48: We could use something like `aiohttp` here, to ensure that requests don't hold up the AsyncIO loop (and make `remote_predict` async).

runtimes/alibi-explain/mlserver_alibi_explain/runtime.py
```diff
@@ -0,0 +1,97 @@
...
+        )
+
+        v2_response = remote_predict(
+            v2_payload=v2_request,
+            predictor_url=self.alibi_explain_settings.infer_uri)
+
+        return np_codec.decode(v2_response.outputs[0])  # type: ignore # TODO: fix mypy and first output
+
+    @abc.abstractmethod
+    def _explain_impl(self, input_data: Any, settings: BaseSettings) -> Explanation:
```
adriangonz, line 84: Would there be any scenario where we wouldn't want to execute it asynchronously?

runtimes/alibi-explain/mlserver_alibi_explain/common.py
```diff
@@ -0,0 +1,84 @@
...
+
+ENV_PREFIX_ALIBI_EXPLAIN_SETTINGS = "MLSERVER_MODEL_ALIBI_EXPLAIN_"
+
+
+class ExplainerEnum(str, Enum):
+    anchor_image = _ANCHOR_IMAGE_TAG
+
+
+# TODO: is this inference response?
+def create_v2_from_any(data: Any, name: str) -> Dict:
```
adriangonz, line 25: Should we wrap this into a Codec? Conscious that it's something kind of temporary, but could be maybe useful in other cases?

runtimes/alibi-explain/mlserver_alibi_explain/runtime.py
sakoush: no but I thought that implementation should not care about that and we get the logic and execute it async in base class

runtimes/alibi-explain/mlserver_alibi_explain/common.py
axsaucedo: I agree, these are the type of components that we'll end up also needing in Tempo if we want to easily create clients that can convert to/from explainers

PR #322, opened 2021-09-24
Infer model-wide content type from MLflow's model signature

### LINKED ISSUE #165: Expose MLflow's model signature types as "annotated" model metadata
As a follow up to #163 and #164, it would be interesting to explore whether it's possible to expose the types information missed when converting from MLflow's model signature to the V2 Dataplane through annotations in the V2 model metadata.

For example, in the case of MLflow's `string` type, one could *annotate* the V2 Dataplane metadata for that particular input saying that even though it should be encoded as `BYTES` at the *low-level*, it's meant to be compatible with MLflow's `string` type.

adriangonz:
Fixes #165 

## Changelog

- Rename `tags` field to `parameters`
- Add `parameters` field to model-wide metadata
- Use the `parameters` from the model metadata as the default `content_type` for the inference request
- Fill in the model metadata `parameters` field from MLflow's model signature


mlserver/codecs/utils.py
```diff
@@ -16,19 +17,19 @@
...
 ) -> Optional[str]:
-    if parametrised_obj.parameters and parametrised_obj.parameters.content_type:
-        return parametrised_obj.parameters.content_type
+    if request.parameters and request.parameters.content_type:
+        return request.parameters.content_type
 
-    if tagged_obj is not None:
-        if tagged_obj.tags and tagged_obj.tags.content_type:
-            return tagged_obj.tags.content_type
+    if metadata is not None:
```
sakoush, line 30: do we care if `request` and `metadata` disagree on `content_type` ?

mlserver/codecs/utils.py
```diff
@@ -59,12 +60,14 @@ def decode_request_input(
 
 
 def decode_inference_request(
-    inference_request: InferenceRequest, metadata_inputs: Dict[str, MetadataTensor] = {}
+    inference_request: InferenceRequest,
+    model_settings: ModelSettings = None,
+    metadata_inputs: Dict[str, MetadataTensor] = {},
```
sakoush, line 65: where is that used?

runtimes/mlflow/tests/test_metadata.py
```diff
@@ -204,3 +205,47 @@ def test_content_types(tensor_spec: TensorSpec, request_input: RequestInput):
...
+    "schema, expected",
+    [
+        (
+            Schema(
+                inputs=[
+                    ColSpec(name="foo", type=DataType.boolean),
+                    ColSpec(name="bar", type=DataType.boolean),
+                ]
+            ),
+            PandasCodec.ContentType,
```
sakoush, line 220: minor: perhaps describe or reference `to_model_content_type` logic as it was not entirely clear in the first instance why `expected` is as such.

runtimes/mlflow/mlserver_mlflow/runtime.py
```diff
@@ -122,10 +128,17 @@ def _sync_metadata(self) -> None:
...
         )
         self.outputs = to_metadata_tensors(
-            self._signature.outputs, prefix=DefaultOutputPrefix
+            schema=self._signature.outputs, prefix=DefaultOutputPrefix
+        )
+
+        if not self._settings.parameters:
+            self._settings.parameters = ModelParameters()
+
+        self._settings.parameters.content_type = to_model_content_type(
```
sakoush, line 140: are we possibly overwriting the setting that is provided by the user? should we have a warning?

docs/examples/mlflow/README.md
```diff
@@ -195,108 +195,72 @@ To learn more about how MLServer uses content type parameters, you can check thi
 import requests
 
 inference_request = {
-    "parameters": {
```
sakoush, line 198: hmm, we are no longer providing `content_type` in the case of mlflow and using the defaults from `Schema`

what happens if we provide the wrong `content_type`. it is similar to a question I have below in the case there is a mismatch between the two.

sakoush (approved):
LGTM, I guess this allow less settings from the user which is good in the case of mlflow.

The main question I have is what is going to happen if there is a mismatched between the inference and the user/settings `content_type`?

mlserver/codecs/utils.py
adriangonz: In that case, I think that the content type `request` should override whatever it's on the metadata one (mainly because in that case the user would be explicitly saying what content type they want to set).

mlserver/codecs/utils.py
adriangonz: There's a `self.decode_request()` method in the `MLModel` base class which calls this one. Initially, there was also a middleware layer which would call this method, although we scrapped that idea for now.

runtimes/mlflow/mlserver_mlflow/runtime.py
adriangonz: Good point, we probably should. Same with `inputs` and `outputs`, which are already overriden by the MLflow model signature. 

I'll add that to the PR.

docs/examples/mlflow/README.md
adriangonz: My understanding is that if the user sends a explicit content type through the request `parameters`, we should then honour that one (as they're trying really hard to use that particular codec). 

I think that should be the case now.

runtimes/mlflow/tests/test_metadata.py
adriangonz: Good point, let me annotate the `parametrize` cases.

PR #323, opened 2021-09-27
Improve logging on MLServer and add logging to gRPC requests

### LINKED ISSUE #37: Add logging interceptor to gRPC server
Currently there are no logs in the gRPC server. For that, we need to add a (custom?) interceptor that logs a message for each incoming request.

adriangonz:
Fixes #37

### DIFF
--- a/mlserver/grpc/interceptors.py
+++ b/mlserver/grpc/interceptors.py
@@ -1,5 +1,3 @@
-import logging
-
 from typing import Callable, Awaitable
 
 from grpc.aio import ServerInterceptor
--- a/mlserver/grpc/server.py
+++ b/mlserver/grpc/server.py
@@ -71,7 +71,8 @@ class GRPCServer:
         await self._server.start()
 
         logger.info(
-            f"gRPC server running on http://{self._settings.host}:{self._settings.grpc_port}"
+            "gRPC server running on "
+            f"http://{self._settings.host}:{self._settings.grpc_port}"
         )
         await self._server.wait_for_termination()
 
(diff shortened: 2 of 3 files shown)


mlserver/grpc/interceptors.py
```diff
@@ -0,0 +1,20 @@
...
+class LoggingInterceptor(ServerInterceptor):
+    def _get_log_message(self, handler_call_details: HandlerCallDetails) -> str:
+        return handler_call_details.method
+
+    async def intercept_service(
+        self,
+        continuation: Callable[[HandlerCallDetails], Awaitable[RpcMethodHandler]],
+        handler_call_details: HandlerCallDetails,
+    ) -> RpcMethodHandler:
+        logger.info(self._get_log_message(handler_call_details))
```
sakoush, line 18: just for my reference, what is an example of `handler_call_details.method` ?

also do we have any potential issues with writing async to the same log file? especially with `ProcessThreadExecutor` ?

sakoush (approved):
LGTM

mlserver/grpc/interceptors.py
sakoush: I realise that `ProcessThreadExecutor` is lower in the call stack so probably not having an effect here in the grpc server.

axsaucedo:
LGTM +1 this will be quite useful. It may be useful also adding documentation for users to use the logger

mlserver/grpc/interceptors.py
adriangonz: That's a good point, not sure what would happen with the logger on the inference workers. What would `spawn` do in this case? Would it just replicate the standard output handler?

On the former point, `handler_call_details.method` would be a string like `inference.GRPCInferenceService/ModelInfer` (i.e. `<proto package>.<service name>/<service method>`). Or at least that's what I found out while experimenting!

PR #328, opened 2021-10-04
Refactor Dockerfile and split images per runtime

### LINKED ISSUE #302: Separate Docker images per runtime
Currently, the `seldonio/mlserver` includes every runtime. This leads to quite a large image. It would be good to also create per-runtime images, where only the specific ones are included.

Additionally, we should also have a "slim" image, which contains no runtimes, to allow for the use case where an environment tarball is also provided.

adriangonz:
Fixes #302 

## Changelog
- Refactor `Dockerfile`, which now builds the images in 2 stages: 
  1. Build all the Python wheels in a first "builder" image
  2. Install the wheels which are selected
- Add a special `slim` image, with no runtimes (meant to be used with custom environments)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,34 +1,44 @@
-FROM python:3.8-slim
+FROM python:3.8-slim AS wheel-builder
+SHELL ["/bin/bash", "-c"]
 
+COPY ./hack/build-wheels.sh ./hack/build-wheels.sh
+
+COPY setup.py .
+COPY README.md .
+COPY ./mlserver/ ./mlserver/
+COPY ./runtimes/ ./runtimes/
+
+# TODO: Cache this step with BuildKit?
+# This will build the wheels and place will place them in the
+# /opt/mlserver/dist folder
+RUN ./hack/build-wheels.sh /opt/mlserver/dist
+
+FROM python:3.8-slim
 SHELL ["/bin/bash", "-c"]
 
 ENV MLSERVER_MODELS_DIR=/mnt/models \
     MLSERVER_ENV_TARBALL=/mnt/models/environment.tar.gz \
-    PATH=/home/default/.local/bin:$PATH
+    PATH=/opt/mlserver/.local/bin:$PATH
 
 RUN apt-get update && \
     apt-get -y --no-install-recommends install \
-    libgomp1 && \
-    apt-get install -y libgl1-mesa-dev && \
-    apt-get install -y libglib2.0-0 && \
-    pip install --upgrade pip wheel setuptools
+        libgomp1 libgl1-mesa-dev libglib2.0-0
 
 RUN mkdir /opt/mlserver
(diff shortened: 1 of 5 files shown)


docs/examples/conda/README.ipynb
```diff
@@ -29,9 +29,17 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 2,
    "metadata": {},
-   "outputs": [],
+   "outputs": [
```
agrski, line 34: :question: Are the examples supposed to show output?  Asking as it seems they were left blank previously

Dockerfile
```diff
@@ -1,34 +1,54 @@
...
+SHELL ["/bin/bash", "-c"]
+
+COPY ./hack/build-wheels.sh ./hack/build-wheels.sh
+
+COPY setup.py .
+COPY README.md .
+COPY ./mlserver/ ./mlserver/
+COPY ./runtimes/ ./runtimes/
 
+# TODO: Cache this step with BuildKit?
```
agrski, line 11: :question: Is this TODO still relevant?

hack/build-wheels.sh
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env bash
+
+set -o nounset
+set -o errexit
+set -o pipefail
+
+ROOT_FOLDER="$(dirname "${0}")/.."
```
agrski, line 7: :thought_balloon: I know this form is used elsewhere, but a nice, reliable alternative for finding the root directory of a git repo is `git rev-parse --show-toplevel`

hack/build-wheels.sh
```diff
@@ -0,0 +1,42 @@
...
+  # `cd` into it and the go back again.
+  cd $_srcPath
+  python setup.py \
+    sdist -d $_outputPath \
+    bdist_wheel -d $_outputPath
+  cd $_currentDir
+}
+
+_main() {
+  local _outputPath="$PWD/$1"
```
agrski, line 30: :confused: In [this line of the Dockerfile](https://github.com/SeldonIO/MLServer/pull/328/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R14) it looks like the given path is intended to be absolute, not relative

Makefile
```diff
@@ -2,7 +2,8 @@ SHELL := /bin/bash
 VERSION := $(shell sed 's/^__version__ = "\(.*\)"/\1/' ./mlserver/version.py)
 IMAGE_NAME := seldonio/mlserver
 
-.PHONY: install-dev _generate generate run build push-test push test lint fmt version clean licenses
+.PHONY: install-dev _generate generate run build build-pypy build-docker\
```
agrski, line 5: :question: I can't see `build-pypy` or `build-docker` here - have those changes not yet been pushed?

PR #377, opened 2021-11-08
fix wheel install

sakoush:
docker fix install latest mlserver (instead of released)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,7 +37,7 @@ RUN useradd -u 1000 -s /bin/bash mlserver -d /opt/mlserver && \
 
 COPY --from=wheel-builder /opt/mlserver/dist ./dist 
 RUN pip install --upgrade pip wheel setuptools && \
-    pip install ./dist/mlserver-*.whl[all]; \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -37,7 +37,7 @@ RUN useradd -u 1000 -s /bin/bash mlserver -d /opt/mlserver && \
 
 COPY --from=wheel-builder /opt/mlserver/dist ./dist 
 RUN pip install --upgrade pip wheel setuptools && \
-    pip install ./dist/mlserver-*.whl[all]; \
+    pip install $(ls ./dist/mlserver-*.whl)[all]; \
```
adriangonz, line 40: Should we add a note of why is the fix required? Mainly to ensure no one (probably us!) sees this in a couple months, thinks "what the heck?" and decides to remove it.

adriangonz (changes requested):

Dockerfile
sakoush: done

adriangonz (approved):
LGTM (once we sort out the tests issues)

Issue #342, opened 2021-10-13
Seldon MLServer only supports a single input tensor 


amnpandey:
We are getting below error when we are trying KFS V2 request with multiple tensors:
```
InvalidArgument (3)
inference.GRPCInferenceService/ModelInfer: INVALID_ARGUMENT: SKLearnModel only supports a single input tensor (4 were received)
```

we have checked the implementations and it looks like it has been designed to work with only single tensor.
https://github.com/SeldonIO/MLServer/blob/56a90a45119ed5ab3cd7c99c68a109f5426828f6/runtimes/sklearn/mlserver_sklearn/sklearn.py#L47

Any thought(s) to make it work for multiple tensors. Thanks

amnpandey:
closing this for now as we would like to investigate more on our side and then come back. Thanks

joerunde:
@adriangonz Are you actively looking into this?

Not sure how you guys like to handle code snippets, but I put up a reproduction of this in https://github.com/SeldonIO/MLServer/pull/379. I'd like to try to contribute a fix, I think I just might need a bit of guidance on how to go about changing the input processing logic in the sklearn runtime without accidentally breaking any other use cases of the runtime.

Is there a set of test suites somewhere with some canonical cases of SKLearn models that should be serveable by the SKLearn runtime? Or do you all rely on input from other contributors and users to try to maintain compatibility with existing use cases?

Any guidance is appreciated!

adriangonz:
Hey @joerunde , that'd be really welcome! Thanks a lot for jumping on this one. :+1: 

I've just realised that you've opened a draft PR with the changes, so we can keep the conversation going there.

adriangonz:
Fixed by #379

PR #379, opened 2021-11-09
Sklearn codecs

joerunde:
~~(Super not an actual pull request, just a reproduction)~~

This reproduces the bug found in https://github.com/SeldonIO/MLServer/issues/342 where the SKLearn runtime does not properly decode inputs with content type `pd`.

~~The files in the `342-scripts` directory can be used to create and serve an SKLearn pipeline that has a ColumnTransformer as its first step, which requires a pandas dataframe with heterogeneous column datatypes as input.
e.g.~~
```bash
# Generate some data
python3 342-scripts/gen.py
# Train a model
python3 342-scripts/pipeline.py
# Serve it
make run
```
~~And in another terminal~~
```bash
python3 342-scripts/request.py
```
~~With the code changes in commit a78709d, this should successfully decode the input and return a response. Without them, it returns a 400 with "SKLearnModel only supports a single input tensor (4 were received)".~~

Update:
I've gone ahead and reproduced the bug with tests in `runtimes/sklearn/tests` and updated the runtime code to what I think should appropriately handle it.

Incidentally, the use case that we were looking for was a regression pipeline that would not support `predict_proba`, so I've also added a runtime check that the actual model has a `predict_proba` method if that's one of the requested outputs.


adriangonz:
Nice one @joerunde! Thanks for your contribution. 

You've nailed the change required (i.e. calling `self.decode_request()`). That's pretty much all that's needed I think.

Regarding tests, there're a few unit tests for both `mlserver` and `mlserver-sklearn`. You should be able to run just the `mlserver-sklearn` ones with:

```bash
pytest runtimes/sklearn
``` 

Would it make sense to convert the snippets which replicate the issue into tests that cover this use case on the `mlserver-sklearn` runtime? It should be possible to follow the current Pytest fixtures that live on the `runtimes/sklearn/tests/conftest.py` module, and just add a new one (e.g. `pandas_sklearn`) that creates a SKLearn models which requires a Pandas input.

Happy to help out if there're any questions!

joerunde:
@adriangonz thanks for the pointers! I've gone ahead and added tests and cleaned up the implementation a bit if you have a few minutes to look it over.

I'm very open to feedback, I haven't looked too closely at the code style in the rest of the repo and I only just started working with `sklearn` a couple days ago so I may have messed some concepts up ¯\\\_(ツ)_/¯

Thanks!

runtimes/sklearn/mlserver_sklearn/sklearn.py
```diff
@@ -60,15 +56,38 @@ def _check_request(self, payload: types.InferenceRequest) -> types.InferenceRequ
...
 
     def _predict_outputs(
         self, payload: types.InferenceRequest
     ) -> List[types.ResponseOutput]:
-        model_input = payload.inputs[0]
-
         default_codec = NumpyCodec()
-        input_data = self.decode(model_input, default_codec=default_codec)
+        # TODO: how to set default codec here? Needs to be in the request...?
+        decoded_request = self.decode_request(payload)
```
adriangonz, line 78: Good point! The `decode_request()` method doesn't support a default codec atm. Maybe we should extend that method so that it does something similar to what the `decode()` method does? Happy to leave that out of the scope of the PR though.

runtimes/sklearn/mlserver_sklearn/sklearn.py
```diff
@@ -60,15 +56,38 @@ def _check_request(self, payload: types.InferenceRequest) -> types.InferenceRequ
...
+
+        # If we decode to an InferenceRequest again, then inputs is probably an array of tensors
+        if isinstance(decoded_request, types.InferenceRequest):
+            if len(decoded_request.inputs) != 1:
+                raise InferenceError(
+                    "SKLearnModel only supports a single input tensor "
+                    f"({len(payload.inputs)} were received)"
+                )
+            input_data = decoded_request.inputs[0]
+        # Otherwise we decoded to a different structure e.g. a pandas.DataFrame
```
adriangonz, line 88: Nice one on handling the edge case that no codec is present! 

I guess once we have a default for `decode_request()`, we'll be able to remove this extra logic. Just to re-emphasise though, happy to leave that outside of the scope of the PR.

adriangonz (approved):
Changes look great! Brilliant work! Thanks so much for your contribution @joerunde! LGTM

joerunde:
@adriangonz I am also perfectly fine with not making changes to `decode_request` in this PR, I really don't want to expand the blast radius past the sklearn runtime.

I ran the formatter locally which should have fixed the couple of linter errors, but I also noticed that the black exclusions weren't working properly. It looks like the automated run is still trying to reformat things from `mlserver/grpc/dataplane_pb2*`, and locally it's trying to reformat all the virtual environments in my `.tox` dir.
I think I've fixed that as well, and I took the liberty to go ahead and exclude `venv/` as well for the rest of the Pycharm homies out there.

adriangonz:
Nice one on the changes! I've gotten into the lazy habit of always doing `make clean lint`, so it's great you managed to add those to the ignore folders as well. 

I think this looks good to go. Thanks for your contribution @joerunde ! :+1:

PR #380, opened 2021-11-09
move to 0.6.0.dev2

sakoush:


sakoush:
@adriangonz not sure if I should use `0.6.0.dev1` instead of `0.6.0.dev2`?

adriangonz (approved):

PR #381, opened 2021-11-11
Make model URI in model-settings.json relative to the config file

### LINKED ISSUE #325: Make the model's URI relative to the model-settings.json file
Currently, the model URI is expected to be relative to where the `mlserver` command is running from. However, this can be unintuitive for users, who may expect that path to be relative to where it's specified, i.e. the `model-settings.json` file.

It would be good to consider whether changing the reference for the path could make this more intuitive for users and what could its side-effects.

adriangonz:
Fixes #325 

## Changelog
- Add private `_source` field to `ModelSettings` and tweak `ModelRepository` to store there the original file path of the `model-settings.json` file (if any).
- Tweak the `get_model_uri` utility method to base the model's URI on the location of the `model-settings.json` file.

### DIFF
--- a/mlserver/repository.py
+++ b/mlserver/repository.py
@@ -42,9 +42,11 @@ class ModelRepository:
 
     def _load_model_settings(self, model_settings_path: str) -> ModelSettings:
         model_settings = ModelSettings.parse_file(model_settings_path)
+        model_settings_folder = os.path.dirname(model_settings_path)
+        model_settings._source = model_settings_folder
 
         # If name not present, default to folder name
-        default_model_name = os.path.basename(os.path.dirname(model_settings_path))
+        default_model_name = os.path.basename(model_settings_folder)
         if model_settings.name:
             if model_settings.name != default_model_name:
                 # Raise warning if name is different than folder's name
--- a/mlserver/repository.py
+++ b/mlserver/repository.py
@@ -7,7 +7,7 @@ from .settings import ModelParameters, ModelSettings
 from .errors import ModelNotFound
 from .logging import logger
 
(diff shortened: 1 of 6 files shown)


mlserver/utils.py
```diff
@@ -13,11 +13,11 @@ async def get_model_uri(
     if not settings.parameters:
         raise InvalidModelURI(settings.name)
 
-    model_uri = settings.parameters.uri
-
-    if not model_uri:
+    relative_model_uri = settings.parameters.uri
```
sakoush, line 16: I think that the variable name `relative_model_uri` is slightly misleading here as it could also be absolute path?

perhaps also check directly whether it is absolute instead of relying on `normpath` (assuming that this is what happens currently which I am not 100% sure)

sakoush (commented):
LGTM

mlserver/repository.py
```diff
@@ -7,7 +7,7 @@
 from .errors import ModelNotFound
 from .logging import logger
 
-DEFAULT_MODEL_SETTINGS_FILENAME = "model-settings.json"
+DefaultModelSettingsFilename = "model-settings.json"
```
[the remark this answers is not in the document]
adriangonz: Good point! I've never been a big fan of the capital case syntax, but if that's what the standard says I guess we'll need to stick to it :grimacing: 

I'll revert back the changes.

mlserver/utils.py
adriangonz: We can call it `model_uri`? 

You're right in that it could also be an absolute path. In that case, `os.path.join()` should be the one handling it (it ignores the paths previous to an absolute path).

Issue #386, opened 2021-11-15
Feature Request: Add support for CORS configuration


JakeNeyer:
**Context:**
As kserve is moving to it's v2 api, it's defaulting the server images to ml-server. It looks like kserve and knative are pushing off adding CORS configurations on their CRDs due to complexity (https://github.com/knative/serving/issues/10040). The simplest way I have seen to get around that is adding an OPTIONS route and CORS options on the actual servers themselves.

**Request**
Add OPTIONS routes and the ability to apply runtime configuration for CORS on the MLServer

adriangonz:
Fixed by #387

PR #387, opened 2021-11-15
Add CORS support

JakeNeyer:
This PR will add configurable CORS middleware support on REST endpoints.

This references https://github.com/SeldonIO/MLServer/issues/386.

### DIFF
--- a/mlserver/rest/app.py
+++ b/mlserver/rest/app.py
@@ -2,6 +2,7 @@ from typing import Callable
 from fastapi import FastAPI
 from fastapi.responses import Response as FastAPIResponse
 from fastapi.routing import APIRoute as FastAPIRoute
+from fastapi.middleware.cors import CORSMiddleware
 
 from .endpoints import Endpoints, ModelRepositoryEndpoints
 from .requests import Request
@@ -100,5 +101,16 @@ def create_app(
         default_response_class=Response,
         exception_handlers=_EXCEPTION_HANDLERS,  # type: ignore
     )
+
+    if settings.cors_settings is not None:
+        app.add_middleware(
+            CORSMiddleware,
+            allow_origins=settings.cors_settings.allow_origins,
+            allow_origin_regex=settings.cors_settings.allow_origin_regex,
+            allow_credentials=settings.cors_settings.allow_credentials,
+            allow_methods=settings.cors_settings.allow_methods,
+            allow_headers=settings.cors_settings.allow_headers,
+            max_age=settings.cors_settings.max_age,
+        )
 
     return app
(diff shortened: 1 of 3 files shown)


mlserver/settings.py
```diff
@@ -38,6 +38,32 @@ class Config:
...
+    """Indicate that cookies should be supported for cross-origin requests"""
+    allow_credentials: Optional[bool] = False
+    """A list of HTTP methods that should be allowed for cross-origin requests"""
+    allow_methods: Optional[List[str]] = ["GET"]
+    """A list of HTTP request headers that should be supported for cross-origin requests"""
+    allow_headers: Optional[List[str]] = []
+    """Indicate any response headers that should be made accessible to the browser"""
+    expose_headers: Optional[List[str]] = []
+    """Sets a maximum time in seconds for browsers to cache CORS responses"""
+    max_age: Optional[int] = 600
```
adriangonz, line 64: nit: would it make sense to group these into a sub-object, e.g. `CORSSettings`? We could replace the `enable_cors` flag for something like this:

```python
class CORSSettings(BaseSettings):
	allow_origins: Optional[List[str]] = []
	...

class Settings(BaseSettings):
	....
	cors_settings: Optional[CORSSettings] = None
```

Since we would have the `cors_settings` as something `None`-able, we could then enable the middleware only when `settings.cors_settings is not None`.

mlserver/rest/app.py
```diff
@@ -101,4 +102,15 @@ def create_app(
         exception_handlers=_EXCEPTION_HANDLERS,  # type: ignore
     )
 
+    if settings.enable_cors:
```
adriangonz, line 105: nit: should the default be to enable CORS? I'm wondering whether it's the kind of thing that users expect to be enabled by default. Having said so, maybe it makes sense to be restrictive by default. We can always point users to settings if they want to enable it.

Yeah, not sure what's best. What are your thoughts @JakeNeyer ?

adriangonz (approved):
Hey @JakeNeyer, thanks for your contribution! This is definitely a must have addition prior to the MLServer `1.0` graduation. 

I've added a couple comments below, where it would be great to get your thoughts. You can treat them as nitpicks though. Besides those points, everything else looks good to go once the tests pass. :+1:

mlserver/settings.py
JakeNeyer: I like that idea @adriangonz, I'll make changes to reflect that

mlserver/rest/app.py
JakeNeyer: My gut tells me to leave it disabled by default. That's typically how I've seen other implementations done. I think it would be tricky to have a sensible defaults for Access-Control-Allow-Origin, and without that sensible default, what's really the point?

adriangonz: Yuuup, makes complete sense. Let's leave it as it is :+1:

adriangonz:
Thanks for addressing those comments @JakeNeyer! This is great work. Once the linter passes, this should be ready to go :+1:

PR #393, opened 2021-11-24
Sherif akoush/quickfix/get names from metadata endpoint

sakoush:
Fetch metadata from v2 endpoint inference model and use the info to construct the internal inference request from alibi explain model.

Also in this pr:
- upgrade to alibi 0.6.2
- bump mlserver version to 0.6.0.dev4
- add KernelShap explainer

### DIFF
--- a/runtimes/alibi-explain/mlserver_alibi_explain/common.py
+++ b/runtimes/alibi-explain/mlserver_alibi_explain/common.py
@@ -15,7 +15,7 @@ from mlserver.types import (
     ResponseOutput,
     InferenceResponse,
     InferenceRequest,
-    Parameters,
+    Parameters, MetadataModelResponse,
 )
 
 EXPLAINER_TYPE_TAG = "explainer_type"
@@ -41,7 +41,7 @@ def convert_from_bytes(output: ResponseOutput, ty: Optional[Type]) -> Any:
         return literal_eval(py_str)
 
 
-# TODO: add retry
+# TODO: add retry and better exceptions handling
 def remote_predict(
     v2_payload: InferenceRequest, predictor_url: str
 ) -> InferenceResponse:
@@ -49,6 +49,13 @@ def remote_predict(
     if response_raw.status_code != 200:
         raise RemoteInferenceError(response_raw.status_code, response_raw.reason)
     return InferenceResponse.parse_raw(response_raw.text)
+
+
+def remote_metadata(url: str) -> MetadataModelResponse:
+    response_raw = requests.get(url)
+    if response_raw.status_code != 200:
+        raise RemoteInferenceError(response_raw.status_code, response_raw.reason)
+    return MetadataModelResponse.parse_raw(response_raw.text)
 
 
 # TODO: this is very similar to `asyncio.to_thread` (python 3.9+),
--- a/runtimes/alibi-explain/mlserver_alibi_explain/common.py
+++ b/runtimes/alibi-explain/mlserver_alibi_explain/common.py
@@ -1,6 +1,7 @@
 import asyncio
 import contextvars
 import functools
+import re
 from asyncio import AbstractEventLoop
 from importlib import import_module
 from typing import Any, Optional, Type, Callable, Awaitable, Union, List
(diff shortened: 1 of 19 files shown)


runtimes/alibi-explain/mlserver_alibi_explain/common.py
```diff
@@ -51,6 +57,19 @@ def remote_predict(
     return InferenceResponse.parse_raw(response_raw.text)
 
 
+def remote_metadata(url: str) -> MetadataModelResponse:
+    """Get metadata from v2 endpoint"""
+    response_raw = requests.get(url)
```
adriangonz, line 62: Not that it matters much, but should we make this call asynchronous?

runtimes/alibi-explain/mlserver_alibi_explain/common.py
```diff
@@ -82,17 +101,49 @@ def import_and_get_class(class_path: str) -> type:
...
+
+    # MLServer does not really care about a correct input name!
+    input_name = _DEFAULT_INPUT_NAME
+    id_name = _DEFAULT_ID_NAME
+    outputs = []
+
+    if metadata is not None:
+        if metadata.inputs:
+            # we only support a big single input numpy
+            input_name = metadata.inputs[0].name
```
adriangonz, line 130: Do we also need to pass on the datatype, shape and parameters from the metadata?

runtimes/alibi-explain/mlserver_alibi_explain/common.py
```diff
@@ -82,17 +101,49 @@ def import_and_get_class(class_path: str) -> type:
...
+        if metadata.outputs:
+            outputs = metadata.outputs
+
     # For List[str] (e.g. AnchorText), we use StringCodec for input
     input_payload_codec = StringCodec if type(input_data) == list else NumpyCodec
     v2_request = InferenceRequest(
+        id=id_name,
```
adriangonz, line 137: I'm not 100% sure, but I think the executor will set this to a unique value. Therefore, should we leave the `id` field empty?

runtimes/alibi-explain/mlserver_alibi_explain/common.py
```diff
@@ -82,17 +101,49 @@ def import_and_get_class(class_path: str) -> type:
...
         parameters=Parameters(content_type=input_payload_codec.ContentType),
         # TODO: we probably need to tell alibi about the expected types to use
         # or even whether it is a probability of classes or targets etc
         inputs=[
             input_payload_codec.encode_request_input(  # type: ignore
-                name="predict", payload=input_data
+                name=input_name, payload=input_data
             )
         ],
+        # set outputs as empty list, this will return everything?
```
adriangonz, line 146: I don't think that the protocol itself defines what should happen, so I guess that the behaviour will depend on the model and model runtime.

runtimes/alibi-explain/mlserver_alibi_explain/common.py
sakoush: thats a good point. having a look at the caller `load`, it has also `load_explainer` that we need  to do async so that the entire function (i.e `load`) is truly async. 

I am tempted however to leave it as a separate ticket?

runtimes/alibi-explain/mlserver_alibi_explain/common.py
sakoush: the type and shape is inferred from the payload. I think we can add a warning it they dont match? I didnt want to get into converting the payload according to the metadata as part of this ticket?

runtimes/alibi-explain/mlserver_alibi_explain/common.py
sakoush: I am not sure about SC, maybe this is the case. However locally it will not work with a None id (for triron case).

runtimes/alibi-explain/mlserver_alibi_explain/common.py
sakoush: actually this comment is not valid anymore as we get that from the metadata.

PR #397, opened 2021-11-26
Add default codec argument to decode_request

adriangonz:
## Changelog

- Add default codec kwarg to `decode_request`
- Extend usage of `decode_request` across most runtimes


mlserver/model.py
```diff
@@ -95,10 +98,20 @@ def decode(
...
+        default_codec: Optional[Union[Type[RequestCodec], RequestCodec]] = None,
+    ) -> Any:
+        decode_inference_request(inference_request, self._settings, self._inputs_index)
+
+        if has_decoded(inference_request):
+            return get_decoded(inference_request)
+
+        if default_codec:
+            return default_codec.decode(inference_request)
+
```
sakoush, line 113: why this logic is not more natural to be hidden in `decode_inference_request`?

otherwise could you add a note what are we doing here? as I am a bit confused what `get_decoded` does without having to look at the implementation?

runtimes/alibi-explain/mlserver_alibi_explain/runtime.py
```diff
@@ -151,12 +154,18 @@ def ready(self, value: bool):
         self._rt.ready = value
 
     def decode(
-        self, request_input: RequestInput, default_codec: Optional[InputCodec] = None
+        self,
+        request_input: RequestInput,
+        default_codec: Optional[Union[Type[InputCodec], InputCodec]] = None,
```
sakoush, line 159: nit: should we have an alias `Union[Type[InputCodec], InputCodec]` as it is being used in many places?

sakoush (approved):
LGTM

mlserver/model.py
adriangonz: I remember there was a reason to have this extra granularity, although I can't say what it was right now. I would leave it as it is for now.

runtimes/alibi-explain/mlserver_alibi_explain/runtime.py
adriangonz: Good point, let me add an alias for that.

Issue #396, opened 2021-11-26
Correct approach to JSON request / response tensor data


iamahern:
I have a legacy model with an existing JSON serialization I am attempting to wrap in MLServe.  Looking at the docs it appears as the intended approach is to use `JSON.dumps()/.loads()` and stuff it as a string / BYTES data.  I have tested and that does work using the REST server, but I could not find anything in the docs to validate that this is the intended usage

In ticket #342, streamlining this use case was called out, but I could not find a corresponding example in the code base.

Questions:
1. Correct approach to handling JSON request response is `JSON.dumps()/.loads()` then stuff it in as BYTES, possibly with content_type hints?
2. Would it be possible for you to include an example where the a custom model `predict()` takes JSON input / output? This is a fairly common use case and will help a lot of teams bootstrap themselves with MLServe before (if possible) moving to a more optimized / tabular format for the data inputs.

adriangonz:
Hey @iamahern , 

That's a very good point. 

At the moment, stuffing it as `BYTES` is the way to go. On top of this, you can also set the content type to `str`, so that MLServer knows that it's a string. Although, when using the REST API this will always be implicitly the case (not so with gRPC). 

Something that we've got on the list is adding a new section to the docs on encoding / decoding common payloads (e.g. tensor, dataframe, image, JSON, etc.) to and from the V2 protocol. Do you think something like this could be useful? Additionally, something else we've been thinking about is extending the current codec functionality to also provide logic to encode a payload to  the V2 protocol (e.g. doing something like `JSONCodec.encode_to_request(my_json_payload)`), which could help ease some of these pain points on the client side.

It would be great to hear your thoughts on both points above!

iamahern:
Thanks @adriangonz for the quick response. So this answers the first of my questions/points. 

For your ideas, I think both the encoding/decoding commons payloads documents and a set of prebuilt codec helpers would be extremely helpful. I would regard the docs as more important as that will be the first point of contact for teams trying to test/adopt MLServer. Once you understand the encoding conceptually, most of the codecs are fairly simple for teams to write themselves.

iamahern:
@adriangonz - created https://github.com/SeldonIO/MLServer/pull/402 to resolve this ticket.

iamahern:
Resolving based on #402

Issue #400, opened 2021-11-29
StringCodec not compliant with dataplane.proto -- gRPC failure


JakeNeyer:
The StringCodec responds with a bytes string instead of an array of byte strings which causes serialization to fail for gRPC requests:

proto:
```
  // Representation for BYTES data type. The size must match what is
  // expected by the tensor's shape. The contents must be the flattened,
  // one-dimensional, row-major order of the tensor elements.
  repeated bytes bytes_contents = 8;

```

```
    def encode(cls, name: str, payload: List[str]) -> ResponseOutput:
        packed, shape = pack(map(_encode_str, payload))
        return ResponseOutput(
            name=name,
            datatype="BYTES",
            shape=shape,
            data=packed,
        )

```

Throws: 
```
ERROR:grpc._cython.cygrpc:Unexpected [TypeError] raised by servicer method [/inference.GRPCInferenceService/ModelInfer]
Traceback (most recent call last):
  File "src/python/grpcio/grpc/_cython/_cygrpc/aio/server.pyx.pxi", line 680, in grpc._cython.cygrpc._handle_exceptions
  File "src/python/grpcio/grpc/_cython/_cygrpc/aio/server.pyx.pxi", line 794, in _handle_rpc
  File "src/python/grpcio/grpc/_cython/_cygrpc/aio/server.pyx.pxi", line 545, in _handle_unary_unary_rpc
  File "src/python/grpcio/grpc/_cython/_cygrpc/aio/server.pyx.pxi", line 402, in _finish_handler_with_unary_response
  File "/Users/jacobneyer/.pyenv/versions/3.7.9/lib/python3.7/site-packages/mlserver/grpc/servicers.py", line 26, in _inner
    return await f(self, request, context)
  File "/Users/jacobneyer/.pyenv/versions/3.7.9/lib/python3.7/site-packages/mlserver/grpc/servicers.py", line 82, in ModelInfer
    response = ModelInferResponseConverter.from_types(result)
  File "/Users/jacobneyer/.pyenv/versions/3.7.9/lib/python3.7/site-packages/mlserver/grpc/converters.py", line 310, in from_types
    for output in type_object.outputs
  File "/Users/jacobneyer/.pyenv/versions/3.7.9/lib/python3.7/site-packages/mlserver/grpc/converters.py", line 310, in <listcomp>
    for output in type_object.outputs
[...]

PR #402, opened 2021-11-30
Add example for JSON encoding as well as fix docker build command on mac

iamahern:
This pull request provides an example model involving JSON input and output. The example is intended to resolve issue #396.

Additionally, the it was found that the `mlserver build` command is broken on MacOS due to how environment variables are handled. This PR includes a fix for this as well.


docs/examples/custom-json/README.md
```diff
@@ -0,0 +1,307 @@
...
+                    data=[response_bytes],
+                    parameters=types.Parameters(content_type="str")
+                )
+            ]
+        )
+
+    def _extract_json(self, payload: types.InferenceRequest) -> Dict[str, Any]:
+        inputs = {}
+        for inp in payload.inputs:
+            inputs[inp.name] = json.loads("".join(self.decode(inp)))
```
adriangonz, line 71: Would it make sense to pass the `default_codec` kwarg on the `self.decode()` call to ensure its not dependant on the `content_type` being set?

docs/examples/custom-json/README.md
```diff
@@ -0,0 +1,307 @@
...
+            "server_response": f"Got your request. Hello from the server."
+        }
+        response_bytes = json.dumps(response).encode("UTF-8")
+
+        return types.InferenceResponse(
+            id=payload.id,
+            model_name=self.name,
+            model_version=self.version,
+            outputs=[
+                types.ResponseOutput(
```
adriangonz, line 58: Would it make sense to use the `StringCodec` here to serialise the response back (i.e. removing the `.encode("UTF-8")` call above)? Or is that currently blocked by #400 ?

docs/examples/custom-json/README.md
```diff
@@ -0,0 +1,307 @@
...
+            name="echo_request",
+            shape=[len(inputs_bytes)],
+            datatype="BYTES",
+            data=[inputs_bytes],
+            parameters=types.Parameters(content_type="str")
+        )
+    ]
+)
+
+inference_request_g = converters.ModelInferRequestConverter.from_types(
```
adriangonz, line 191: Nice find! 

We'd never meant for the converters to be used publicly, but after looking at this, I can see that they could make for an easier dev experience compared to using the protobufs directly.

docs/examples/custom-json/README.md
```diff
@@ -0,0 +1,307 @@
...
+)
+
+grpc_channel = grpc.insecure_channel("localhost:8081")
+grpc_stub = dataplane.GRPCInferenceServiceStub(grpc_channel)
+
+response = grpc_stub.ModelInfer(inference_request_g)
+response
+```
+
+## Deployment
```
adriangonz, line 204: To ensure the example is kept as simple as possible, should we remove the `## Deployment` section altogether? There's now a dedicated page for this, so I'm wondering whether we should consider that outside of the scope of the example (i.e. correct treatment of arbitrary JSON payloads on a custom wrapper).

mlserver/cli/build.py
```diff
@@ -43,7 +43,8 @@ def build_image(folder: str, dockerfile: str, image_tag: str) -> str:
         dockerfile_path = write_dockerfile(tmp_dir, dockerfile)
 
         build_cmd = f"docker build {folder} -f {dockerfile_path} -t {image_tag}"
-        build_env = {"DOCKER_BUILDKIT": "1"}
+        build_env = os.environ.copy()
+        build_env["DOCKER_BUILDKIT"] = "1"
```
adriangonz, line 47: Oh wow, I'm guessing this is the fix for MacOS? That **must have not** been easy to find :grimacing: 

Thanks a lot for debugging and fixing this one! Across the team we use mainly Linux, so this type of issues often get undiscovered. Therefore, these kind of contributions are always extra appreciated :+1:

adriangonz (commented):
Nice one @iamahern ! Thanks a lot for spending the time on putting this one together. 

I've added a couple minor comments down there, but this looks really good. This will be extra helpful for the next user who's looking to send arbitrary JSON payloads. It's always great to see these contributions from the community!

docs/examples/custom-json/README.md
iamahern: Yes - this was is currently blocked by issue #400. It will work fine via the REST API endpoint, but breaks if I use the gRPC endpoint.

docs/examples/custom-json/README.md
iamahern: Yes, this is essentially copy and pasted from the other custom model example. I'll remove this section.

mlserver/cli/build.py
iamahern: Not too difficult to find. I think you would have rolled out 0.6.1 the next day once the rest of the community got access to the feature.

iamahern:
@adriangonz - Should be good for another round of checks. Made the changes requested along with comments as to why I didn't use the JsonCodec in one case. Think I resolved the linting issues initially flagged.

docs/examples/custom-json/README.md
adriangonz: Got it, thanks for clarifying that point :+1:

adriangonz (approved):
Changes look great! I think once the CI checks pass, this PR should be good to go :+1: 

Thanks a lot for this work @iamahern !

adriangonz:
BTW It seems that the test failure is due to a recent update on either MLflow or PyTorch (or both). This can be replicated on `master`, so totally out of the scope of this PR.

The linter error seems to complain just about a couple missing trailing commas. If you can add those, we can get this one merged.

iamahern:
Comments added.

PR #404, opened 2021-12-01
Build MLflow PyTorch model on the fly

adriangonz:
Fixes issue with previously saved artifact after the latest MLflow release.


sakoush (approved):
LGTM

PR #405, opened 2021-12-01
Pin nest_asyncio to 1.5.1

adriangonz:
It seems that `nest_asyncio==1.5.2` makes the tests fail with the following error (see https://github.com/erdewit/nest_asyncio/issues/62 for more details):

```
runtimes/alibi-explain/tests/conftest.py:29: in <module>
    nest_asyncio.apply()
../../.virtualenvs/mlserver-37/lib/python3.7/site-packages/nest_asyncio.py:14: in apply
    raise ValueError('Can\'t patch loop of type %s' % type(loop))
E   ValueError: Can't patch loop of type <class 'NoneType'>
```

With `nest_asyncio==1.5.1`, tests still seem to pass, so pinning it for now until we find out more about it.

### DIFF
--- a/runtimes/alibi-explain/requirements-dev.txt
+++ b/runtimes/alibi-explain/requirements-dev.txt
@@ -1,3 +1,3 @@
 numpy
-nest_asyncio
+nest_asyncio==1.5.1
 types-requests


sakoush (approved):
LGTM

adriangonz:
The only error will be fixed by #404, so we can merge this one.

PR #406, opened 2021-12-03
Sherif akoush/quickfix/explainer output

sakoush:
add a custom endpoint `/explain` for alibi runtime for backward compatibility with legacy downstream users.

### DIFF
--- a/runtimes/alibi-explain/mlserver_alibi_explain/runtime.py
+++ b/runtimes/alibi-explain/mlserver_alibi_explain/runtime.py
@@ -51,7 +51,7 @@ class AlibiExplainRuntimeBase(MLModel):
         self.alibi_explain_settings = explainer_settings
         super().__init__(settings)
 
-    async def explain_v1_output(self, request: Request) -> Response:
+    async def explain_v1_output(self, request: InferenceRequest) -> Response:
         """
         A custom endpoint to return explanation results in plain json format (no v2
         encoding) to keep backward compatibility of legacy downstream users.
@@ -59,11 +59,7 @@ class AlibiExplainRuntimeBase(MLModel):
         This does not work with multi-model serving as no reference to the model exists
         in the endpoint.
         """
-        # convert raw to v2
-        raw_data = await request.body()
-        payload = InferenceRequest.parse_raw(raw_data)
-
-        v2_response = await self.predict(payload)
+        v2_response = await self.predict(request)
         explanation = json.loads(v2_response.outputs[0].data.__root__)
 
         return Response(content=explanation, media_type="application/json")
@@ -201,5 +197,5 @@ class AlibiExplainRuntime(MLModel):
     # we add _explain_v1_output here to enable the registration and routing of custom
     # endpoint to `_rt.explain_v1_output`
     @custom_handler(rest_path="/explain")
-    async def _explain_v1_output(self, request: Request) -> Response:
+    async def _explain_v1_output(self, request: InferenceRequest) -> Response:
         return await self._rt.explain_v1_output(request)
--- a/runtimes/alibi-explain/mlserver_alibi_explain/runtime.py
+++ b/runtimes/alibi-explain/mlserver_alibi_explain/runtime.py
@@ -61,7 +61,7 @@ class AlibiExplainRuntimeBase(MLModel):
         """
         # convert raw to v2
         raw_data = await request.body()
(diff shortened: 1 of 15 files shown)


runtimes/alibi-explain/mlserver_alibi_explain/runtime.py
```diff
@@ -46,6 +51,24 @@ def __init__(
         self.alibi_explain_settings = explainer_settings
         super().__init__(settings)
 
+    @custom_handler(rest_path="/explain_v1_output")
```
axsaucedo, line 54: This coudl actually just be changed to be `/explain`, as in theory it would then align with the path that the v1 sever would use (as opposed to the actual v2 compliant which would be `/.../infer`

runtimes/alibi-explain/mlserver_alibi_explain/runtime.py
```diff
@@ -1,7 +1,11 @@
+import json
 from typing import Any, Optional, List, Dict
 
+import orjson
```
adriangonz, line 4: Now that `orjson` is an optional dependency on MLServer, would it make sense to add it as a required dep of the `mlserver-alibi-explain` package to ensure it's always there? We did a similar thing on `mlserver-alibi-detect` BTW.

runtimes/alibi-explain/mlserver_alibi_explain/runtime.py
```diff
@@ -46,6 +51,23 @@ def __init__(
         self.alibi_explain_settings = explainer_settings
         super().__init__(settings)
 
+    async def explain_v1_output(self, request: Request) -> Response:
```
adriangonz, line 54: It should be possible to replace `Request` there for `InferenceRequest`, and then the decoding will happen automagically.

runtimes/alibi-explain/mlserver_alibi_explain/runtime.py
```diff
@@ -46,6 +51,23 @@ def __init__(
...
+        in the endpoint.
+        """
+        # convert raw to v2
+        raw_data = await request.body()
+        payload = InferenceRequest.parse_raw(raw_data)
+
+        v2_response = await self.predict(payload)
+        explanation = orjson.dumps(json.loads(v2_response.outputs[0].data.__root__))
+
+        return Response(content=explanation, media_type="application/json")
```
adriangonz, line 69: As `Response`, we could use an `mlserver.responses.Response` object (or an ORJSONResponse object from FastAPI directly, given that `orjson` will be present). This should take care of the extra `orjson.dumps` step.

runtimes/alibi-explain/tests/conftest.py
```diff
@@ -203,6 +203,70 @@ async def integrated_gradients_runtime() -> AlibiExplainRuntime:
     return rt
 
 
+@pytest.yield_fixture
+async def dummy_alibi_explain_client(tmp_path, settings) -> TestClient:
```
adriangonz, line 207: nit: would it make sense to try to split this fixture into smaller ones?

runtimes/alibi-explain/mlserver_alibi_explain/runtime.py
sakoush: done

runtimes/alibi-explain/mlserver_alibi_explain/runtime.py
sakoush: done

runtimes/alibi-explain/mlserver_alibi_explain/runtime.py
sakoush: done

runtimes/alibi-explain/tests/conftest.py
sakoush: created a build rest server helper function for now as we do not use the inner objects outside of the function.

runtimes/alibi-explain/mlserver_alibi_explain/runtime.py
sakoush: done

runtimes/alibi-explain/mlserver_alibi_explain/runtime.py
sakoush: done

PR #409, opened 2021-12-10
Fix gRPC encoding for BYTES payloads

### LINKED ISSUE #400: StringCodec not compliant with dataplane.proto -- gRPC failure

adriangonz:
Fixes #400

## Changelog
- Change current codec logic so that `BYTES` payloads don't get packed into a single bytes array, but instead is treated as a list of multiple bytes arrays.
- Add tests to ensure codecs work properly in both gRPC and REST

### DIFF
--- a/runtimes/alibi-explain/mlserver_alibi_explain/common.py
+++ b/runtimes/alibi-explain/mlserver_alibi_explain/common.py
@@ -33,7 +33,7 @@ EXPLAIN_PARAMETERS_TAG = "explain_parameters"
 
 
 #  TODO: add this utility in the codec.
-def convert_from_bytes(output: ResponseOutput, ty: Optional[Type]) -> Any:
+def convert_from_bytes(output: ResponseOutput, ty: Optional[Type] = None) -> Any:
     """
     This utility function decodes the response from bytes string to python object dict.
     It is related to decoding StringCodec
--- a/runtimes/alibi-explain/mlserver_alibi_explain/common.py
+++ b/runtimes/alibi-explain/mlserver_alibi_explain/common.py
@@ -37,10 +37,15 @@ def convert_from_bytes(output: ResponseOutput, ty: Optional[Type]) -> Any:
     This utility function decodes the response from bytes string to python object dict.
     It is related to decoding StringCodec
     """
+    if output.shape != [1]:
+        # TODO: Does this default value make sense here?
+        return ""
+
     if ty == str:
-        return bytearray(output.data).decode("UTF-8")
+        return bytearray(output.data[0]).decode("UTF-8")
     else:
-        py_str = bytearray(output.data).decode("UTF-8")
+        py_str = bytearray(output.data[0]).decode("UTF-8")
+
         from ast import literal_eval
 
         return literal_eval(py_str)
--- a/runtimes/alibi-explain/mlserver_alibi_explain/common.py
+++ b/runtimes/alibi-explain/mlserver_alibi_explain/common.py
@@ -11,7 +11,6 @@ import requests
 from pydantic import BaseSettings
 
 from mlserver.codecs import StringCodec, NumpyCodec
-from mlserver.errors import RemoteInferenceError
 from mlserver.types import (
     ResponseOutput,
     InferenceResponse,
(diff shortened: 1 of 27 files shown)


runtimes/alibi-explain/mlserver_alibi_explain/runtime.py
```diff
@@ -58,7 +58,12 @@ async def explain_v1_output(self, request: InferenceRequest) -> Response:
         in the endpoint.
         """
         v2_response = await self.predict(request)
-        explanation = json.loads(v2_response.outputs[0].data.__root__)
+        output = v2_response.outputs[0]
+
+        # TODO: Does this default value make sense here?
+        explanation = {}
```
sakoush, line 64: perhaps we should raise an exception if we cannot recover the explanation dict?

runtimes/alibi-explain/mlserver_alibi_explain/common.py
```diff
@@ -37,10 +37,15 @@ def convert_from_bytes(output: ResponseOutput, ty: Optional[Type]) -> Any:
     This utility function decodes the response from bytes string to python object dict.
     It is related to decoding StringCodec
     """
+    if output.shape != [1]:
+        # TODO: Does this default value make sense here?
+        return ""
```
sakoush, line 42: should we raise an exception?

sakoush (approved):
Lovely :) with regards to alibi runtime I feel that if the response is not according to spec that perhaps we could raise an exception instead of return empty dict that could confuse the consumer?

runtimes/alibi-explain/mlserver_alibi_explain/runtime.py
adriangonz: Good point! Let me do that :+1:

PR #415, opened 2021-12-21
Initial changes to allow loading of multiple versions

### LINKED ISSUE #433: Support version policy in MMS
Support version policies properly when loading models with multiple versions available. To start with, we could take a similar approach to how Triton handles it.

cliveseldon:
Fixes #433 

## Changelog

- After loading a new model, refresh existing ones to ensure stale models (i.e. loaded, but not present in the repo anymore) get unloaded.
- When (re)loading an existing model (i.e. same name and version, but potentially different artifacts), unload the previous instance of the model.
- Ensure that default model (i.e. reached when there's no version in the URI path) points to the "latest" version, where latest is the highest integer version (if all versions are integers), or the lexicographically highest version (if all versions are not integers).
- Allow models to be loaded from folders named after the model's version (note that this won't change the model version!!).

### DIFF
--- a/mlserver/cli/main.py
+++ b/mlserver/cli/main.py
@@ -37,10 +37,10 @@ async def start(folder: str):
     """
     Start serving a machine learning model with MLServer.
     """
-    settings, models = await load_settings(folder)
+    settings, models_settings = await load_settings(folder)
 
     server = MLServer(settings)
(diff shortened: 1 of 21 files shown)


cliveseldon:
@cliveseldon Looks good after disucssion

Issue #416, opened 2021-12-22
Dataframe model outputs?


joerunde:
Yo @adriangonz!

I had previously added support for dataframe input to SKLearn runtimes here: https://github.com/SeldonIO/MLServer/pull/379

Our users are supper happy with that 👍 , and now we've been asked about sort of the opposite: supporting dataframes as an output from an SKLearn model.

We have users with something like this as the last stage of an SKLearn pipeline:
```
class SuperModelThing():
...
    def predict(self, X):
        # Some cool prediction thing
        ...
        output = pandas.DataFrame()
        output["object"] = "Apple"
        output["object_probability"] = 0.79
        output["is_vegetable"] = 1
        output["is_vegetable_probability"] = 0.42
        return output
```

The SKLearn runtime doesn't currently support encoding a response from a data frame, and it's made more awkward by the fact that it also assumes each item in the outputs is a response from calling one function on the model. 
I get that fits the normal usage paradigm- you have a model that returns an array of labels from `predict()` and an array of probabilities of those labels from `predict_proba()`, so I'm wondering how best to proceed here.

1. We could ask the user to adapt their code to use `predict` and `predict_proba` as expected...
```
class SuperModelThing():
...
    def predict(self, X):
        ...
        object = "Apple"
        is_vegetable = 1
        
        return np.array([
            class_name_to_int(object),
            is_vegetable
        ])

        
    def predict_proba(self, X):
        object_proba = 0.79
        is_vegetable_proba = 0.42

        return np.array([
            object_proba,
            is_vegetable_proba
        ])
```
But then they'd be responsible for more mapping on the caller side, they'd need to understand which index in the output corresponds to which feature and undoing any transformations like going from class name -> integer code -> class name.

2. We could ask the user to expose each output feature as a callable function on the model:
```
class SuperModelThing():
...
    def object(self, X):
        ...

    def object_proba(self, X):
        ...

    def is_vegetable(self, X):
        ...

    def is_vegetable_proba(self, X):
        ...
```
...But then I'm not quite sure how you'd go about actually implementing that in a stateless and threadsafe manner without re-running the underlying models for each output, which seems like a non-starter.

3. We could modify the SKLearn runtime to include some oddball custom logic like:
```
if len(payload.outputs == 1) and (payload.outputs[0].name = PREDICT_OUTPUT):
    output = self.model.predict(X)
    if isinstance(output, pd.DataFrame):
        return PandasCodec.encode(output)
```
[...]

PR #425, opened 2022-01-05
Return DataFrames from SKLearn models

joerunde:
For https://github.com/SeldonIO/MLServer/issues/416

Still kinda WIP- we're focusing on getting this working internally at the moment.

This PR splits out the `_predict_outputs` logic into two functions, `_get_model_outputs` to run the model and `_encode_outputs` to encode the results.

This allows us to easily check if the list of model outputs contains just a single dataframe, and use the dataframe codec to encode that as the entire output. Otherwise, we roll through the outputs and use the numpy codec to encode each one, as `_predict_outputs` does currently.

If multiple DataFrames are output, an error will be thrown

### DIFF
--- a/runtimes/sklearn/mlserver_sklearn/sklearn.py
+++ b/runtimes/sklearn/mlserver_sklearn/sklearn.py
@@ -7,7 +7,6 @@ from pandas import DataFrame
 from sklearn.pipeline import Pipeline
 
 from mlserver import types
-from mlserver.codecs.base import find_input_codec, find_request_codec
 from mlserver.model import MLModel
 from mlserver.errors import InferenceError
 from mlserver.types import ResponseOutput, RequestOutput
@@ -95,7 +94,7 @@ class SKLearnModel(MLModel):
 
         return outputs
 
-    def _encode_outputs(self, outputs: List[Tuple[RequestOutput, object]]) -> List[types.ResponseOutput]:
+    def _encode_outputs(self, outputs: List[Tuple[RequestOutput, object]]) -> List[ResponseOutput]:
         response_outputs = []
 
         all_output_names = [o[0].name for o in outputs]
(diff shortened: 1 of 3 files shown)


adriangonz (approved):
Really good work as usual @joerunde ! Thanks a lot for this one!

I think is is a great first stab at passing the outputs from models through codecs. Once the tests pass, it should be good to go. Is there anything that you feel like would be worth adding before merging @joerunde ?

joerunde:
@adriangonz Sweet, glad you're cool with the first pass in general here!

There's one py37 compatibility bug with the test setup that I need to fix, but also I was looking around at how the MLFlow runtime has a separate `encoding.py` file with nicer type hinting set up. I'd like to take a couple minutes to kinda follow that pattern here, maybe that will also help later if/when we try to push this logic down to the codecs themselves like you mentioned.

joerunde:
Alright I think I'm happier with this now that it's not just throwing tuples around with tons of type hint warnings :D 
I think I've also fixed the py37 testing error, but I'll have to go set up a new venv locally to double check.

Also, if it's not too much to ask @adriangonz I promise not to abuse the build system if you let the checks run automatically on my commits!

joerunde:
Ah, formatting errors and an import conflict :(

I should have known better

adriangonz:
Hey @joerunde , 

Apologies for the approval required for checks! Just noticed now that we had the repo's set to the most strict option. Hopefully now the CI pipelines should run automatically for existing contributors.

adriangonz (approved):
That looks great @joerunde ! Thanks a lot for this last effort on cleaning it up further. 

Let's wait for the tests to pass, and then it should be good to go :+1:

joerunde:
Ah shoot, I was unaware that `make fmt` doesn't reformat comment lines to shorten them up. :/

adriangonz:
Looks good to go @joerunde ! Thanks again for a great contribution! :+1: :rocket:

PR #432, opened 2022-01-06
Add support to manipulate headers in inference runtimes

### LINKED ISSUE #257: Support for headers in InferenceRequest and InferenceResponse

adriangonz:
Fixes #257 

## Changelog
- Wire the REST headers and gRPC metadata to the `parameters.headers` field of the request / response. 

## TODO
- [x] Document how headers can be used in custom runtimes.


axsaucedo (approved):
Nice one looks great - the only thing I can think of is perhaps on docs to add explicitly the example returning an InferenceResponse object

adriangonz:
@axsaucedo good point, done :+1:

PR #438, opened 2022-01-10
Add docs section on content types

### LINKED ISSUE #283: Document input formats and codecs

adriangonz:
Fixes #283


docs/user-guide/content-type.md
```diff
@@ -0,0 +1,232 @@
...
+example](../examples/content-type/README.md).
+
+```{note}
+It's important to keep in mind that content types can be specified at both the
+**request level** and the **input level**.
+The former will apply to the **entire set of inputs**, whereas the latter will
+only apply to a **particular input** of the payload.
+```
+
+## Available Content Types
```
sakoush, line 97: should we have a note about the default `content_type` if it is not specified by the user?

docs/user-guide/content-type.md
```diff
@@ -0,0 +1,232 @@
...
+The `base64` content type will decode a binary V2 payload into a Base64-encoded
+string (and viceversa), taking into account the following:
+
+- The expected `datatype` is `BYTES`.
+- The `data` field should contain the **raw bytes** (i.e. before encoding them
+  into Base64).
+- The `shape` field represents the number of binary strings that are encoded in
+  the payload.
+
+### Datetime
```
sakoush, line 221: can we give more details why this is required as a separate type as opposed to be part of `np` datatype? is it because it will come serialised / efficient and easier to decode?

docs/user-guide/content-type.md
```diff
@@ -0,0 +1,232 @@
...
+This can be used as a helper for models which only expect a single string or of
+set of strings.
+
+### Base64
+
+The `base64` content type will decode a binary V2 payload into a Base64-encoded
+string (and viceversa), taking into account the following:
+
+- The expected `datatype` is `BYTES`.
+- The `data` field should contain the **raw bytes** (i.e. before encoding them
```
sakoush, line 216: whats the reason it doesnt come in base64? it is because json will serialise/deserialise anyway in base64?

docs/user-guide/content-type.md
```diff
@@ -0,0 +1,232 @@
...
+
+The `np` content type will decode / encode V2 payloads to a NumPy Array, taking
+into account the following:
+
+- The `datatype` field will be matched to the closest [NumPy
+  `dtype`](https://numpy.org/doc/stable/reference/arrays.dtypes.html).
+- The `shape` field will be used to reshape the flattened array expected by the
+  V2 protocol into the expected tensor shape.
+
+When using the NumPy Array content type at the request-level, it will decode
```
sakoush, line 129: give example?

docs/user-guide/content-type.md
```diff
@@ -0,0 +1,232 @@
...
+
+### Base64
+
+The `base64` content type will decode a binary V2 payload into a Base64-encoded
+string (and viceversa), taking into account the following:
+
+- The expected `datatype` is `BYTES`.
+- The `data` field should contain the **raw bytes** (i.e. before encoding them
+  into Base64).
+- The `shape` field represents the number of binary strings that are encoded in
```
sakoush, line 218: give example?

docs/user-guide/content-type.md
```diff
@@ -0,0 +1,232 @@
...
+  the payload.
+
+### Datetime
+
+The `datetime` content type will decode a V2 input into a [Python
+`datetime.datetime`
+object](https://docs.python.org/3/library/datetime.html#datetime.datetime),
+taking into account the following:
+
+- The expected `datatype` is `BYTES`.
```
sakoush, line 228: give example?

sakoush (approved):
Left some clarification comments.

docs/user-guide/content-type.md
adriangonz: The default content type is dependant on each runtime, but maybe we can just mention that instead of listing all the defaults?

docs/user-guide/content-type.md
adriangonz: Do you mean the `datetime` type? There are cases where datetimes have to be decoded without a NumPy array, for example when they are part of a Pandas DataFrame (as in, one of the columns represents a date). This is also required for the MLflow runtime, which allows that type on their previous server.

docs/user-guide/content-type.md
adriangonz: This is mainly thinking of protobuf, where sending the payload as bytes would be more efficient. You're right though in that, when using JSON, probably sending the actual base64 could be better.

Having said that, I'd say that changing the codecs is outside the scope of this PR, which only aims to document them.

docs/user-guide/content-type.md
sakoush: a reference to where uses can get this information is fine as it seems it might affect how they construct the payload?

docs/user-guide/content-type.md
adriangonz: Just had a look, and it actually expects the base64-encoded arrays, and then it decodes them into binaries.

docs/user-guide/content-type.md
adriangonz: Done

docs/user-guide/content-type.md
```diff
@@ -0,0 +1,232 @@
...
+Machine learning models generally expect their inputs to be passed down as a
+particular Python type.
+Most commonly, this type ranges from _"general purpose"_ NumPy arrays or Pandas
+DataFrames to more granular definitions, like `datetime` objects, `Pillow`
+images, etc.
+Unfortunately, the definition of the [V2 Inference
+Protocol](https://kserve.github.io/website/modelserving/inference_api/) doesn't
+cover any of the specific use cases.
+This protocol can be thought of a wider _"lower level"_ spec, which only
+defines what fields should a payload have.
```
edshee, line 12: "fields a payload should have"

docs/user-guide/content-type.md
```diff
@@ -0,0 +1,232 @@
...
+which offer a way to let MLServer know how it should _"decode"_ V2-compatible
+payloads.
+When shaped in the right way, these payloads should _"encode"_ all the
+information required to extract the higher level Python type that will be
+required for a model.
+
+To illustrate the above, we can think of a Scikit-Learn pipeline, which takes
+in a Pandas DataFrame and returns a NumPy Array.
+Without the use of **content types**, the V2 payload itself would probably lack
+information about how should this payload be treated by MLServer.
```
edshee, line 24: "about how this payload should be treated by MLServer."

docs/user-guide/content-type.md
```diff
@@ -0,0 +1,232 @@
...
+| First Name | Age |
+| ---------- | --- |
+| Joanne     | 34  |
+| Michael    | 22  |
+
+This table, could be specified in the V2 protocol as the following payload, where we declare that:
+
+- The whole set of inputs should be decoded as a Pandas Dataframe (i.e. setting
+  the content type as `pd`).
+- The Age column should be decoded as a UTF-8 string (i.e. setting the content
```
edshee, line 58: Did you mean "The First name column"? The Age column is decoded as an integer

docs/user-guide/content-type.md
```diff
@@ -0,0 +1,232 @@
...
+Python string, taking into account the following:
+
+- The expected `datatype` is `BYTES`.
+- The `shape` field represents the number of "strings" that are encoded in
+  the payload (e.g. the `["hello world", "one more time"]` payload will have a
+  shape of 2 elements).
+
+When using the `str` content type at the request-level, it will decode the
+entire request by considering only the first `input` element.
+This can be used as a helper for models which only expect a single string or of
```
edshee, line 294: "or a" set of strings

edshee (approved):
Very detailed and informative, approving assuming a couple of typos get fixed

docs/user-guide/content-type.md
adriangonz: Good catch! :+1:

edshee (approved):

PR #446, opened 2022-01-14
Sherif akoush/quickfix/fix alibi import test

sakoush:
pin `numpy` to `<1.22` as `numba 0.55.0` (alibi runtime `KernelShap`) requires it.


setup.py
```diff
@@ -38,7 +38,8 @@ def _load_description() -> str:
         "fastapi",
         "grpcio",
         "importlib-metadata;python_version<'3.8'",
-        "numpy",
+        # numba 0.55.0 requires: numpy <1.22 (for Shap / alibi)
+        "numpy==1.21.5",
```
adriangonz, line 42: Just wondering, would it work if we pinned the version of `numpy` in the dependencies of the Alibi runtime package?

sakoush: I didnt work for some reason but let me double check.

PR #450, opened 2022-01-17
Fix broken custom runtimes link

edshee:


README.md
```diff
@@ -65,7 +65,7 @@ you interact with a subset of common frameworks.
 This allows you to start serving models saved in these frameworks straight
 away.
 However, it's also possible to **write [custom
-runtimes](./docs/runtimes/custom)**.
```
adriangonz, line 68: Great spot!

I think this should be pointing to https://github.com/SeldonIO/MLServer/blob/master/docs/runtimes/custom.md instead, which is meant as the "homepage" for custom runtimes.

adriangonz (commented):
Nice catch @edshee !

README.md
edshee: Yep you're right, I linked the examples one instead

edshee:
Leaving for you to merge just for the sake of process :nerd_face:

PR #464, opened 2022-02-01
Install custom Python packages sytem-wide

### LINKED ISSUE #388: Install requirements system-wide when building custom images
Currently, custom dependencies coming from a `requirements.txt` get installed on the base environment for user `1000` (i.e. the default non-`root` user in the image). Therefore, if the image is loaded with a different user ID, these custom dependencies are not available.

To work around this, we should install these custom deps under `root`, so that they are installed system-wide and accessible by any user.

adriangonz:
Fixes #388 

## Changelog
- Install custom `requirements.txt` under `root` to ensure custom packages are available for every user at runtime (we can't know in advance what will be the user ID at runtime).
- Update custom runtime example and docs to reflect that setting the `securityContext` of the `SeldonDeployment` manifest is no longer needed
- Fix SKLearn runtime and ensure each runtime linter stops the build if there are failures


mlserver/cli/constants.py
```diff
@@ -46,7 +46,13 @@
     ./requirements.tx[t] \\
     .
 
-RUN ./hack/build-env.sh . ./envs/base
+USER root
+# Install dependencies system-wide, to ensure that they are available for every
+# user
+RUN ./hack/build-env.sh . ./envs/base && \
+    chown -R 1000:0 ./envs/base && \\
```
sakoush, line 53: is there a reason why we need `chown -R 1000:0 ./envs/base` ?

runtimes/sklearn/mlserver_sklearn/sklearn.py
```diff
@@ -83,7 +83,7 @@ def _get_model_outputs(
         for request_output in payload.outputs:  # type: ignore
             predict_fn = getattr(self._model, request_output.name)
             y = predict_fn(decoded_request)
-            payload = SKLearnPayload(requested_output=request_output, model_output=y)
-            outputs.append(payload)
+            output = SKLearnPayload(requested_output=request_output, model_output=y)
```
sakoush, line 86: were these just mypy errors?

sakoush (approved):
LGTM

runtimes/sklearn/mlserver_sklearn/sklearn.py
adriangonz: Yes, exactly. The `payload` variable was already used above with a different type, and `mypy` didn't like that.

mlserver/cli/constants.py
adriangonz: Do you mean besides the `chmod`? Good point - maybe it would be enough with the `chmod`... I just mimicked what we already have in the main `Dockerfile`. I didn't to try my luck too much with permissions :grimacing:

PR #477, opened 2022-02-10
Adding play icon and link to intro video on splash image

edshee:
Having looked at the potential ways of embedding a video, it seems the best approach for now is just to keep it as a clickable image. I'll have a look at using the ..include directive to have an embedded video for the readthedocs version and open a separate PR for it.


adriangonz (approved):
Looks great @edshee ! :+1: :rocket:

PR #485, opened 2022-02-14
docs: fix docstrings under mlserver.settings

mert-kirpici:
The docstrings above the class fields used to cause broken documentation.
Apparently sphinx expects the docstrings to be under the declaration
lines.

Signed-off-by: Mert Kırpıcı <smertk@gmail.com>

### DIFF
--- a/mlserver/settings.py
+++ b/mlserver/settings.py
@@ -12,35 +12,35 @@ class CORSSettings(BaseSettings):
     class Config:
         env_prefix = ENV_PREFIX_SETTINGS
 
+    allow_origins: Optional[List[str]] = []
     """
     A list of origins that should be permitted to make
     cross-origin requests. E.g. ['https://example.org', 'https://www.example.org'].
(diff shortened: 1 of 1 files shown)


adriangonz:
Great catch @mert-kirpici ! 

Why would they even do that? For extra context, I was also able to find an example from the `autodoc_pydantic` module which  confirms that this is the right way to document fields:

https://autodoc-pydantic.readthedocs.io/en/stable/users/examples.html

mert-kirpici:
@adriangonz 
> https://autodoc-pydantic.readthedocs.io/en/stable/users/examples.html

I hadn't seen this :) thanks!

Do you think this change is causing the unsuccessful checks? Seems like they're unrelated to me but cannot tell what's going on

adriangonz:
Good point @mert-kirpici, I think they may be caused by a flaky test in there.  I'll re-run them and that should hopefully sort it out. :+1:

PR #488, opened 2022-02-17
More accurate REST/gRPC response status codes for errors

njhill:
Currently MLServer returns a 400 HTTP response or `INVALID_ARGUMENT` gRPC response in all error cases. It would be useful to distinguish between different classes of error, in particular indicating the model-not-found case with a 404 or gRPC `NOT_FOUND` status. Model-mesh uses `NOT_FOUND` to take corrective action i.e. re-load the model in question if necessary.

I kept the default value as 400 for `MLServerError` subclasses that don't provide a specific code, but maybe something else like 500 would be more appropriate?

Note that this is affected by https://github.com/SeldonIO/MLServer/pull/487 but I can fix this one up after that's merged.

### DIFF
--- a/mlserver/errors.py
+++ b/mlserver/errors.py
@@ -32,4 +32,6 @@ class InferenceError(MLServerError):
 
 class ModelParametersMissing(MLServerError):
     def __init__(self, model_name: str):
-        super().__init__(f"Parameters missing for model {model_name}", status.HTTP_400_BAD_REQUEST)
+        super().__init__(
+            f"Parameters missing for model {model_name}", status.HTTP_400_BAD_REQUEST
+        )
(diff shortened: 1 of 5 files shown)


adriangonz:
Hey @njhill, that's a great point.

When we initially implemented this, we decided to go this route due to the lack of details on the V2 Protocol spec (and then come back to it at some point later in time). However, I totally agree that the error codes should be more explicit. At least for the "easy" ones, like the `404` / `Model X not found`.

Do you know whether Triton follows some kind of convention for these? Ideally, the end goal would be to update the V2 spec to ensure everyone aligns with a particular set of error codes.

njhill:
@adriangonz the test failure is only on 3.7 and doesn't appear to be related to these changes?

adriangonz:
Good point @njhill , it seems that we've got a flaky test there (unrelated to this PR as you say). I've kicked off the tests again - they should pass this time.

PR #489, opened 2022-02-20
Fix docs for adaptive batching

dinever:
Signed-off-by: Peixuan Ding <dingpeixuan911@gmail.com>

Hi, Thanks for creating this tool! I've been exploring this over the weekend and found a few mistakes in the docs.

Apparently, the fields for adaptive batching are called `max_batch_...` instead of `maximum_batch_...`, see the code below:

https://github.com/SeldonIO/MLServer/blob/a4ea7cb6d85cc696db2b042c0e7ce0a2950760e8/mlserver/settings.py#L147-L152

Thanks!


adriangonz (approved):
That's a great spot @dinever ! Thanks a lot for correcting those.

This should be good to once we get a green light from the tests. :+1:

PR #504, opened 2022-02-28
Use dtype in the request to create input DataFrame

kspitale:
While testing MLServer in Seldon Core, I discovered a bug due to which the datatype sent in the request input is not used to create the dataframe. Instead it uses the default data types in Pandas which do not include `int32`. This PR proposes a fix for this bug. 

The change has been tested using unit tests and locally by running mlserver. 

Related slack thread - https://seldondev.slack.com/archives/C8Y9A8G0Y/p1645517701847759?thread_ts=1645093614.150229&cid=C8Y9A8G0Y

### DIFF
--- a/mlserver/codecs/numpy.py
+++ b/mlserver/codecs/numpy.py
@@ -30,7 +30,7 @@ _NumpyToDatatype["object"] = "BYTES"
 _NumpyToDatatype["S"] = "BYTES"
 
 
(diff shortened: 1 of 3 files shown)


kspitale:
Fixed indentation in unit tests

adriangonz (approved):
Looks great @kspitale! Massive thanks again for putting the effort to debug, diagnose and fix the issue! :+1:

PR #515, opened 2022-03-15
fix: Fix selective runtime logic in Dockerfile

njhill:
The selective runtime inclusion logic was broken due to missing `mlserver_` prefix in the generated wheel filenames.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@ COPY ./runtimes ./runtimes
 COPY \
     setup.py \
     README.md \
-    .
+    ./
 
 # This will build the wheels and place will place them in the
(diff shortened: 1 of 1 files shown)


adriangonz:
Hey @njhill, 

The main release scripts do seem to take this already into account when calling `docker build`. For example, in the [`./hack/build-images.sh`](https://github.com/SeldonIO/MLServer/blob/master/hack/build-images.sh) script (called by the `make build` target), it will pass the runtimes already prefixed with `mlserver-*`:

https://github.com/SeldonIO/MLServer/blob/master/hack/build-images.sh#L31

Is there any other instance where we're missing this which could lead to issues?

Issue #514, opened 2022-03-15
Adaptive Batching ignores requested output


shane-breeze:
I followed the sklearn example [here](https://mlserver.readthedocs.io/en/latest/examples/sklearn/README.html), using a RandomForestClassifier instead.

I have the following files:
```bash
.
├── model
│   ├── model.joblib
│   └── model-settings.json
└── settings.json
```

I set the adaptive batching options. Then I start the server
```bash
export MLSERVER_MODEL_MAX_BATCH_TIME=1
export MLSERVER_MODEL_MAX_BATCH_SIZE=10
mlserver start .
```

A normal request works as expected, but when I request the `predict_proba` output, I receive the wrong results:

```python
import requests
from sklearn import datasets

digits = datasets.load_digits()
n_samples = len(digits.images)
data = digits.images.reshape((n_samples, -1))
x_0 = data[0:1]

inference_request = {
    "inputs": [
        {
            "name": "predict_proba",
            "shape": x_0.shape,
            "datatype": "FP32",
            "data": x_0.tolist()
        }
    ],
    "outputs": [
        {
            "name": "predict_proba",
            "shape": [-1],
            "datatype": "FP32",
        }
    ]
}

endpoint = "http://localhost:8080/v2/models/model/versions/v0.1.0/infer"
response = requests.post(endpoint, json=inference_request)
print(response.json())
```

Result:
```bash
{'id': '<id>',
 'model_name': 'model',
 'model_version': None,
 'outputs': [{'data': [0],
              'datatype': 'INT64',
              'name': 'predict',
              'parameters': None,
              'shape': [1]}],
 'parameters': None}
```

Expected result:
```bash
{'id': '<id>',
 'model_name': 'model',
 'model_version': 'v0.1.0',
 'outputs': [{'data': [0.96, 0.0, 0.0, 0.0, 0.02, 0.0, 0.0, 0.02, 0.0, 0.0],
              'datatype': 'FP64',
              'name': 'predict_proba',
              'parameters': None,
              'shape': [1, 10]}],
 'parameters': None}
```

adriangonz:
Hey @shane-breeze , 

Thanks for raising this one and for the level of detail. 

Just to test something quickly, could you try disabling the adaptive batching feature? You can do this by setting either of the `MAX_BATCH_*` settings to either `0` or a negative value. This is mainly to pinpoint what may be causing issues here.

shane-breeze:
Hi @adriangonz,

[...]

PR #516, opened 2022-03-16
fix: Adaptive Batching ignoring requested output

shane-breeze:
Address #514.

Parameterise the test fixtures `inference_request` and `inference_response` to test inference requests with outputs.


adriangonz:
Hey @shane-breeze, 

It seems like the GH Actions responsible for running the tests have gotten stuck for some reason (never seen that before!). Could you try pushing a dummy commit to see if that re-triggers them? 

image

shane-breeze:
Hi @adriangonz,

No problem. I can fix the failing lint tests.

PR #518, opened 2022-03-18
Added _parse_dict_values method

dtpryce:
- New method to parse the only BaseSetting that might involve a JSON / dict object 
- Replaces single quotes within value with doubles
- Replaces double quotes around the environment variable with singles
- Tested outputs in a docker image with some custom extra dictionary and starts server

### DIFF
--- a/hack/generate_dotenv.py
+++ b/hack/generate_dotenv.py
@@ -78,7 +78,7 @@ def save_default_env(env: dict, output: str):
 
 
 def _parse_dict_values(name: str, value: str) -> str:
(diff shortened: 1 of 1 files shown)


vanducng:
Seem like same issue #520 with `inputs` & `outputs` as well.

dtpryce:
> Seem like same issue #520 with `inputs` & `outputs` as well.

Yep agreed ... should be fixed in recent update. Happy to take any other suggestions :)

dtpryce:
@adriangonz over to you :)

PR #519, opened 2022-03-19
fix: Correct return MLModel setting for outputs

vanducng:
With current setting, the return meta data for `outputs` is taken from `inputs` one.


adriangonz (approved):
Great spot @vanducng ! Thanks for fixing this one.

PR #529, opened 2022-03-22
add catboost classifier support

theofpa:
add catboost support:

* add a new runtime by heavily copying lightgbm runtime
* add test that builds a model, sends an inference request, validates the prediction
* add example in docs
* list in index/readme


docs/runtimes/index.md
```diff
@@ -23,11 +23,12 @@ class in your `model-settings.json` file.
...
 | Scikit-Learn | `mlserver-sklearn`      | `mlserver_sklearn.SKLearnModel`            | [Scikit-Learn example](../examples/sklearn/README.md)      | [MLServer SKLearn](./sklearn)                                    |
 | XGBoost      | `mlserver-xgboost`      | `mlserver_xgboost.XGBoostModel`            | [XGBoost example](../examples/xgboost/README.md)           | [MLServer XGBoost](./xgboost)                                    |
 | Spark MLlib  | `mlserver-mllib`        | `mlserver_mllib.MLlibModel`                | Coming Soon                                                | [MLServer MLlib](./mllib)                                        |
 | LightGBM     | `mlserver-lightgbm`     | `mlserver_lightgbm.LightGBMModel`          | [LightGBM example](../examples/lightgbm/README.md)         | [MLServer LightGBM](./lightgbm)                                  |
+| CatBoost     | `mlserver-catboost`     | `mlserver_catboost.CatboostModel`          | [CatBoost example](../examples/catboost/README.md)         | [MLServer LightGBM](./lightgbm)                                  |
```
adriangonz, line 31: This line seems to point to the LightGBM one.

runtimes/catboost/mlserver_catboost/catboost.py
```diff
@@ -0,0 +1,35 @@
+from catboost import CatBoostClassifier
+
+from mlserver import types
+from mlserver.model import MLModel
+from mlserver.utils import get_model_uri
+from mlserver.codecs import NumpyCodec, NumpyRequestCodec
+
+
+WELLKNOWN_MODEL_FILENAMES = ["model.bst"]
```
adriangonz, line 9: The "well-known" filenames list is used to try to automatically find the right artifacts when they are not specified explicitly in the `model-settings.json`. We can think of it as a list of "standardised" names. 

Based on the CatBoost example, should `model.cbm` be considered as a well known filename? The `model.bst` seems more relevant to XGBoost, but I'm not very familiar with CatBoost, so it would be great to hear your thoughts!

runtimes/catboost/mlserver_catboost/catboost.py
```diff
@@ -0,0 +1,35 @@
...
+    """
+    Implementation of the MLModel interface to load and serve `catboost` models.
+    """
+
+    async def load(self) -> bool:
+        model_uri = await get_model_uri(
+            self._settings, wellknown_filenames=WELLKNOWN_MODEL_FILENAMES
+        )
+
+        self._model = CatBoostClassifier()
```
adriangonz, line 22: Would there be a need in the future to also support regressor models, etc?

adriangonz (commented):
This looks great @theofpa! Thanks a lot for submitting a new runtime! :rocket: 

It also seems to already tick most of the boxes. I've added a few questions, but it looks almost ready to go!

runtimes/catboost/mlserver_catboost/catboost.py
theofpa: nice catch! addressed in d49909e, searching github for how people name their models, bring the cbm and bin as the two most popular extensions for catboost

runtimes/catboost/mlserver_catboost/catboost.py
theofpa: indeed, we'll have to introduce a mechanism to choose which class to instantiate, so added a TODO for the regressor and ranker!

theofpa:
Thanks for the review @adriangonz . Since we're adding only the classifier capability I've renamed the contribution and added a TODO for the regressor and ranker capabilities as a future improvement.

adriangonz (approved):
Nice one @theofpa! Thanks for addressing those comments.

Really great job! Once the tests are green, this one should be good to go :+1:

adriangonz:
Hey @theofpa, this PR should deal with the `click` / `black` issues that broke off the internet today :grimacing: 

https://github.com/SeldonIO/MLServer/pull/544

It has been merged already, so if you can rebase, all tests should look better :+1:

adriangonz:
Hey @theofpa , 

It seems that the linter is flagging a couple minor changes. The other failing test seems like a non-issue. So if you can update those linter changes, we should be good to go :+1:

adriangonz:
Hey @theofpa ,

Conscious it's been a while since this PR was opened, so we'll close the PR for now. However, if you get a chance to have a look in the near future, it would be great to pick it up again!

PR #531, opened 2022-03-22
Uvicorn logging settings

### LINKED ISSUE #530: Allow to configure uvicorn logging
Trying to add custom handlers to the application logger we have found that the uvicorn ones are not working as expected. It seems that uvicorn overrides the logging configuration when the Config object is initialized.

It would be nice to allow users to configure the uvicorn logger. This can be done by setting the path to a registry configuration file when creating uvicorn's Config object.

pablobgar:
Closes #530 

add support to uvicorn logging setting:

- add a parameter to define a logging configuration file into application settings
- edit uvicorn Config initialization to add this parameter

### DIFF
--- a/mlserver/rest/server.py
+++ b/mlserver/rest/server.py
@@ -53,7 +53,7 @@ class RESTServer:
 
     async def start(self):
         cfg = uvicorn.Config(
-            self._app, host=self._settings.host, port=self._settings.http_port
+            self._app, host=self._settings.host, port=self._settings.http_port, log_config=self._settings.uvicorn_logging_settings
         )
         self._server = _NoSignalServer(cfg)
         await self._server.serve()
--- a/mlserver/settings.py
+++ b/mlserver/settings.py
@@ -89,6 +89,10 @@ class Settings(BaseSettings):
     `None` to disable it
     """
 
+    # Logging settings
+    uvicorn_logging_settings: Optional[str] = None
+    """Path to uvicorn logging config file"""
+
 
 class ModelParameters(BaseSettings):
     """


adriangonz:
Hey @pablobgar ,

Thanks for opening this PR! It's really great to see these community-driven efforts. 

Just so that I understand a bit more about Uvicorn's logging config file, wouldn't those settings apply to every logging handler configured in the process? Or does Uvicorn filter out the handlers and formatters to only apply the ones that it will use? Looking at [this issue](https://github.com/encode/uvicorn/issues/491), I can see that the logging config file seems quite generic, so I'd assume that it gets applied globally?

pablobgar:
Hi @adriangonz,

Thanks for your answer.

You are right, this logging config file is applied globally. Uvicorn applies it by calling python's [logging.config.fileConfig](https://docs.python.org/3/library/logging.config.html#logging.config.fileConfig) which applies it globally.  Anyway, in the file you can specify that the configuration affects only certain loggers, here you can see an [example](https://github.com/encode/uvicorn/issues/491#issuecomment-814048773) of it.

Maybe this is not a good solution, I know it can be quite confusing. We can try to find a way to make a more generic approach to allow users to configure custom handlers on all application loggers.

Any ideas?

adriangonz (approved):
That looks fantastic @pablobgar! I think it makes a lot of sense to apply them globally. Thanks a lot for pushing this one upstream! :rocket: 

Once the tests are green, we should be good to go.

adriangonz:
Alright, it seems like there's a small linter error (an unused import). Once that's resolved, this should good to go ahead @pablobgar :+1: 

```
mlserver/logging.py:19: error: Module has no attribute "config"
```

(no need to worry about the docs error BTW - that was due to an outdated dep and should be fixed already in master)

adriangonz:
Hey @pablobgar ,

It looks like the linter has failed again. If you want to run this locally, it should be possible to use the `make lint` target (which would run both the linter and `mypy`).

pablobgar:
Alright, it seems to have worked now!

I had some troubles with Optional[str] type and the linter... 😅

What are the next steps @adriangonz?

pablobgar:
Hi @adriangonz !!!

Good to see this feature merged!!! When is a release coming out with these changes?

adriangonz:
We're currently working towards the `1.1.0` milestone. You can see on the link below the current progress and what's still left to get there:

https://github.com/SeldonIO/MLServer/projects/2?query=is%3Aopen+sort%3Aupdated-desc

PR #535, opened 2022-03-25
Support null string inputs

kspitale:
This PR adds support for null string input features. For certain models, absence of a string input needs to be treated differently from an empty string input. This change will allow models to distinguish between the two.


adriangonz (approved):
Nice one @kspitale! You hit the right spot with where to add this. Really elegant solution :+1: 

All tests look happy, so I think this is good to go. Thanks a lot for contributing this fix! Great job as usual! :rocket:

Issue #457, opened 2022-01-26
Tag model-specific Prometheus metrics


adriangonz:
Prometheus metrics exposed by MLServer currently don't tag model-specific ones (e.g. request count, or request time), which could be useful to assess how each separate model is performing. It could be useful to tag them with a `model_name` (and `model_version`) label so that they can be traced back.

cliveseldon:
Note Triton has following metrics

```
# HELP nv_inference_request_success Number of successful inference requests, all batch sizes
# TYPE nv_inference_request_success counter
nv_inference_request_success{model="iris",version="1"} 4.000000
# HELP nv_inference_request_failure Number of failed inference requests, all batch sizes
# TYPE nv_inference_request_failure counter
nv_inference_request_failure{model="iris",version="1"} 0.000000
# HELP nv_inference_count Number of inferences performed
# TYPE nv_inference_count counter
nv_inference_count{model="iris",version="1"} 4.000000
# HELP nv_inference_exec_count Number of model executions performed
# TYPE nv_inference_exec_count counter
nv_inference_exec_count{model="iris",version="1"} 4.000000
# HELP nv_inference_request_duration_us Cummulative inference request duration in microseconds
# TYPE nv_inference_request_duration_us counter
nv_inference_request_duration_us{model="iris",version="1"} 16233.000000
# HELP nv_inference_queue_duration_us Cummulative inference queuing duration in microseconds
# TYPE nv_inference_queue_duration_us counter
nv_inference_queue_duration_us{model="iris",version="1"} 319.000000
# HELP nv_inference_compute_input_duration_us Cummulative compute input duration in microseconds
# TYPE nv_inference_compute_input_duration_us counter
nv_inference_compute_input_duration_us{model="iris",version="1"} 569.000000
# HELP nv_inference_compute_infer_duration_us Cummulative compute inference duration in microseconds
# TYPE nv_inference_compute_infer_duration_us counter
nv_inference_compute_infer_duration_us{model="iris",version="1"} 14489.000000
# HELP nv_inference_compute_output_duration_us Cummulative inference compute output duration in microseconds
# TYPE nv_inference_compute_output_duration_us counter
nv_inference_compute_output_duration_us{model="iris",version="1"} 403.000000
```

I don't see any differentiation between grpc and REST.

PaulEdwardBrennan:
Can you explain what it means to "tag" a Prometheus metric? Do you mean tag as per cliveseldon's comment above on the 27th Jan?

[...]

PR #560, opened 2022-04-19
Tag Prometheus metrics with model name and version

PaulEdwardBrennan:
This PR relates to ticket Tag model-specific Prmoetheus metics #457 

https://github.com/SeldonIO/MLServer/issues/457

These changes create a new Prometheus metric for each model inference which is prefixed with "model_infer_request_". This metric is tagged with the model name and the model version.

### DIFF
--- a/mlserver/handlers/dataplane.py
+++ b/mlserver/handlers/dataplane.py
@@ -8,6 +8,17 @@ from ..types import (
 )
 from ..middleware import inference_middlewares
 from ..utils import generate_uuid
+from ..middleware import inference_middlewares
+from ..utils import generate_uuid
+from prometheus_client import (
+    Counter, 
+    Summary,
+)
+
+
+_ModelInferRequestSuccess =  Counter('model_infer_request_success', 'Model infer request success count', ['model', 'version'])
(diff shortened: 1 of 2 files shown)


adriangonz (approved):
Hey @PaulEdwardBrennan ,

Thanks a lot for pushing this one! That's really appreciated. :+1: 

The implementation looks pretty clean BTW. I wasn't aware of all the helpers that come with Prom's Python SDK! That `.count_exceptions()` and that `.time()` are super neat! 

I think this is a great first pass for this one, so once the tests are green, we should be good to go ahead :rocket:

adriangonz:
Alright, looking at the tests results, it seems that there're only some import errors coming from the linter (duplicated imports):

```
./mlserver/handlers/dataplane.py:11:1: F811 redefinition of unused 'inference_middlewares' from line 9
./mlserver/handlers/dataplane.py:12:1: F811 redefinition of unused 'generate_uuid' from line 10
```

Once these are sorted out, we should be good to go ahead :+1: 

(PS: feel free the ignore the docs error - this was due to an outdated dep and should be already fixed in `master`)

adriangonz:
Tests seem to pass, so this one should be good to go ahead now! Massive thanks again @PaulEdwardBrennan! This is a great first stab at improving metrics in MLServer! :rocket:

PR #572, opened 2022-05-05
Use uvloop event loop implementation

lada-kunc:
- Add support for uvloop event loop
implementation that is faster than
standard CPython asyncio:
https://uvloop.readthedocs.io/

- Added auto-detection code with
uvloop used as default wherever possible

- Improved throughput on benchmarking tasks
against baseline by 15-20% in gRPC case,
higher improvements with access logs disabled

### DIFF
--- a/mlserver/parallel/utils.py
+++ b/mlserver/parallel/utils.py
@@ -15,7 +15,8 @@ def configure_inference_pool(settings: Settings):
 
     # Set MP method
     try:
-        multiprocessing.set_start_method("spawn")
+        # force set to True to override the "fork" setting used in pytest
+        multiprocessing.set_start_method("spawn", force=True)
     except RuntimeError:
         # TODO: Log warning saying that mp start method couldn't be set
         pass
--- a/mlserver/parallel/worker.py
+++ b/mlserver/parallel/worker.py
@@ -6,6 +6,7 @@ from multiprocessing import Process, Queue, JoinableQueue
 from concurrent.futures import ThreadPoolExecutor
 
 from ..registry import MultiModelRegistry
+from ..utils import install_uvloop_event_loop
 
 from .messages import ModelUpdateType, ModelUpdateMessage, InferenceResponseMessage
 from .utils import terminate_queue, END_OF_QUEUE
@@ -34,6 +35,7 @@ class Worker(Process):
         return self.__executor
 
     def run(self):
+        install_uvloop_event_loop()
         asyncio.run(self.coro_run())
 
     def __inner_init__(self):
--- a/runtimes/alibi-explain/requirements-dev.txt
+++ b/runtimes/alibi-explain/requirements-dev.txt
@@ -1,3 +1,2 @@
-nest_asyncio==1.5.4
 types-requests
 httpx==0.22.0
--- a/runtimes/alibi-explain/tests/conftest.py
+++ b/runtimes/alibi-explain/tests/conftest.py
@@ -5,7 +5,6 @@ from pathlib import Path
 from typing import AsyncIterable, Dict, Any, Iterable
 from unittest.mock import patch
 
-import nest_asyncio
 import pytest
 import tensorflow as tf
 from alibi.api.interfaces import Explanation
@@ -14,6 +13,7 @@ from fastapi import FastAPI
 from httpx import AsyncClient
 
 from helpers.tf_model import get_tf_mnist_model_uri
+from helpers.run_async import run_async_as_sync
 from mlserver import MLModel
 from mlserver.handlers import DataPlane, ModelRepositoryHandlers
 from mlserver.registry import MultiModelRegistry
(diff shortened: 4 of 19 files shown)


adriangonz (commented):
Nice one @lada-kunc !! Thanks a lot for pushing this one.

We had this one on the "bucket list" for a while, but never got around to trying it out. It will be really interesting to see the improvements with `uvloop`!!

Looking at the changes, would it make sense to always set the event loop to `uvloop`? Would there be any caveats? That would also mean you don't need to read the config synchronously first, which should simplify the implementation quite a lot, right?

lada-kunc:
> Looking at the changes, would it make sense to always set the event loop to uvloop? Would there be any caveats? That would also mean you don't need to read the config synchronously first, which should simplify the implementation quite a lot, right?

@adriangonz yes, that would simplify the implementation a lot! Thanks a lot for the review. I guess there shouldn't be any caveats if I implement it in a way the "auto mode" is done right now. If `uvloop` is available then it is going to be used, if not it falls back to standard asyncio.

Looking at the test results, there is a slightly different issue with `nest_asyncio`. I didn't realize that one of projects test suite is using it for reentrant event loop. Unfortunately `uvloop` can't be patched with `nest_asyncio`. So I will also need to figure out how to solve this issue in alibi-runtime tests.

mlserver/parallel/utils.py
```diff
@@ -15,7 +15,8 @@ def configure_inference_pool(settings: Settings):
 
     # Set MP method
     try:
-        multiprocessing.set_start_method("spawn")
+        # force set to True to override the "fork" setting used in pytest
+        multiprocessing.set_start_method("spawn", force=True)
```
adriangonz, line 19: Nice one! That must have been a hard one to debug. 

I've been trying to find out more info about the `force=` kwarg, but it doesn't seem to be listed in the Python docs. Do you know whether there's any caveat that we should be aware of?

mlserver/parallel/worker.py
```diff
@@ -34,6 +35,7 @@ def _executor(self):
         return self.__executor
 
     def run(self):
+        install_uvloop_event_loop()
```
adriangonz, line 38: Great spot initialising this on the workers as well! :+1:

runtimes/alibi-explain/requirements-dev.txt
```diff
@@ -1,3 +1,2 @@
-nest_asyncio==1.5.4
```
adriangonz, line 1: Massive thanks for taking the time to sort out the issue between `nest_asyncio` and `uvloop`!

runtimes/alibi-explain/tests/conftest.py
```diff
@@ -138,24 +144,11 @@ def mock_metadata(*args, **kwargs):
...
-                except Exception:
-                    loop = asyncio.new_event_loop()
-                    asyncio.set_event_loop(loop)
-                    res = loop.run_until_complete(
-                        custom_runtime_tf.predict(kwargs["v2_payload"])
-                    )
-                    return res
+                return run_async_as_sync(
+                    custom_runtime_tf.predict, kwargs["v2_payload"]
+                )
```
adriangonz, line 151: Good one simplifying this logic. Really appreciated!

setup.py
```diff
@@ -44,6 +50,7 @@ def _load_description() -> str:
         "uvicorn",
         "starlette_exporter",
         "py-grpc-prometheus",
+        "uvloop;" + env_marker_cpython,
```
adriangonz, line 53: Interesting! I didn't know that `uvloop` wasn't supported in Win. Thanks for adding the guard!

mlserver/parallel/utils.py
lada-kunc: > Do you know whether there's any caveat that we should be aware of?

Thanks a lot for the review. Yes, this one was really tough to figure out :) Regarding caveats of `force=True`. It's undocumented. The main caveat is in possible IPC object leaks in case that for example `fork` was set first and multiprocessing actually created new processes and after that you change it to `spawn`. Then cleaning-up those previously (fork-ed) created processes can be impacted. But if you haven't used multiprocessing and it's only resetting the value for the first time, it should be fine.

Probably the cleanest way would be propagating the multiprocessing context for all multiprocessing interactions, i.e.

```
ctx = mp.get_context('spawn')
q = ctx.Queue()
p = ctx.Process(target=foo, args=(q,))
```

then you can have several contexts `ctx` with different start methods in your python program

adriangonz: Got it. Thanks for that extra context @lada-kunc!

Based on that, since MLServer is ran as a standalone process, it should be alright to keep the `force=` kwarg as-is. I agree that using `ctx` would probably be cleaner, but that's out of the scope of this PR. That's a good one to keep in mind though!

PR #573, opened 2022-05-06
Huggingface Transformer Runtime with Optimum Integration

axsaucedo:
Initial PR to introduce Huggingface Tranformer Runtime with Optimum integration:

* [x] Add basic huggingface runtime 
* [x] Support pipeline initialisation with pretrained model & artifact
* [x] Support Optimum optimized models with onnx runtime
* [x] Test with Seldon Core (parameters from model)
* [x] Add documentation

#### Currently it's possible to initialise pipelines with pretrained models and tokenisers via model settings

```json
{
    "name": "gpt2-model",
    "parameters": {
        "extra": {
            "task": "text-generation"
        }
    }
}
```

#### Request is sent as string / bytes:

```python
import requests
inference_request = {
    "inputs": [
        {
          "name": "huggingface",
          "shape": [1],
          "datatype": "BYTES",
          "data": ["this is an input"],
        }
    ]
}
endpoint = "http://localhost:8080/v2/models/gpt2-model/infer"
response = requests.post(endpoint, json=inference_request)
response.json()
```

#### Response provided as string / bytes:

```python
{'model_name': 'gpt2-model',
     'model_version': None,
     'id': '0751f80e-b846-4eb4-a098-a38b05eb3037',
     'parameters': None,
     'outputs': [{'name': 'huggingface',
       'shape': [1],
       'datatype': 'BYTES',
       'parameters': None,
       'data': ['[[{"generated_text": "this is an input on the page, it can be turned on or off at will. When it is turned on the page is open in the browser and you may be able to see the current page\'s content. If not, you can either press"}]]']}]}
```


runtimes/huggingface/mlserver_huggingface/common.py
```diff
@@ -0,0 +1,114 @@
...
+            f"Remote inference call failed with {code}, {reason}", status_code=code
+        )
+
+
+class InvalidTranformerInitialisation(MLServerError):
+    def __init__(self, code: int, reason: str):
+        super().__init__(
+            f"Remote inference call failed with {code}, {reason}",
+            status_code=code,
+        )
```
adriangonz, line 50: Does the runtime use these errors?

runtimes/huggingface/mlserver_huggingface/common.py
```diff
@@ -0,0 +1,114 @@
...
+    task: str = ""
+    pretrained_model: Optional[str] = None
+    pretrained_tokenizer: Optional[str] = None
+    optimum_model: bool = False
+
+
+def import_and_get_class(class_path: str) -> type:
+    last_dot = class_path.rfind(".")
+    klass = getattr(import_module(class_path[:last_dot]), class_path[last_dot + 1 :])
+    return klass
```
adriangonz, line 70: I'm guessing this one may have been copied over from the Alibi runtime, but it's not needed for this one, right?

runtimes/huggingface/mlserver_huggingface/common.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+def parse_parameters_from_env() -> Dict:
```
adriangonz, line 73: Oh wow, is this parsing params from Seldon Core? We'll need to think how to move this one to Seldon Core, so that these parameters get translated into model settings.

runtimes/huggingface/mlserver_huggingface/common.py
adriangonz: Actually, just saw that `InvalidTransformerInitialisation` is used, but looks like the error message was copied over from the other one.

runtimes/huggingface/requirements-dev.txt
```diff
@@ -0,0 +1,2 @@
+nest_asyncio==1.5.4
```
adriangonz, line 1: I think these are only needed for the Alibi tests - should we remove this `requirements-dev.txt`?

adriangonz (approved):
Nice! This looks really good!

Added a couple comments below, but mainly just pointing to places where there's some leftover stuff from the Alibi runtime. Shouldn't block merging the PR.

axsaucedo:
@adriangonz thanks for the comments - they should now have been updated, I'll open another issue with the extra tasks

Issue #576, opened 2022-05-09
HuggingFace Optimum Runtime extensions


axsaucedo:
As a follow-up from the initial PR that introduced HuggingFace Optimum Runtime via #4081 we have identified a set of followup tasks to improve the servers:

* [x] Extend for complex outputs / inputs to use v2 protocol
* [ ] Support loading artifacts from provided model-URI
* [x] Ensure Optimum library is pinned with v1.2 is released (currently on `master`) - solved via https://github.com/SeldonIO/MLServer/pull/580
* [x] Fix with parallel server (currently single server by default)
* [ ] Allow further parameters from pipeline to be provided via config (eg device, or all via kwargs)
* [x] Ensure MLServer liveness probe works before loading first model to avoid timeout
* [x] Update runtime to use SUPPORTED_TASKS from the optimum pipelines module as per https://github.com/huggingface/optimum/issues/172
* [x] Ensure TRANFORMERS_CACHE is respected and remove once updated https://github.com/huggingface/optimum/issues/186

adriangonz:
Moved the remaining bullet points to separate issues:

#1076
#1016

PR #582, opened 2022-05-11
Improve logic for health endpoints

adriangonz:
Part of #576 

## Changelog
- Ensure that the server's health endpoints are exposed before starting to load the initial set of models
- Ensure that models show as "not ready" while they are getting loaded
- To refresh / reload existing models, ensure that we do a "rolling update" (i.e. keep the old version available while we load the new one)


axsaucedo:
Nice one - looks great, merging and adding follow-up PR with async download

PR #584, opened 2022-05-12
Ensure HuggingFace runtime doesn't block asyncio when loading model

axsaucedo:
Fixes part of #576 

* Ensuring the transformer model is downloaded async and cached
* Model then gets loaded without affecting liveness/readiness probes
* Allows for very large transformers to be loaded
* Uses the Optimum SUPPORTED_TASKS object

### DIFF
--- a/runtimes/huggingface/mlserver_huggingface/common.py
+++ b/runtimes/huggingface/mlserver_huggingface/common.py
@@ -19,6 +19,9 @@ ENV_PREFIX_HUGGINGFACE_SETTINGS = "MLSERVER_MODEL_HUGGINGFACE_"
 HUGGINGFACE_PARAMETERS_TAG = "huggingface_parameters"
 PARAMETERS_ENV_NAME = "PREDICTIVE_UNIT_PARAMETERS"
 
+# Required as workaround until solved https://github.com/huggingface/optimum/issues/186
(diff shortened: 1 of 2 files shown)


runtimes/huggingface/mlserver_huggingface/common.py
```diff
@@ -94,3 +87,33 @@ def parse_parameters_from_env() -> Dict:
...
+    # TODO: Support URI for locally downloaded artifacts
+    # uri = model_parameters.uri
+    model = hf_settings.pretrained_model
+    tokenizer = hf_settings.pretrained_tokenizer
+
+    if model and not tokenizer:
+        tokenizer = model
+
+    if hf_settings.optimum_model:
+        print(f"optimum cache {TRANSFORMER_CACHE_DIR}")
```
adriangonz, line 105: nit: should we use `logger.debug` for this one?

adriangonz (approved):
Looks good! Added a small nitpick, but happy to ignore and move forward with this one.

runtimes/huggingface/mlserver_huggingface/common.py
axsaucedo: Good catch - totally forgot to remove this one, just updated

PR #586, opened 2022-05-13
Extending Hugging Face Runtime for multi-input codec allowing multiple tasks

axsaucedo:
Part of #576

* Adds support for multi-input string codecs (allowing for v2 protocol format)
* Extended to work with multiple pipeline tasks (with extended examples)

Doesn't include converting outputs into v2 protocol.


axsaucedo:
@adriangonz this should be good to go for tomorrow morning merge / dev5 release.

We may have a follow-up PR if there is a fix or workaround for parallel processing in GPU with optimum enabled as per https://github.com/huggingface/optimum/issues/191, but this shouldn't block the dev5 release

PR #587, opened 2022-05-13
Refactor codec's interface 

### LINKED ISSUE #355: Restructure codec interface
Currently mlserver provides codecs for inputs, for some use cases such as in alibi-explain runtime, we also need to be able to encode / decode requests (for the underlying inference call).

There is a big overlap with `InputCodec` and `RequestCodec` so perhaps we need to change the interface to make it more common.

This could be useful / intersects with `tempo`

adriangonz:
Fixes #355 

This PR modifies the codecs interface so that they can be used client-side to help users both encode and decode V2 payloads.

## Changelog
- Rename `encode()` and `decode()` methods to:
  - `InputCodec.encode()` --> `InputCodec.encode_output()`
  - `InputCodec.decode()` --> `InputCodec.decode_input()`
  - `RequestCodec.encode()` --> `RequestCodec.encode_response()`
  - `RequestCodec.decode()` --> `RequestCodec.decode_request()`
- Add new methods to be used client-side:
  - `InputCodec.encode_input()`
  - `InputCodec.decode_output()`
  - `RequestCodec.encode_request()`
  - `RequestCodec.decode_response()`

## TODO
- [x] Add docs to showcase how the new codecs method can be used.


docs/examples/content-type/README.md
```diff
@@ -112,6 +112,40 @@ response = requests.post(
 )
 ```
 
+### Codecs
+
+In order to prepare our V2 request payload, it's often better (and recommended) to use codecs.
```
cliveseldon, line 117: Can we rephrase as its hard to understand - if its recommended then should not be in brackets and the second sentence is hard to follow

adriangonz: Good point!

I've rephrased that section, as well as added more info to the main content-types doc page. Would be great to hear your thoughts on the updated versions!

PR #592, opened 2022-05-17
Added showcase of adaptive batching for GPU parallelisation

axsaucedo:
Fixes part of #576 

Adds showcase of adaptive batching


adriangonz (approved):
Looks great!

PR #605, opened 2022-05-26
Adding --no-cache / -n option to mlserver build

### LINKED ISSUE #604: Add docker build option to not use cache
To ensure fresh environment and while potentially developing a new model version users will not want to use cached layers in the docker build for their image.

In docker this is the --no-cache option, I would make it a boolean option within the cli and also add the --rm option (although default is true might be worth making sure) to remove any intermediate containers after success to again ensure a clean environment in CI/CD as well as locally.

dtpryce:
Attempting to give users the option to use docker build cache or not, this is useful in development of multiple model versions but also in keeping a CI / CD environment clean.

* Added new -n / --no-cache option to mlserver build
* Added --rm option explicitly to ensure intermediate containers are deleted after successful build

Should fix #604

### DIFF
--- a/mlserver/cli/main.py
+++ b/mlserver/cli/main.py
@@ -49,7 +49,7 @@ async def start(folder: str):
 @click.option("-t", "--tag", type=str)
 @click.option("-n", "--no-cache", type=bool)
 @click_async
-async def build(folder: str, tag: str):
+async def build(folder: str, tag: str, no_cache: bool):
     """
     Build a Docker image for a custom MLServer runtime.
     """
--- a/mlserver/cli/main.py
+++ b/mlserver/cli/main.py
@@ -47,14 +47,17 @@ async def start(folder: str):
 @root.command("build")
 @click.argument("folder", nargs=1)
 @click.option("-t", "--tag", type=str)
-@click.option("-n", "--no-cache", type=bool)
+@click.option("--cache/--no-cache", default=False)
 @click_async
-async def build(folder: str, tag: str, no_cache: bool):
+async def build(folder: str, tag: str, cache: bool):
     """
     Build a Docker image for a custom MLServer runtime.
     """
     dockerfile = generate_dockerfile()
-    build_image(folder, dockerfile, tag, no_cache=no_cache)
+    if cache:
+        build_image(folder, dockerfile, tag)
+    else:
+        build_image(folder, dockerfile, tag, no_cache=True)
     logger.info(f"Successfully built custom Docker image with tag {tag}")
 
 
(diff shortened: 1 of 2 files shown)


dtpryce:
Tested usage on our side:

* `mlserver build . -t image:label` behaves the same and uses the cached images / layers
* `mlserver build . --no-cache -t image:label` does not use the build cache at all

mlserver/cli/main.py
```diff
@@ -47,13 +47,17 @@ async def start(folder: str):
 @root.command("build")
 @click.argument("folder", nargs=1)
 @click.option("-t", "--tag", type=str)
+@click.option("--cache/--no-cache", default=False)
```
adriangonz, line 50: Shouldn't this default be `True` so that it aligns with the previous behaviour?

mlserver/cli/main.py
```diff
@@ -47,13 +47,17 @@ async def start(folder: str):
...
+async def build(folder: str, tag: str, cache: bool):
     """
     Build a Docker image for a custom MLServer runtime.
     """
     dockerfile = generate_dockerfile()
-    build_image(folder, dockerfile, tag)
+    if cache:
+        build_image(folder, dockerfile, tag)
+    else:
+        build_image(folder, dockerfile, tag, no_cache=True)
```
adriangonz, line 60: nit: should we just change the `no_cache` kwarg to `cache=cache`? That way we could get rid of this `if` / `else`.

adriangonz (commented):
Looks great @dtpryce !

I've added a small question down below. I always get confused with `click` parameters, so I may be wrong there. There's also a small nitpick comment, but feel free to ignore that one.

mlserver/cli/main.py
dtpryce: Yeah I ummed and arred over this depends on how you want to use it and docker doesn't make it easy either, their usage is --no-cache since the cache is used by default. So if we want to mimic docker I guess we can make it into just --no-cache an then is_flag=True instead of the default=True but I think we would still need to handle since it's None or True surely? Let me test and adjust.

dtpryce:
> Looks great @dtpryce !
> 
> I've added a small question down below. I always get confused with `click` parameters, so I may be wrong there. There's also a small nitpick comment, but feel free to ignore that one.

Both super valid and I think my newest commit fixes both, now just adds --no-cache boolean and able to remove the if / else block :)

adriangonz:
Nice one @dtpryce! Thanks a lot for pushing this one upstream. :rocket: 

This should be good to go now! There're a couple failing tests, but that seems to be unrelated to this PR (a couple tests have suddenly become flaky in `master`).

PR #614, opened 2022-05-31
Add deprecation note for parallel_workers at model settings level

### LINKED ISSUE #578: Show deprecation note when using model's parallel worker setting

adriangonz:
Fixes #578

### DIFF
--- a/docs/user-guide/parallel-inference.md
+++ b/docs/user-guide/parallel-inference.md
@@ -9,8 +9,9 @@ section](#concurrency-in-python) below.
 
 ![](../assets/parallel-inference.svg)
 
-By default, MLServer will spin up a pool of `N` worker processes to run
-inference, where `N` equals the numbers of CPU cores available.
+By default, MLServer will spin up a pool with only one worker process to run
+inference.
+All models will be loaded uniformly across the inference pool workers.
 To read more about advanced settings, please see the [usage section
 below](#usage).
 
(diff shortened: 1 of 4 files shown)


docs/user-guide/parallel-inference.md
```diff
@@ -9,7 +9,8 @@ section](#concurrency-in-python) below.
 
 ![](../assets/parallel-inference.svg)
 
-By default, MLServer will spin up a pool of 4 workers for each loaded model.
+By default, MLServer will spin up a pool of `N` worker processes to run
```
axsaucedo, line 12: Would the expectation not be that perhaps only 1 worker be expected instead of N?

mlserver/parallel/pool.py
```diff
@@ -191,12 +191,23 @@ async def unload_model(self, model: MLModel):
...
     def _should_load_model(self, model: MLModel):
-        # NOTE: This is a remnant from the previous architecture for parallel
-        # workers, where each worker had its own pool.
-        # For backwards compatibility, we will respect when a model disables
-        # parallel inference.
-        if not model.settings.parallel_workers:
-            return False
+        if model.settings.parallel_workers is not None:
+            logger.warning(
+                "DEPRECATED!! The `parallel_workers` setting has now been moved "
```
axsaucedo, line 196: I guess we could look towards full removal in 1.2 and maybe add a note in the warning

axsaucedo (approved):
LGTM

docs/user-guide/parallel-inference.md
adriangonz: By default the pool size is the number of available cores. Should we lower that default?

axsaucedo: It was just a general - I would say 1 May be more intuitive as the default, but I don't think it should block the pr - my thinking is that for larger models like transformers it would be a surprise of the sheer amount of memory loaded by default - I think this was also the reason why in sc v1 python wrapper the default was 1 as well.

adriangonz: That's a great point! Let's lower that default to 1.

PR #624, opened 2022-06-06
Ensure support for grpcio 1.34.0

### LINKED ISSUE #574: Support older versions of grpcio
Some attributes, like `trailing_metadata()` don't seem to exist in older versions of grpcio (e.g. `1.34.0`). Ideally we should also provide support for these ones to avoid clashes with older versions of TF (e.g. `2.5.x`).

adriangonz:
Fixes #574


jklaise (approved):
LGTM

PR #635, opened 2022-06-19
Fix alibi explain runtime issues

cliveseldon:
- [x] Make load async otherwise blocks MLServer
- [x] Move initial metadata call for black box explainers to first explain call. 
- [x] Fix path when loading artifacts to get absolute path

### DIFF
--- a/runtimes/alibi-explain/mlserver_alibi_explain/explainers/black_box_runtime.py
+++ b/runtimes/alibi-explain/mlserver_alibi_explain/explainers/black_box_runtime.py
@@ -14,7 +14,6 @@ from mlserver_alibi_explain.common import (
     construct_metadata_url,
 )
 from mlserver_alibi_explain.runtime import AlibiExplainRuntimeBase
-import traceback
 
 
 class AlibiExplainBlackBoxRuntime(AlibiExplainRuntimeBase):
--- a/runtimes/alibi-explain/mlserver_alibi_explain/explainers/black_box_runtime.py
+++ b/runtimes/alibi-explain/mlserver_alibi_explain/explainers/black_box_runtime.py
@@ -68,7 +68,8 @@ class AlibiExplainBlackBoxRuntime(AlibiExplainRuntimeBase):
 
         if self.infer_metadata is None:
             meta_url = construct_metadata_url(self.infer_uri)
-            # get the metadata of the underlying inference model via v2 metadata endpoint
+            # get the metadata of the underlying inference model via v2
+            # metadata endpoint
             self.infer_metadata = remote_metadata(meta_url)
 
         v2_request = to_v2_inference_request(input_data, self.infer_metadata)
--- a/runtimes/alibi-explain/mlserver_alibi_explain/explainers/black_box_runtime.py
+++ b/runtimes/alibi-explain/mlserver_alibi_explain/explainers/black_box_runtime.py
@@ -38,8 +38,7 @@ class AlibiExplainBlackBoxRuntime(AlibiExplainRuntimeBase):
         super().__init__(settings, explainer_settings)
 
     async def load(self) -> bool:
-        # get the metadata of the underlying inference model via v2 metadata endpoint
-        self.infer_metadata = remote_metadata(construct_metadata_url(self.infer_uri))
+        self.infer_metadata = None
 
         # TODO: use init explainer field instead?
         if self.alibi_explain_settings.init_parameters is not None:
@@ -47,7 +46,7 @@ class AlibiExplainBlackBoxRuntime(AlibiExplainRuntimeBase):
             init_parameters["predictor"] = self._infer_impl
             self._model = self._explainer_class(**init_parameters)  # type: ignore
         else:
-            self._model = self._load_from_uri(self._infer_impl)
+            self._model = await self._load_from_uri(self._infer_impl)
 
         self.ready = True
         return self.ready
@@ -67,6 +66,10 @@ class AlibiExplainBlackBoxRuntime(AlibiExplainRuntimeBase):
         # The contract is that alibi-explain would input/output ndarray
         # in the case of AnchorText, we have a list of strings instead though.
         # TODO: for now we only support v2 protocol, do we need more support?
+
+        if self.infer_metadata is None:
+            # get the metadata of the underlying inference model via v2 metadata endpoint
+            self.infer_metadata = remote_metadata(construct_metadata_url(self.infer_uri))
 
         v2_request = to_v2_inference_request(input_data, self.infer_metadata)
         v2_response = remote_predict(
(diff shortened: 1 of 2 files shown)


runtimes/alibi-explain/mlserver_alibi_explain/explainers/black_box_runtime.py
```diff
@@ -38,16 +38,15 @@ def __init__(self, settings: ModelSettings, explainer_class: Type[Explainer]):
         super().__init__(settings, explainer_settings)
 
     async def load(self) -> bool:
-        # get the metadata of the underlying inference model via v2 metadata endpoint
-        self.infer_metadata = remote_metadata(construct_metadata_url(self.infer_uri))
+        self.infer_metadata = None
```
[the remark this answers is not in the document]
sakoush: also given that perhaps the main fix is to make `_load_from_uri` async. I am not sure if it is still required to move this logic to `_infer_impl`...?

runtimes/alibi-explain/mlserver_alibi_explain/explainers/black_box_runtime.py
```diff
@@ -38,16 +38,15 @@ def __init__(self, settings: ModelSettings, explainer_class: Type[Explainer]):
...
 
         # TODO: use init explainer field instead?
         if self.alibi_explain_settings.init_parameters is not None:
             init_parameters = self.alibi_explain_settings.init_parameters
             init_parameters["predictor"] = self._infer_impl
             self._model = self._explainer_class(**init_parameters)  # type: ignore
         else:
```
sakoush, line 48: should we make ` self._model = self._explainer_class(**init_parameters) ` async as well? perhaps add an issue to consider.

runtimes/alibi-explain/mlserver_alibi_explain/runtime.py
```diff
@@ -119,7 +121,8 @@ def _load_from_uri(self, predictor: Any) -> Explainer:
         uri = model_parameters.uri
         if uri is None:
             raise InvalidModelURI(self.name)
-        return load_explainer(uri, predictor=predictor)
+        absolute_uri = await get_model_uri(self.settings)
```
sakoush, line 124: thanks. this has fallen through the cracks and got unnoticed.

sakoush: perhaps remove 
```uri = model_parameters.uri
        if uri is None:
            raise InvalidModelURI(self.name)``` as not needed anymore?

**Review by sakoush — APPROVED** [review]
Great!

Note sure if we still need to move `remote_metadata` to `_infer_impl`....

File: runtimes/alibi-explain/mlserver_alibi_explain/explainers/black_box_runtime.py [hunk 4]
```diff
@@ -68,6 +67,10 @@ def _infer_impl(self, input_data: Union[np.ndarray, List[str]]) -> np.ndarray:
         # in the case of AnchorText, we have a list of strings instead though.
         # TODO: for now we only support v2 protocol, do we need more support?
 
+        if self.infer_metadata is None:
+            # get the metadata of the underlying inference model via v2 metadata endpoint
+            self.infer_metadata = remote_metadata(construct_metadata_url(self.infer_uri))
+
```
sakoush, line 73: `_infer_impl` is still going to be called via the `_load_from_uri` as alibi needs to make a dummy call at load time still.

runtimes/alibi-explain/mlserver_alibi_explain/explainers/black_box_runtime.py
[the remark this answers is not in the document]
cliveseldon: Just in case there are cases or updates to Alibi where we can not call the model immediately probably best to leave the move. I will remove the setting to None.

runtimes/alibi-explain/mlserver_alibi_explain/runtime.py
cliveseldon: but its optional so might still be None?

runtimes/alibi-explain/mlserver_alibi_explain/explainers/black_box_runtime.py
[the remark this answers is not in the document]
sakoush: `_load_from_uri` is going to call `_infer_impl` and infer_metadata could come in the future from the artifact etc. i.e. outside the call to infer.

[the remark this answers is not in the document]
cliveseldon: if we change the process suggest we change it at that point

runtimes/alibi-explain/mlserver_alibi_explain/explainers/black_box_runtime.py
sakoush: we probably need to look more into the behaviour of async. but in the case of
-  `_load_from_uri`, which will call `alibi.load_explainer` 
- and in the case of `_explain_impl`, which will call `alibi.Explainer.explain`

These calls are not `async` in alibi but we still managed to get them to run asynchronously in mlserver. Given that we dont have an immediate use case for `_explainer_class` it is probably going to require more testing, happy for it to be done as a separate ticket if you feel so. So perhaps add an issue so that we dont forget.

runtimes/alibi-explain/mlserver_alibi_explain/runtime.py
sakoush: I should have been precise. It is already being checked in `get_model_uri`.

runtimes/alibi-explain/mlserver_alibi_explain/explainers/black_box_runtime.py
cliveseldon: yes I suppose that would also need to be added to an async event loop call. Maybe need more example tests in MLServer code for this so maybe a future issue/PR?

runtimes/alibi-explain/mlserver_alibi_explain/runtime.py
```diff
@@ -109,17 +111,18 @@ async def _async_explain_impl(
...
             raise ModelParametersMissing(self.name)
-        uri = model_parameters.uri
-        if uri is None:
-            raise InvalidModelURI(self.name)
-        return load_explainer(uri, predictor=predictor)
+        absolute_uri = await get_model_uri(self.settings)
+        evtLoop = asyncio.get_event_loop()
+        return await evtLoop.run_in_executor(
+            None, load_explainer, absolute_uri, predictor
+        )
```
sakoush, line 125: there is a helper function `execute_async` that could use instead

adriangonz:
Leaving aside the benchmarks, it seems like every other check seems to pass now, so this should be good to go! :+1:

PR #650, opened 2022-06-28
Improve worker's processing loop

### LINKED ISSUE #643: Ensure workers pick up multiple requests at the same time
For some models (e.g. explainers), the inference implementation will leverage AsyncIO to make further IO calls. However, currently the workers' implementation means that only one request will be processed at a time. Therefore, MLServer won't make the most out of these AsyncIO calls.

To account for this, we should update the current main loop in the workers so that instead of `awaiting` on processing each request, it just sets a callback to return back the response once done. This will allow workers to take on multiple messages at the same time.

On top of this, to avoid a single worker accepting every message in the queue, we'll need to split the current "shared" input queue into one per worker, doing a "round-robbin" across workers. That way we'll ensure that the load is shared between them.

adriangonz:
Fixes #643 

## Changelog
- Create one requests queue per worker
- Ensure that the worker always schedules messages as they come, processing the responses through callbacks
- Forward messages to workers from the pool in a round-robin manner

### DIFF
--- a/mlserver/parallel/pool.py
+++ b/mlserver/parallel/pool.py
@@ -1,5 +1,6 @@
 import asyncio
 
+from itertools import cycle
 from asyncio import Future
 from concurrent.futures import ThreadPoolExecutor
 from multiprocessing import Queue
@@ -45,16 +46,17 @@ class InferencePool:
 
         self._workers = {}
         self._settings = settings
-        self._requests: Queue[InferenceRequestMessage] = Queue()
         self._responses: Queue[InferenceResponseMessage] = Queue()
         self._async_responses: Dict[str, Future[InferenceResponse]] = {}
         self._executor = ThreadPoolExecutor()
         for idx in range(self._settings.parallel_workers):
             # TODO: Set callback to restart worker if it goes down (would
             # `worker.join` help with that?)
-            worker = Worker(self._requests, self._responses)
+            worker = Worker(self._responses)
             worker.start()
             self._workers[worker.pid] = worker
+
+        self._workers_round_robin = cycle(self._workers.keys())
 
         # Start processing responses
         self._start_processing_responses()
@@ -117,7 +119,10 @@ class InferencePool:
             model_version=model_version,
             inference_request=inference_request,
         )
-        self._requests.put(request_message)
+
+        worker_pid = next(self._workers_round_robin)
+        worker = self._workers[worker_pid]
+        worker.send_request(request_message)
 
         loop = asyncio.get_running_loop()
         async_response = loop.create_future()
@@ -227,7 +232,6 @@ class InferencePool:
         await terminate_queue(self._responses)
         await cancel_task(self._process_responses_task)
         self._responses.close()
-        self._requests.close()
         self._executor.shutdown()
 
     async def _close_workers(self):
--- a/mlserver/parallel/worker.py
+++ b/mlserver/parallel/worker.py
@@ -2,6 +2,7 @@ import asyncio
 import select
 import signal
 
+from asyncio import Task
 from queue import Empty
 from multiprocessing import Process, Queue, JoinableQueue
 from concurrent.futures import ThreadPoolExecutor
@@ -75,13 +76,10 @@ class Worker(Process):
 
     async def coro_run(self):
         self.__inner_init__()
+        loop = asyncio.get_event_loop()
 
         while self._active:
-            readable, _, _ = select.select(
-                [self._requests._reader, self._model_updates._reader],
-                [],
-                [],
-            )
+            readable = await loop.run_in_executor(self._executor, self._select)
             for r in readable:
                 if r is self._requests._reader:
                     try:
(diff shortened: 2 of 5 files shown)


mlserver/parallel/pool.py
```diff
@@ -117,7 +119,10 @@ async def predict(
             model_version=model_version,
             inference_request=inference_request,
         )
-        self._requests.put(request_message)
```
sakoush, line 120: I am interested to know why we think round robin in more effective that having a single queue especially if one of the workers is very busy?

I am assuming that if one worker is very busy, the other workers will fetch work from the shared queue, which will make the system have an overall better throughput?

mlserver/parallel/pool.py
```diff
@@ -117,7 +119,10 @@ async def predict(
...
+
+        worker_pid = next(self._workers_round_robin)
+        worker = self._workers[worker_pid]
```
sakoush, line 124: could we abstract this bit so that we have different "load balancing" in the future. For example if we can get the number of remaining messages on each queue we could return the id of the worker with the least work etc.

mlserver/parallel/worker.py
```diff
@@ -91,35 +94,50 @@ async def coro_run(self):
...
-                        self.model_updates.task_done()
+                        self._model_updates.task_done()
                         return
 
-                    await self._process_model_update(model_update)
-                    self.model_updates.task_done()
+                    update_task = asyncio.create_task(
+                        self._process_model_update(model_update)
+                    )
+                    update_task.add_done_callback(self._update_cb)
```
sakoush, line 111: code similar to creating task  and assigning call back for inference. perhaps consider having a helper function?

mlserver/parallel/worker.py
```diff
@@ -91,35 +94,50 @@ async def coro_run(self):
...
+    async def _process_request(self, request) -> InferenceResponseMessage:
         try:
             model = await self._model_registry.get_model(
                 request.model_name, request.model_version
             )
 
             inference_response = await model.predict(request.inference_request)
 
-            response = InferenceResponseMessage(
+            return InferenceResponseMessage(
```
sakoush, line 130: remove `self._responses.put(response)` ?

mlserver/parallel/worker.py
```diff
@@ -132,22 +150,38 @@ async def _process_model_update(self, update: ModelUpdateMessage):
...
+            logger.error(e)
+
+        self._model_updates.task_done()
+
+    def send_request(self, request_message: InferenceRequestMessage):
+        """
+        Send an inference request message to the worker.
+        Note that this method should be both multiprocess- and thread-safe.
+        """
+        loop = asyncio.get_event_loop()
```
sakoush, line 165: why do we need this statement?

mlserver/parallel/worker.py
```diff
@@ -132,22 +150,38 @@ async def _process_model_update(self, update: ModelUpdateMessage):
...
     async def send_update(self, model_update: ModelUpdateMessage):
         """
         Send a model update to the worker.
         Note that this method should be both multiprocess- and thread-safe.
         """
         loop = asyncio.get_event_loop()
-        self.model_updates.put(model_update)
-        await loop.run_in_executor(self._executor, self.model_updates.join)
+        self._model_updates.put(model_update)
+        await loop.run_in_executor(self._executor, self._model_updates.join)
```
sakoush, line 175: do we still need `send_update` to be `async`?

sakoush (commented):
Great stuff!

I have left some comments.

+ should we have a "unit" test that exercises some of the issues that Clive saw with the previous implementation?

mlserver/parallel/worker.py
adriangonz: Good point. We could have a `call_with_cb` helper method?

mlserver/parallel/worker.py
adriangonz: Your'e right, no need for this one - I think that was from copy & pasting the other method.

mlserver/parallel/worker.py
adriangonz: Yes, because model update messages can be awaited until the worker confirms that the model has been loaded / unloaded.

mlserver/parallel/pool.py
sakoush: Ah I see, a worker can basically consume all work from the queue and convert them to coroutines that are waiting to execute on the worker event loop? can we have a  limit, perhaps if the current count of gorouttines on a specific worker is beyond a certain threshold then the worker is not able to pull more from the queue? Having one queue per-worker enables more smart scheduling decisions so I think this is the correct approach and given what you described, round robin is going to spread load across workers evenly which we didnt have before. Maybe a future PR would schedule work to the queue with least messages, assuming we can control how the worker from consuming from the queue?

adriangonz: Yuuup, totally. 

In the future, it would make sense to choose the worker based on its backlog. Since each worker has its own queue, this shouldn't be too hard TBH.

adriangonz:
Failure in tests seem to be caused by the flakiness that we're trying to sort out in #649. Therefore, it should be alright to go ahead with this one :+1:

Issue #235, opened 2021-07-15
Benchmarking in CI/CD


axsaucedo:
Run benchmarks as part of CI/CD, detecting whether there’s been a slow down.

adriangonz:
Fixed by #655

PR #655, opened 2022-07-05
Update benchmarks and add automation

adriangonz:
## Changelog
- Split benchmarks into two separate ones to avoid some issues with k6
- Add GH Action to automate benchmarks on each PR with thresholds on throughput
- Update benchmarks' README results


adriangonz:
Fixes #235

PR #657, opened 2022-07-08
handle missing model settings folders during load

cliveseldon:
If a folder disappears during model load this can cause an exception.

https://github.com/SeldonIO/MLServer/blob/acae8a77c04b339c1346775456ab241d80a276fe/mlserver/repository.py#L28-L29

The glob returns all model-settings files. If between this time and the time of a load the file disappears this will cause an exception 

https://github.com/SeldonIO/MLServer/blob/acae8a77c04b339c1346775456ab241d80a276fe/mlserver/repository.py#L31-L33

This fix ignores these exceptions for loading model settings files.

In future, it may be better to try to find the model to load in a more directed way rather than searching the whole repository.

### DIFF
--- a/mlserver/repository.py
+++ b/mlserver/repository.py
@@ -33,6 +33,7 @@ class ModelRepository:
                     model_settings = self._load_model_settings(model_settings_path)
                     all_model_settings.append(model_settings)
                 except:
+                    logger.exception(f"Failed load model settings at {model_settings_path}")
                     continue
 
         # If there were no matches, try to load model from environment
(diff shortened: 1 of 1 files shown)


mlserver/repository.py
```diff
@@ -29,8 +29,11 @@ async def list(self) -> List[ModelSettings]:
             matches = glob.glob(pattern, recursive=True)
 
             for model_settings_path in matches:
-                model_settings = self._load_model_settings(model_settings_path)
-                all_model_settings.append(model_settings)
+                try:
+                    model_settings = self._load_model_settings(model_settings_path)
+                    all_model_settings.append(model_settings)
+                except:
```
sakoush, line 35: should we be specific with the exception here i.e. `FileNotFoundError`?

cliveseldon: I think we need to be as wide as possible as otherwise what happens if someone includes a broken model settings file. This would stop all models from loading? 
If there is a set of standard PyDantic exceptions we could catch maybe?

PR #665, opened 2022-07-14
Template creation with cookiecutter for base project initialisation

axsaucedo:
Initial work to create a basic template for mlserver thourgh the "Secure Production ML Engineering Project Base" https://github.com/EthicalML/sml-security/. Currently making the template variable configurable in case people may want to create their own org templae, as well as if we want to create different templates as well that cover potentially different usecases.

### DIFF
--- a/mlserver/cli/init_project.py
+++ b/mlserver/cli/init_project.py
@@ -1,0 +1,13 @@
+import subprocess
+from ..logging import logger
+
+
+def init_cookiecutter_project(template: str):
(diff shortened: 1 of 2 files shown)


adriangonz (approved):
That's really cool!! It's way simpler than I expected. 

Cookiecutter should then ask for all the parameters, right?

axsaucedo:
@adriangonz that is correct - does look pretty cool in action

index

PR #671, opened 2022-07-24
Add headers to alibi responses

cliveseldon:


runtimes/alibi-explain/mlserver_alibi_explain/runtime.py
```diff
@@ -87,8 +87,16 @@ async def predict(self, payload: InferenceRequest) -> InferenceResponse:
             model_name=self.name,
             model_version=self.version,
             outputs=[output_data],
+            parameters=Parameters(headers={
+                "x-seldon-alibi-type":"explanation",
```
sakoush, line 91: do we want also to be specific about which type of explanation we are returning?

runtimes/alibi-detect/mlserver_alibi_detect/runtime.py
```diff
@@ -74,9 +74,30 @@ async def predict(self, payload: types.InferenceRequest) -> types.InferenceRespo
...
+
+    def getAlibiType(self) -> str:
+        module: str = type(self._model).__module__
+        method = module.split(".")[-2]
+        if method == "cd":
+            return "drift"
+        elif method == "od":
+            return "outlier"
+        elif method == "ad":
+            return "adversarial"
```
sakoush, line 101: perhaps these should be backed up by a map and make it more visible.

runtimes/alibi-explain/mlserver_alibi_explain/runtime.py
```diff
@@ -87,8 +87,16 @@ async def predict(self, payload: InferenceRequest) -> InferenceResponse:
...
+                "x-seldon-alibi-method": self.getAlibiMethod()
+            }),
         )
 
+    def getAlibiMethod(self) -> str:
+        module: str = type(self._model).__module__
+        return module.split(".")[-1]
```
sakoush, line 98: nit: if we ever change the name of the module, is that going to affect anything downstream using it? if so maybe we need to introduce a mapping?

sakoush (commented):
Left some comments here.

runtimes/alibi-explain/mlserver_alibi_explain/runtime.py
cliveseldon: I'd rather not have another map of alibi module names to our names as there are so many and we want to be dynamic

runtimes/alibi-explain/mlserver_alibi_explain/runtime.py
cliveseldon: That should come from the method header we return no?

runtimes/alibi-detect/mlserver_alibi_detect/runtime.py
```diff
@@ -74,9 +74,30 @@ async def predict(self, payload: types.InferenceRequest) -> types.InferenceRespo
...
+            "x-seldon-alibi-method": self.getAlibiMethod()
+        }
         return types.InferenceResponse(
             model_name=self.name,
             model_version=self.version,
             parameters=y["meta"],
             outputs=outputs,
         )
+
+    def getAlibiMethod(self) -> str:
```
adriangonz, line 89: nit: could we use `snake_case` instead of `camelCase` to align with Python conventions?

runtimes/alibi-explain/mlserver_alibi_explain/runtime.py
sakoush: oh yes I thought we are just parsing the module of the mlserver runtime (i.e. `black_box_runtime` etc.) and not the alibi module. In this case then yes.

maybe then we could have a simple test case to make sure that these headers are set as expected?

PR #678, opened 2022-07-30
Allow sklearn transform

cliveseldon:


runtimes/sklearn/mlserver_sklearn/sklearn.py
```diff
@@ -51,14 +52,27 @@ async def predict(self, payload: InferenceRequest) -> InferenceResponse:
 
     def _check_request(self, payload: InferenceRequest) -> InferenceRequest:
         if not payload.outputs:
+            found_predict_fn = False
+            if self.settings.parameters:
+                print(self.settings)
```
adriangonz, line 57: nit: Should we remove this `print()`?

runtimes/sklearn/mlserver_sklearn/sklearn.py
```diff
@@ -51,14 +52,27 @@ async def predict(self, payload: InferenceRequest) -> InferenceResponse:
...
+                if self.settings.parameters.extra:
```
adriangonz, line 58: Should we put it at the `parameters` level instead of `parameters.extra`?

runtimes/sklearn/mlserver_sklearn/sklearn.py
```diff
@@ -51,14 +52,27 @@ async def predict(self, payload: InferenceRequest) -> InferenceResponse:
...
+                            )
+                        ]
+                        found_predict_fn = True
             # By default, only return the result of `predict()`
-            payload.outputs = [RequestOutput(name=PREDICT_OUTPUT)]
+            if not found_predict_fn:
+                payload.outputs = [RequestOutput(name=PREDICT_OUTPUT)]
         else:
-            for request_output in payload.outputs:
+            for request_output in payload.outputs:  # type: ignore
```
adriangonz, line 70: Do you know why does the linter complains about this line?

adriangonz (commented):
Looks good!

Should we add a test to ensure it calls the `transform` method, and that it respects the value from the model settings?

runtimes/sklearn/mlserver_sklearn/sklearn.py
cliveseldon: I suppose its following the definition of `extra` which is `Arbitrary settings, dependent on the inference runtime
    implementation.`

runtimes/sklearn/mlserver_sklearn/sklearn.py
cliveseldon: removed and seems ok now. Not sure why was failing before

runtimes/sklearn/mlserver_sklearn/sklearn.py
adriangonz: Good point, let's leave it there.

I'm starting to see the `extra` field a bit redundant, as (semantically) `parameters` should already be able to hold extra model parameters. That's totally outside the scope of this PR though.

adriangonz (approved):
Looks great! Thanks for adding that test @cliveseldon.

Issue #656, opened 2022-07-06
Convert numpy.float32 output values to float


ryanpmccaffrey:
The output of certain huggingface models (e.g., pipelines with the task of [token-classification](https://github.com/huggingface/transformers/blob/main/src/transformers/pipelines/token_classification.py)) outputs values of type numpy.float32.  Data of this type is not [JSON serializable](https://stackoverflow.com/questions/53082708/typeerror-object-of-type-float32-is-not-json-serializable).  Therefore, these models will fail when dumping the output, which is typically addressed with a post-processing step.

In [mlserver-huggingface runtime.py](https://github.com/SeldonIO/MLServer/blob/master/runtimes/huggingface/mlserver_huggingface/runtime.py#L86-L88) before Line 86 could you add a post-processing fix to convert values from  numpy.float32 to float?

Minimum viable code to reproduce issue:
```
import json
from transformers import pipeline

nlp = pipeline("ner", model='dslim/bert-base-NER')
output = nlp("My name is Wolfgang and I live in Berlin")
json.dumps(output)
```

```
TypeError: Object of type float32 is not JSON serializable
```

ryanpmccaffrey:
For reference:
https://github.com/numpy/numpy/issues/16432

adriangonz:
Hey @ryanpmccaffrey, 

Thanks for raising this one.

MLServer supports the use of [content types](https://mlserver.readthedocs.io/en/latest/user-guide/content-type.html) which let you describe how should MLServer treat each input and output. This is how, for instance, MLServer can encode back multiple response types on the MLflow runtime. However, it seems like the HuggingFace runtime forces every response back to be a JSON-encoded string. 

We'll have a look into this to make the HF runtime more flexible (i.e. similar to how the MLServer-MLflow runtime behaves).

dtpryce:
@adriangonz I have also recently seen a similar error when trying to create a custom runtime that might include numpy int types.

e.g.
```
async def predict(self, payload: types.InferenceRequest) -> types.InferenceResponse:
    request = self._extract_json(payload)[0]
    outcomes = self.clf.predict(request)
    output = {
        "id": request["id"],
        "target": outcome[0]
    }
    return types.InferenceResponse(
            id=payload.id,
            outputs=[
                types.ResponseOutput(
                    name=request['id'],
                    shape=[len(output)],
                    datatype="BYTES",
                    data=[output],
                    parameters=types.Parameters(content_type="str")
                )
            ]
        )
```

[...]

PR #692, opened 2022-08-17
support json serialize all kind's of huggingface pipelines inputs/outputs 

pepesi:
huggingface_runtime output JSON serializer does not support NumPy basic datatypes when the data is a dict value


adriangonz (commented):
Hey @pepesi ,

Nice spot! Thanks a lot for contributing this one. 

Changes look good to me! :+1: 

Before merging though, would you be able to provide a test case that covers the issue that this PR fixes?

pepesi:
I fixed the lint error and add some tests for the NumpyEncoder, but I can't run the test successfully in my local environment because of the error `ImportError: cannot import name 'deepspeed_reinit' from 'transformers.deepspeed'`. Can I run tests on Github Actions?

pepesi:
> I fixed the lint error and add some tests for the NumpyEncoder, but I can't run the test successfully in my local environment because of the error `ImportError: cannot import name 'deepspeed_reinit' from 'transformers.deepspeed'`. Can I run tests on Github Actions?

Tests passed on my Desktop

pepesi:
After test more huggingface models, I think the NumpyEncoder should renamed to `HuggingfaceOutputJSONEncoder`. Because huggingface pipeline output is not only numpy datatypes but also Pillow's Image, before more tests run, i'm not sure how many python types exists in pipeline's outputs.

pepesi:
A small script tested passed in my local, it seems ok now, but I won't add it to tests, because I think it's so heavy 

```
from transformers.pipelines import pipeline
from transformers import Conversation
import json
import numpy as np
from PIL import Image
import io
import base64


class CommonJSONEncoder(json.JSONEncoder):
    def default(self, obj):
        if isinstance(obj, np.ndarray):
            return obj.tolist()
        if isinstance(obj, (np.float_, np.float16, np.float32, np.float64)):
            return float(str(obj))
        if isinstance(obj, (np.int_, np.int8, np.int16, np.int32, np.int64)):
            return int(obj)
        if isinstance(obj, Image.Image):
            buf = io.BytesIO()
            obj.save(buf, format="png")
            return base64.b64encode(buf.getvalue()).decode()
        if isinstance(obj, Conversation):
            return {
                'uuid': str(obj.uuid),
                'past_user_inputs': obj.past_user_inputs,
                'generated_responses': obj.generated_responses,
                'new_user_input': obj.new_user_input
            }
        return json.JSONEncoder.default(self, obj)


sumarytext = "The tower is 324 metres (1,063 ft) tall, about\
the same height as an 81-storey building, and the tallest structure \
in Paris. Its base is square, measuring 125 metres (410 ft) on each side.\
During its construction, the Eiffel Tower surpassed the\
Washington Monument to become the tallest man-made structure\
in the world, a title it held for 41 years until the Chrysler\
Building in New York City was finished in 1930. It was the\
first structure to reach a height of 300 metres. Due to the\
addition of a broadcasting aerial at the top of the tower in\
1957, it is now taller than the Chrysler Building by 5.2 metres\
(17 ft). Excluding transmitters, the Eiffel Tower is the second\
tallest free-standing structure in France after the Millau Viaduct."


ALL_TASKS_TESTS = {
    "audio-classification": [
        {"args": (), "kwargs": {"inputs": "fixtures/audio.mp3"}}
    ],
    "automatic-speech-recognition": [
        {"args": (), "kwargs": {"inputs": "fixtures/audio.mp3"}}
    ],
    "feature-extraction": [
        {
            "args": (),
            "kwargs": {
                "inputs": "this pure text",
            },
        }
    ],
    "text-classification": [
        {
            "args": (
                "A greeting brings the care of the heart,\
        a blessing brings the care of the body, and a short\
        message brings the love. May you be lucky and happy,\
        and your life is sweet as honey, and your career is \
        successful and official!",
            ),
            "kwargs": {},
        }
    ],
    "token-classification": [
        {
            "args": (),
            "kwargs": {"inputs": "Hello I'm Omar and I live in Zürich."}
        }
    ],
    "question-answering": [
        {
            "args": (),
            "kwargs": {
                "question": "what's her job?",
                "context": "Her name is Lily, she is a singer",
            },
        }
    ],
    "table-question-answering": [
        {
            "args": (),
            "kwargs": {
                "table": {
                    "actors": [
                        "brad pitt",
                        "leonardo di caprio",
                        "george clooney"
                    ],
                    "age": ["56", "45", "59"],
                    "number of movies": ["87", "53", "69"],
                    "date of birth": [
                        "7 february 1967",
                        "10 june 1996",
                        "28 november 1967",
                    ],
                },
                "query": "when brad pitt born?",
            },
        }
    ],
    "visual-question-answering": [
        {
            "args": (),
            "kwargs": {
                "image": "fixtures/dogs.jpg",
                "question": "how many dogs here?"
            },
        }
    ],
    "fill-mask": [
        {
            "args": (),
            "kwargs": {
                "inputs": "i am come from <mask>",
            },
        }
    ],
    "summarization": [{"args": (sumarytext,), "kwargs": {}}],
    "translation_en_to_de": [
        {"args": ("My name is Sarah and I live in London",), "kwargs": {}}
    ],
    "text2text-generation": [
        {
            "args": (
                "question: What is 42 ? context: 42 is the answer to life,\
            the universe and everything",
            ),
            "kwargs": {},
        },
        {
            "args": ("translate from English to French: I'm very happy",),
            "kwargs": {}
        },
    ],
    "text-generation": [
        {"args": ("My name is Tylor Swift and I am",), "kwargs": {}}
    ],
    "zero-shot-classification": [
        {
            "args": (),
            "kwargs": {
                "sequences": "Last week I upgraded my iOS version and ever\
            since then my phone has been overheating whenever I\
            use your app.",
                "candidate_labels": [
                    "mobile",
                    "website",
                    "billing",
                    "account",
                    "access",
                ],
                "multi_class": False,
            },
        }
    ],
    "zero-shot-image-classification": [
        {
            "args": (),
            "kwargs": {
                "images": "fixtures/dogs.jpg",
                "candidate_labels": ["dogs", "cats", "tigers"],
            },
        }
    ],
    "conversational": [
        {
            "args": (),
            "kwargs": {
                "conversations": [
                    Conversation("Hello"),
                    Conversation("how are you"),
                    Conversation("where are you from"),
                ],
            },
        }
    ],
    "image-classification": [
        {
            "args": (),
            "kwargs": {
                "images": "fixtures/dogs.jpg",
            },
        }
    ],
    "image-segmentation": [
        {
            "args": (),
            "kwargs": {
                "inputs": "fixtures/dogs.jpg",
            },
        }
    ],
    "object-detection": [
        {
            "args": (),
            "kwargs": {
                "inputs": "fixtures/dogs.jpg",
            },
        }
    ],
}


all_types = {}
outs = []


def inspect_types(output):
    if isinstance(output, dict):
        for k, v in output.items():
            inspect_types(k)
            inspect_types(v)
    elif isinstance(output, (list, tuple)):
        for el in output:
            inspect_types(el)
    elif isinstance(output, (int, str, float)):
        all_types[type(output)] = 1
    else:
        all_types[type(output)] = 1


for task, argslist in ALL_TASKS_TESTS.items():
    print("-" * 80)
    print(task)
    p = pipeline(task)
    for arg in argslist:
        output = p(*arg["args"], **arg["kwargs"])
        inspect_types(output)
        outs.append(output)


print("all kinds of datatypes:")
print(list(all_types.keys()))


print(json.dumps(outs, cls=CommonJSONEncoder))
```

runtimes/huggingface/mlserver_huggingface/common.py
```diff
@@ -1,13 +1,16 @@
 import os
+import io
+import base64
 import json
 from typing import Optional, Dict
 from distutils.util import strtobool
+from PIL import Image
```
adriangonz, line 7: Should we add this as a dependency of the HuggingFace runtime? Or is it already a direct dependency of the `transformers` library?

adriangonz (commented):
Hey @pepesi ,

Massive thanks for the effort you're putting to make the encoder for HuggingFace models fully complete! It's certainly not an easy task! :rocket: :muscle:

runtimes/huggingface/mlserver_huggingface/common.py
pepesi: Sorry, I forgot to add the dependency to the requirement.  I will fix it now.

pepesi:
`timm` is the requirement for some visualization-related Huggingface tasks, `Pillow` is one of the dependencies for it;

`ffmpeg` is the system requirement for some audio-related Huggingface tasks, I found there is no place to install single apt package for a specific runtime, So I add it to MLServer's basic runtime

pepesi:
WIP

pepesi:
What changes in this PR

I'm using Seldon and MLServer in a project, that is aimed to fast deploy and experience any Hugginface model; when I test with some visual-related models, I go some JSON serialize error like this https://github.com/SeldonIO/MLServer/issues/656, At first, I just want fix the JSON serialize error. But when I get further, I get more trouble with inputs; 

changes:
1. add `is_single` filed in parameters to support decode request input as a single value, this field is now affected `StringCodec` `NumpyCodec` `Base64Codec`;
2. add `HuggingfaceRuntimeCodec`, which can auto decode the inputs as pipelines args 

**May not be compatible**
`NumpyCodec` always converts the inputs to a single `np.ndarray` before, now if no parameter `is_single` provided, `NumpyCodec` would decodes the inputs as an list.  

A remedy here, change the is_sinle default value to None, then change NumpyCodec`s default behavior, if is_single is None,  set is_single to True; but which may confusing 

@adriangonz  any advise here?

adriangonz:
Hey @pepesi ,

Thanks a lot for the time you've put on this PR - I know it hasn't been an easy journey!

It's grown quite massively, so it will take us a bit of time to review it. Particularly, considering that there are breaking changes to some of the existing functionality. We will do our best though!

Dockerfile
```diff
@@ -25,7 +25,7 @@ ENV MLSERVER_MODELS_DIR=/mnt/models \
 RUN apt-get update && \
     apt-get -y --no-install-recommends install \
         unattended-upgrades \
-        libgomp1 libgl1-mesa-dev libglib2.0-0 build-essential && \
+        libgomp1 libgl1-mesa-dev libglib2.0-0 build-essential ffmpeg && \
```
adriangonz, line 28: We have now moved to a different base image for Docker, but it should already install ffmpeg (although in a diff way). Therefore, feel free to pick up `master`'s version of the `Dockerfile` when rebasing :+1:

mlserver/codecs/numpy.py
```diff
@@ -105,17 +105,28 @@ class NumpyCodec(InputCodec):
     TypeHint = np.ndarray
 
     @classmethod
-    def can_encode(csl, payload: Any) -> bool:
-        return isinstance(payload, np.ndarray)
+    def can_encode(cls, payload: Any) -> bool:
```
adriangonz, line 108: Good spot :+1:

mlserver/codecs/utils.py
```diff
@@ -47,6 +47,19 @@ def _get_content_type(
     return None
 
 
+def set_content_type(request: Parametrised, content_type: str):
+    if request.parameters is None:
+        request.parameters = Parameters(content_type=content_type)
+    else:
+        request.parameters.content_type = content_type
+
```
adriangonz, line 55: Nice one! This is super useful to ensure codecs set their own content type :+1:

runtimes/huggingface/mlserver_huggingface/codecs.py
```diff
@@ -1,89 +1,351 @@
...
-    element.
+    for huggingface JSONCodec has high priority
     """
+    if JSONCodec.can_encode(data):
+        return JSONCodec
+    if ImageCodec.can_encode(data):
+        return ImageCodec
+    if ConversationCodec.can_encode(data):
+        return ConversationCodec
+    return find_input_codec_by_payload(data)
```
adriangonz, line 69: If we decorate them with `@register_input_codec` or `@register_request_codec` we should then be able to find them with `find_input_codec(payload=data)` (as in, they'll go into the general codec registry, which is not a bad thing).

runtimes/huggingface/mlserver_huggingface/codecs.py
```diff
@@ -1,89 +1,351 @@
...
+
+
+@register_input_codec
+class ImageCodec(InputCodecTy):
+    ContentType = CONTENT_TYPE_IMAGE
```
adriangonz, line 74: Is there any reason why we need separate constants for these values? As in, instead of just keeping the actual value here and referring to `ImageCodec.ContentType` everywhere else?

runtimes/huggingface/mlserver_huggingface/codecs.py
```diff
@@ -1,89 +1,351 @@
...
+        return JSONCodec
+    if ImageCodec.can_encode(data):
+        return ImageCodec
+    if ConversationCodec.can_encode(data):
+        return ConversationCodec
+    return find_input_codec_by_payload(data)
+
+
+@register_input_codec
+class ImageCodec(InputCodecTy):
```
adriangonz, line 73: I was thinking that, to help future maintainers, we should split these codecs into their own files. As in, instead of keeping them all under `mlserver_huggingface/codecs.py`, we should create a new `mlserver_huggingface/codecs` package (i.e. folder), and then have:

- `mlserver_huggingface/codecs/image.py`
- `mlserver_huggingface/codecs/conversation.py`
- `mlserver_huggingface/codecs/json.py`
- ....

Each of these files could also have any codec-specific helpers (like `get_img_bytes`, which could live in `mlserver_huggingface/codecs/image.py`). 

This will also help with the code review :+1:

runtimes/huggingface/mlserver_huggingface/common.py
```diff
@@ -128,8 +137,207 @@ def load_pipeline_from_settings(hf_settings: HuggingFaceSettings) -> Pipeline:
...
+            return int(obj)
+        elif isinstance(obj, Image.Image):
+            return b64encode_image(obj)
+        elif isinstance(obj, Conversation):
+            return serialize_conversation(obj)
+        else:
+            return json.JSONEncoder.default(self, obj)
+
+
+class MetaInfo:
```
adriangonz, line 193: If this one is a static dict, we could also convert it into a constant dict (i.e. as opposed to a class).

runtimes/huggingface/mlserver_huggingface/runtime.py
```diff
@@ -80,25 +77,23 @@ async def load(self) -> bool:
...
-            model_version=self.version,
-            outputs=[prediction_encoded],
+        model_metadata = MetadataModelResponse(
+            name=self.name,
+            platform=self._settings.platform,
+            versions=self._settings.versions,
+            inputs=self._settings.inputs or metadata.get("inputs", []),
+            outputs=self._settings.outputs or metadata.get("outputs", []),
         )
+        return model_metadata
```
adriangonz, line 99: Oh this is brilliant! It's gonna be so useful to expose the pipeline's metadata here! 

Having said that, the best way to do this would be to edit the `self.inputs` and / or `self.outputs` fields, similar to how the MLflow's runtime handles these changes:

https://github.com/SeldonIO/MLServer/blob/f5297d42f00be647954bbed8b12b0bea29c00994/runtimes/mlflow/mlserver_mlflow/runtime.py#L132-L161

mlserver/batch_processing.py
```diff
@@ -88,8 +88,12 @@ def json_to_triton(
     inference_request = InferenceRequest.parse_obj(orjson.loads(request_json))
     for request_input in inference_request.inputs or []:
         new_input = httpclient.InferInput(
-            request_input.name, request_input.shape, request_input.datatype
+            request_input.name,
+            request_input.shape,
+            request_input.datatype,
         )
+        if request_input.parameters is not None:
+            new_input._parameters.update(request_input.parameters.dict())
```
RafalSkolasinski, line 96: Wondering about this change and exact motivation behind. It is true that currently there is no option to set parameters on the inputs and request as a whole but I am worried that modifying triton's internal parameters (as indicated by `_`) could lead to unexpected breaking changes in the future.

adriangonz, line 96: Good catch! 

We initially considered leveraging the `_parameters` field, but then decided not to, in order to avoid using Tritonclient's internal interface. Instead of that, the general advise when using `mlserver infer` is to just configure the `content_type` through the model's metadata (i.e. the `model-settings.json` file). 

Would be great to hear your thoughts behind this change though.

runtimes/huggingface/mlserver_huggingface/codecs.py
```diff
@@ -1,89 +1,351 @@
-from typing import Optional, Type, Any, Dict
+import json
+from typing import Optional, Type, Any, Dict, List, Union
+from functools import partial
+from xmlrpc.client import Boolean
```
adriangonz, line 4: What is `xmlrpc`? Shouldn't this be `typing.Boolean` instead?

adriangonz (changes requested):
Hey @pepesi ,

I've finally found the time to make a first pass at this one. 

Before anything else, I have to say this PR is really impressive! The attention to detail shown, and the thoroughness, is incredible! Massive thanks for spending the time on these changes! :rocket: 

I've added a couple comments below. Would be great to hear your thoughts on those ones. Besides that, I'm not 100% convinced on the introduction of the `is_single` parameter and wanted to hear your reasoning behind why it is required. 

My view is that codecs should always operate with lists of things (i.e. batches), and then is up to the runtime to deal with these. This simplifies other features like batching, as well as simplifies the code, which doesn't need to deal with multiple types (i.e. type T or list of type T).

runtimes/huggingface/mlserver_huggingface/codecs.py
pepesi: Some data may encode by multiple codecs, but there is no priority with codecs, `mlserver.codecs._find_codec_by_payload` just return the first codecs matched; I want to find codecs in order with priority, and that's the reason; 

If my understanding of the encoder registry is wrong, please let me know

runtimes/huggingface/mlserver_huggingface/codecs.py
pepesi: Your suggestion is right. I will revise it according to that

runtimes/huggingface/mlserver_huggingface/codecs.py
pepesi: Sorry, it's a typo, VSCODE auto add this and I forget to check.

pepesi:
Firstly I agree with the opinion, **codecs should always operate with lists of things (i.e. batches),  and then is up to the runtime to deal with these**.  Before that, I didn't realize that it was a principle to follow. So I think I will remove `is_single` next
 
---

And then, Let me explain why I add `is_single` before. 

For `visual-question-answering` and `table-question-answering` task pipelines, they support args like this `pipeline(image=Image.Image, question=str)`, but codecs would decode the request data like this `[Image.Image]` `[question]` (a list). So I add the parameter to support decode data as a single value, which would make the codecs decode the request data to like `Image.Image` `str`(a single value).

due to adding the parameter is_single,  triton httpclient request need to pass it to the server also, that why I modify InferInput._parameters. as @RafalSkolasinski said, it may cause breaking changes.

runtimes/huggingface/mlserver_huggingface/codecs.py
adriangonz: Oh, I see. Ok, that makes sense. In that case, it may be worth dropping a comment on that method to briefly explain that reasoning (i.e. and avoid people changing that in the future).

adriangonz (approved):
This looks great @pepesi ! Thanks a lot for making those changes.

I know this has been a long journey - what started as a fix for numpy, is now a full-blown update to the HF runtime - but it's such a great addition! Massive thanks for all the effort that has gone behind this contribution. :rock: 

I think the changes looks great. I've added a minor question, but I don't think it's necessarily a blocker. It would be great to have your thoughts on that one. Besides that, once tests are green, this should be good to go ahead! :+1:

runtimes/huggingface/tests/test_runtime.py
```diff
@@ -29,10 +29,10 @@ async def test_infer_multiple(
 
     res = await runtime.predict(inference_request)
 
-    assert len(res.outputs) == 1
+    assert len(res.outputs) == 2
```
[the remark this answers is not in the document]
pepesi: That test case is running a `question-answering` task. the input length is 2,  So the task output is a list considered with two answers, like this `[{score: 1, answer: something}, {score: 1, answer: something}]`,  when encoding that data as a response,   RequestCodec `HuggingfaceRequestCodec` matched. ThenHuggingfaceRequestCodec will encode the data.  The HuggingfaceRequestCodec always considers that data is a batch output and always encodes elements in that data as JSON (why always JSON? I can't find a better way now. ). The output is like this 

```
{
    model_name: name,
    version: version,
    outputs: [
        {
           "name": "output_1",
           "data": ["{score: 1, answer: something}"]
        }, {
           "name": "output_2",
           "data": ["{score: 1, answer: something}"]
        }
    ]
}
```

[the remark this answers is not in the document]
adriangonz: Ah right, in this case, it would fit better with the protocol to return a single output (e.g. named `output`), that contains two entries. When you look at the input, it's done a similar way (i.e. you will have a `question` input, with two elements, and a `context` input with two elements).

What codec is the one handling the encoding in this case? Would that be a big change over the current PR?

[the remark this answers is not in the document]
pepesi: In this case, RequestCodec is `HuggingfaceRequestCodec`. 

For the input side, HuggingfaceRequestCodec  will  find the input codec  in `HuggingfaceRequestCodec.InputCodecsWithPriority`, if not found, then found MLserver build-in InputCodec, if not found again, just `HuggingfaceRequestCodec.DefaultCodec` to encoding data; 

For the output side, the InputCodec `HuggingfaceJSONCodec` is always used;

That may be a big change, The default behavior changed, So should I keep this behavior like before, and keep hole prediction as a single output?  advice need here.

I noticed that `mlserver.codecs.utils.encode_inference_response` has a TODO, https://github.com/SeldonIO/MLServer/blob/5e8afc33c03e98f5fb226748dcfbe96d5fa753b5/mlserver/codecs/utils.py#L90-L95

If the codec could be overridden by metadata, This may be easy to deal with. But this may be another feature;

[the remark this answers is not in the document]
adriangonz: I see... would there be any reason not to use the `HuggingfaceListJSONCodec` instead? That one seems to have the same logic as the runtime had previously, where multiple instances of the same output are treated as extra elements of the same `output` field.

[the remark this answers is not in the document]
pepesi: OK, I will change it to HuggingfaceListJSONCodec soon

adriangonz:
Thanks a lot for making those changes @pepesi. Once again, massive thanks for all the effort you've put into this one. 

PR looks great and tests are green, so this should be good to go! :rocket:

PR #700, opened 2022-08-24
Add support to configurable metrics_rest_server_prefix

jonathansp:
This PR moves the string prefix "_rest_server_", currently hardcoded to settings, keeping its default value as _rest_server_.

Because we currently set up our application prefixes as "_http_application_" this can be achieved with MLServer by removing the hardcoded string.

cc @adriangonz


adriangonz (approved):
Nice one! Thanks for your contribution @jonathansp! :+1: 

Changes look good to me, so once tests are green we can go ahead and merge. :rocket:

PR #711, opened 2022-09-01
custom server requirements.txt file generation added to the notebook

### LINKED ISSUE #710: missing mentioning the requirements.txt in the "serving custom model page"
Although I can see it is in the [directory](https://github.com/SeldonIO/MLServer/blob/master/docs/examples/custom/requirements.txt) it is not mentioned in the [webpage](https://mlserver.readthedocs.io/en/latest/examples/custom/README.html) and will result in error for people copy/pasting from the webpage and not using the notebook.

saeid93:
Fixes #710

### DIFF
--- a/docs/examples/custom/README.ipynb
+++ b/docs/examples/custom/README.ipynb
@@ -339,6 +339,26 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "### Specifying requirements\n",
+    "MLServer will automatically find your requirements.txt file and install necessary python packages"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%%writefile models.py\n",
+    "numpy==1.22.0\n",
+    "numpyro==0.8.0\n",
+    "jax==0.2.24"
+   ]
+  },
+  {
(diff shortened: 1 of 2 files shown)


docs/examples/custom/README.ipynb
```diff
@@ -335,6 +335,26 @@
...
+    "MLServer will automatically find your requirements.txt file and install necessary python packages"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%%writefile models.py\n",
```
adriangonz, line 352: Should this line be `%%writefile requirements.txt` instead?

adriangonz (changes requested):
Nice one @saeid93! Thanks a lot for adding this one.

Just added a small comment below about the file name. Once that's addressed, I think this should be good to go! :+1:

docs/examples/custom/README.ipynb
saeid93: Sure, That's why I shouldn't do all the pull requests late at night. :D
Should work now. Please check.

adriangonz (approved):
Nice one! Thanks for adding the changes.

Once the tests are green, this should be good to go :+1:

Issue #712, opened 2022-09-01
missing mentioning pushing image to the docker registry in the "serving custom model page"


saeid93:
Pushing the built docker container to a docker registry is not mentioned after the build process in [Serving a custom model](https://mlserver.readthedocs.io/en/latest/examples/custom/README.html). In my case, it was necessary but I wasn't sure if this is a generalizable case to add this as a pull request. If necessary I can add an extra step to the notebook for pushing to the container registry step.

adriangonz:
Hey @saeid93 ,

That's a great point.

I agree with you though, in that it's not that generalizable. For starters, we would need to know the Docker Hub repo and image name that the user wants to use. 

What we could do instead, is to add a "Note block" clarifying that Seldon Core may expect that the image is available on Docker Hub or an alternative Docker registry and therefore the user may need to push it there?

PR #713, opened 2022-09-02
Custom server container registry

saeid93:
solves #712


docs/examples/custom/README.md
```diff
@@ -311,6 +322,8 @@ Now that we've built a custom image and verified that it works as expected, we c
 There is a large number of tools out there to deploy images.
 However, for our example, we will focus on deploying it to a cluster running [Seldon Core](https://docs.seldon.io/projects/seldon-core/en/latest/).
 
+Also consider that depending on your Kubernetes installation Seldon Core might excpect to get the container image from a public container registry like [Docker hub](https://hub.docker.com/) or [Google Container Registry](https://cloud.google.com/container-registry). For that you need to do an extra step of pushing the container to the registry using `docker tag <image name> <container registry>/<image name>` and `docker push <container registry>/<image name>` and also updating the `image` section of the yaml file to `<container registry>/<image name>`.
```
adriangonz, line 325: Would it make sense to add this as a `note` block instead? It feels like, that way, it may be less distracting from the main the main walkthrough of what's going on (i.e. as something else to just keep in mind). 

This can be done adding a "code block" in markdown, tagged as `{note}`, as in:

````
```{note}
bli bla blu
```
````

**Review by adriangonz — COMMENTED** [review]
That's a great addition @saeid93! Thanks for contributing that one.

I've added a small comment below, but it would be great to hear your thoughts as well.

File: docs/examples/custom/README.md (hunk 1 above)
Comment (adriangonz) [reply docs/examples/custom/README.md:325]: Hey @saeid93 ,

Not sure I follow.

It's possible to have multi-line code blocks within notes. Just to clarify, under the hood we use `myst`, which is a markdown preprocessor to render the docs with Sphinx. So it will turn the `note` block to a visual box, separate from the main text.

My main concern is that, otherwise, the current clarification breaks the previous flow between its adjacent paragraphs. You can see that the next paragraph starts with `For that, we will need to create a...`, which, after the change, seems to refer to the action of pushing the image instead of deploying the model.

Since the mention to pushing the image is mainly a clarification, I think it makes sense to move it to a separate note. As it's mainly something to keep in mind and not a specific call to action.

Comment (saeid93) [reply docs/examples/custom/README.md:326]: Sure, I almost forgot that it will be also used for generating the webpages and not just for the jupyter notebook. Please check the new one.

File: docs/examples/custom/README.md [hunk 2]
```diff
@@ -311,6 +322,10 @@ Now that we've built a custom image and verified that it works as expected, we c
 There is a large number of tools out there to deploy images.
 However, for our example, we will focus on deploying it to a cluster running [Seldon Core](https://docs.seldon.io/projects/seldon-core/en/latest/).
 
+```{note}
+Also consider that depending on your Kubernetes installation Seldon Core might excpect to get the container image from a public container registry like [Docker hub](https://hub.docker.com/) or [Google Container Registry](https://cloud.google.com/container-registry). For that you need to do an extra step of pushing the container to the registry using `docker tag <image name> <container registry>/<image name>` and `docker push <container registry>/<image name>` and also updating the `image` section of the yaml file to `<container registry>/<image name>`.
```
adriangonz, line 326: typo: `expect` instead of `excpect`

adriangonz (changes requested):
Nice one! Changes look great @saeid93!

Just added a small comment flagging a typo, but everything else looks great! 

BTW you're able to check a live preview of the docs through the link below. The new `Note` block looks fantastic!

https://mlserver--713.org.readthedocs.build/en/713/examples/custom/README.html#deploying-our-custom-image

image

saeid93:
@adriangonz is it all good now? please let me know if anything else is needed.

adriangonz:
All looks good! Thanks for making the changes @saeid93 !

Issue #726, opened 2022-09-13
No information about how to retrieve back the raw dictionary data from mlserver output in documentation


saeid93:
There is no information about how to retrieve back the raw dictionary data from mlserver output in the [documentation](https://mlserver.readthedocs.io/en/latest/examples/custom-json/README.html). I will add a pull request for discussion.

PR #727, opened 2022-09-13
added code for extracting back text output

saeid93:
solves #726
Not sure if needed but I didn't notice myself at the first pass of the notebook. Maybe happens for other users too.
Not sure if the changes are necessary but added for discussion anyway.

### DIFF
--- a/docs/examples/custom-json/README.md
+++ b/docs/examples/custom-json/README.md
@@ -44,7 +44,7 @@ class JsonHelloWorldModel(MLModel):
         request = self._extract_json(payload)
         response = {
             "request": request,
-            "server_response": "Got your request. Hello from the server."
+            "server_response": "Got your request. Hello from the server.",
         }
         response_bytes = json.dumps(response).encode("UTF-8")
 
@@ -58,9 +58,9 @@ class JsonHelloWorldModel(MLModel):
                     shape=[len(response_bytes)],
                     datatype="BYTES",
                     data=[response_bytes],
-                    parameters=types.Parameters(content_type="str")
+                    parameters=types.Parameters(content_type="str"),
                 )
-            ]
+            ],
         )
 
     def _extract_json(self, payload: types.InferenceRequest) -> Dict[str, Any]:
@@ -124,14 +124,14 @@ For that, we can use the Python types that `mlserver` provides out of box, or we
 ```python
 import requests
 import json
+from pprint import PrettyPrinter
 
-inputs = {
-    "name": "Foo Bar",
-    "message": "Hello from Client (REST)!"
-}
+pp = PrettyPrinter(indent=1)
+
+inputs = {"name": "Foo Bar", "message": "Hello from Client (REST)!"}
 
 # NOTE: this uses characters rather than encoded bytes. It is recommended that you use the `mlserver` types to assist in the correct encoding.
-inputs_string= json.dumps(inputs)
+inputs_string = json.dumps(inputs)
 
 inference_request = {
     "inputs": [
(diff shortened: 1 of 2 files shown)


docs/examples/custom-json/README.md
```diff
@@ -124,30 +124,35 @@ For that, we can use the Python types that `mlserver` provides out of box, or we
...
 }
 
 endpoint = "http://localhost:8080/v2/models/json-hello-world/infer"
 response = requests.post(endpoint, json=inference_request)
 
-response.json()
+print(f"full response:\n")
+print(response)
+# retrive text output as dictionary
+output = json.loads(response.outputs[0].contents.bytes_contents[0])
```
adriangonz, line 153: These look like the fields for gRPC outputs. However, they will be different in the case of REST.

To simplify things, could you use the `StringRequestCodec` instead? That would look something like this (I'm writing this from the top of my head, so pls double check the imports and methods):

```python
import json

from mlserver.types import InferenceResponse
from mlserver.codecs.string import StringRequestCodec

inference_response = InferenceResponse.parse_raw(response.text)
raw_json = StringRequestCodec.decode_response(inference_response)
output = json.loads(raw_json)
```

docs/examples/custom-json/README.md
```diff
@@ -183,22 +188,26 @@ inference_request = types.InferenceRequest(
...
 grpc_channel = grpc.insecure_channel("localhost:8081")
 grpc_stub = dataplane.GRPCInferenceServiceStub(grpc_channel)
 
 response = grpc_stub.ModelInfer(inference_request_g)
-response
+
+print(f"full response:\n")
+print(response)
+# retrive text output as dictionary
+output = json.loads(response.outputs[0].contents.bytes_contents[0])
```
adriangonz, line 208: Similar to the above, you could run this as:

```python
import json

from mlserver.grpc.converters import ModelInferResponseConverter
from mlserver.codecs.string import StringRequestCodec

# Note that here we just convert from the gRPC types to the MLServer types
inference_response = ModelInferResponseConverter.to_types(response)
raw_json = StringRequestCodec.decode_response(inference_response)
output = json.loads(raw_json)
```

adriangonz (changes requested):
That's a great addition @saeid93 ! 

We've got a few examples showing how to encode requests, but not much showing how to decode responses.

I think that the snippets could be simplified using codecs though, so I've added a couple suggestions there. Would be great to hear your thoughts!

docs/examples/custom-json/README.md
saeid93: This isn't working as the following line:
```
inference_response = ModelInferResponseConverter.to_types(response)
```
is calling the following, which the `inference_response` output for the subsequent line of the scripts will be `None`
https://github.com/SeldonIO/MLServer/blob/fac90c83a3536dbbea9e7075c590e33ace424537/mlserver/grpc/converters.py#L301

docs/examples/custom-json/README.md
adriangonz: Good spot! It will return a list of strings by default.

docs/examples/custom-json/README.md
adriangonz: Damnit, you're completely right. There's no implementation for that method yet. 

One to add for the next version of MLServer. While that's not there though, what you already have for the gRPC block is probably your best bet :+1:

docs/examples/custom-json/README.md
saeid93: Thanks! All done!

saeid93:
All done according to the comments. Thank you.

saeid93:
@adriangonz could you please have a look at the build error once you got a chance, I couldn't understand the error.

adriangonz:
Hey @saeid93 ,

It seems to be a flaky test error. I just kicked off the tests again, and they should pass now. :+1:

adriangonz (approved):
Changes look great @saeid93 ! Nice one! :rocket: 

Once tests are green, we can merge this one.

PR #730, opened 2022-09-17
llink in the hyperlink fixed

saeid93:
Minor type fix


adriangonz (approved):
Thanks for the change @saeid93 ! :+1:

Issue #739, opened 2022-09-23
check for consistency between MLServer and HuggingFace Batch size


saeid93:
HuggingFace runtime has a batch_size variable in its setting. This should be checked against the MLServer max_batch_size setting for consistency.

```json
{
    "name": "transformer",
    "implementation": "mlserver_huggingface.HuggingFaceRuntime",
    "max_batch_size": 5,
    "max_batch_time": 1,
    "parameters": {
        "extra": {
            "task": "text-generation",
            "pretrained_model": "distilgpt2",
            "device": 0,
            "batch_size": 5
        }
    }
}

```

PR #740, opened 2022-09-23
check for huggingface batch sizes

saeid93:
Solves #739

### DIFF
--- a/runtimes/huggingface/mlserver_huggingface/runtime.py
+++ b/runtimes/huggingface/mlserver_huggingface/runtime.py
@@ -62,7 +62,7 @@ class HuggingFaceRuntime(MLModel):
             logger.warning(
                 f"hf batch_size: {self.hf_settings.batch_size} is different "
                 f"from MLServer max_batch_size: {settings.max_batch_size}"
-                )
+            )
 
         super().__init__(settings)
 
--- a/runtimes/huggingface/mlserver_huggingface/runtime.py
+++ b/runtimes/huggingface/mlserver_huggingface/runtime.py
@@ -57,6 +57,15 @@ class HuggingFaceRuntime(MLModel):
                     ),
                 )
 
+        if settings.max_batch_size != self.hf_settings.batch_size:
+            raise InvalidTranformerInitialisation(
+                500,
+                (
+                    f"HggingFace batch_size: {self.hf_settings.batch_size}."
+                    f" Should match the MLServer max_batch_size: {settings.max_batch_size}"
+                ),
+            )
+
         super().__init__(settings)
 
     async def load(self) -> bool:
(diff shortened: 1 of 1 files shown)


runtimes/huggingface/mlserver_huggingface/runtime.py
```diff
@@ -57,6 +57,15 @@ def __init__(self, settings: ModelSettings):
                     ),
                 )
 
+        if settings.max_batch_size != self.hf_settings.batch_size:
+            raise InvalidTranformerInitialisation(
```
adriangonz, line 61: Instead of an error, would it make sense to just log here a warning instead (i.e. with `logger.warning()`)?

adriangonz (commented):
Nice one @saeid93! Thanks for following up with a PR. :rocket: 

Just added a small comment below - let me know what you think.

runtimes/huggingface/mlserver_huggingface/runtime.py
saeid93: Thanks! Glad to be of any help.

About the comment, I think we can do it both ways, however, as it will work even if the batch sizes don't match a warning might be a better choice. I changed it in the next commit.

But, I think maybe in a future version or in another PR it might be a better design option to just read the hf_batch size into the MLServer max_batch_size unless there is an important reason for covering `max_batch_size != max_batch_size` cases. I cannot think of such situations but there might be some situations in which it makes sense to separate these two, can you think of any?

saeid93:
re solved comment:
Thanks! I'm glad that it is now merged.

Yeah, I agree but even if it's 1. MLServer batching being set up 2. huggingFace batching then it makes even more sense to just merge the two variables into the MLServer one (unless I'm missing an important point?). I'm not quite sure if I get this part correctly:
_"So it would need some work to let the runtime tweak the settings dynamically before the batcher bit gets initialised."_

adriangonz:
Yes, I agree. 

To clarify that last point, I was thinking about syncing both settings in both directions. That is, injecting the MLServer batch size into the HF model parameters (i.e. MLServer --> HF), and enabling adaptive batching when the HF model parameters define a batch size (i.e. HF --> MLServer). On the former, it should be straightforward to inject the batch size into the HF parameters when the MLServer one is provided.

However, on the latter, there's no way atm for the `HuggingfaceRuntime` to initialise or modify the adaptive batching settings **after** the model has been loaded. So this would also need some work at the MLServer core package to allow inference runtimes to do that.

saeid93:
hmm, thanks for the explanation. Now I get it. But does it need to be necessarily bi-directional? To me, it would be better to have one and only one variable in charge of setting the batch size regardless of the underlying framework (e.g. hugging face). Then the users will only need to know/learn the MLServer setting format and the rest is abstracted from them by each of the underlying framework's custom runtime implementations. That way MLServer will be able to provide a unified interface for all the underlying frameworks. I think having it both ways will violate the simplicity aimed to be offered by MLServer as it will be confusing to have two configs in `setting.json` doing the same thing. I would like to know your thoughts on this.

adriangonz:
Hey @saeid93 ,

Fully agree with those thoughts. 

If we can have it just in the MLServer --> HF direction, that would be the simplest (both implemention- and usability-wise).

Issue #741, opened 2022-09-24
Add support for `dictConfig` from logging settings JSON file


dumaas:
Adding support for `logging.config`'s [dictConfig ](https://docs.python.org/3/library/logging.config.html#logging.config.dictConfig)will allow for more advanced logging configurations (with third-party modules like [structlog](https://www.structlog.org/en/stable/), for example).

Users will be now be able to point `logging_settings` to a JSON file conforming to the [dictConfig schema](https://docs.python.org/3/library/logging.config.html#logging-config-dictschema).

Fix should be relatively simple: (see PR https://github.com/SeldonIO/MLServer/pull/742)

PR #742, opened 2022-09-24
Add support for dictConfig with json file

dumaas:
Resolves https://github.com/SeldonIO/MLServer/issues/741

### DIFF
--- a/mlserver/logging.py
+++ b/mlserver/logging.py
@@ -28,7 +28,7 @@ def apply_logging_file(logging_settings: Union[str, Dict]):
         else:
             logging.config.fileConfig(fname=logging_settings, disable_existing_loggers=False)
     elif isinstance(logging_settings, Dict):
-        logging.config.dictConfig(config)
+        logging.config.dictConfig(logging_settings)
     else:
         logger.warning("Unable to parse logging_settings.")
 
--- a/mlserver/logging.py
+++ b/mlserver/logging.py
@@ -1,7 +1,9 @@
+import json
 import logging
 import sys
 
 from logging import Formatter, StreamHandler
+from pathlib import Path
 import logging.config
 
 from .settings import Settings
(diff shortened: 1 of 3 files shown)


mlserver/logging.py
```diff
@@ -17,7 +19,12 @@ def get_logger():
 
 
 def apply_logging_file(logging_settings: str):
-    logging.config.fileConfig(fname=logging_settings, disable_existing_loggers=False)
+    if "json" in Path(logging_settings).suffix:
```
adriangonz, line 22: Should we add a third case where `logging_settings` contains the dictionary directly (as opposed to the file name with the dictionary)?

adriangonz (commented):
Nice one @dumaas! I think this is a great addition :rocket: 

Just added one point below. It would be great to get your thoughts on that!

mlserver/logging.py
dumaas: Thanks for taking a look at this @adriangonz! I've added some logic to support all 3 options...what do you think?

It's a bit messier now, with `logging_settings` potentially being either a `str` or `Dict`, but I like the added functionality. Let me know if you have any suggestions for how this might be cleaned up / if that's needed!

```python
def apply_logging_file(logging_settings: Union[str, Dict]):
    if isinstance(logging_settings, str) and Path(logging_settings).is_file():
        if "json" in Path(logging_settings).suffix:
            with open(logging_settings) as settings_file:
                config = json.load(settings_file)
            logging.config.dictConfig(config)
        else:
            logging.config.fileConfig(fname=logging_settings, disable_existing_loggers=False)
    elif isinstance(logging_settings, Dict):
        logging.config.dictConfig(logging_settings)
    else:
        logger.warning("Unable to parse logging_settings.")
```

adriangonz: Nice one @dumaas ! I didn't realise that supporting the 3rd case would also mean supporting multiple types there. I think you worked your way around that beautifully though! Changes look great! :rocket: 

Once the tests are green, this should be good to go :+1:

dumaas: Thank you! 😄 

@adriangonz I believe this commit should resolve the linting errors: https://github.com/SeldonIO/MLServer/pull/742/commits/6ea65673512ac175e0d2cf5e43d067fd0a0140b5

PR #748, opened 2022-09-26
Add custom server settings

### LINKED ISSUE #674: Custom uvicorn configuration
We are having some difficulties with Uvicorn configuration. We need to apply some custom settings on the rest server like using custom timeouts or applying a custom ssl certificate. This configuration is exposed by Uvicorn and can be set ussing uvicorn.Config method, but it is not exposed by MLServer configuration file.

To support these new configurations, we could add new fields in the settings file and then add them to the Uvicorn configuration input, but we believe this is an inflexible and not scalable solution.

So could it make sense to allow users to directly give a kwargs dictionary in Uvicorn format with their own settings? By doing this, we would have an easier way to apply this configuration and users would have more flexibility to give a complete and customized Uvicorn configuration. Allowing them to use any configuration, even if it is not supported in the MLServer settings file.

Maybe a similar model could be applied to the GRPC server and the new metrics server too.

What do you think?

pablobgar:
Allow custom uvicorn and grpc configuration.

Closes #674

### DIFF
--- a/mlserver/grpc/server.py
+++ b/mlserver/grpc/server.py
@@ -66,8 +66,8 @@ class GRPCServer:
     def _get_options(self) -> List[Tuple[str, Any]]:
         options_dict = {}
 
-        if self._settings.custom_grpc_server_settings:
-            options_dict.update(self._settings.custom_grpc_server_settings)
+        if self._settings._custom_grpc_server_settings:
+            options_dict.update(self._settings._custom_grpc_server_settings)
 
         max_message_length = self._settings.grpc_max_message_length
         if max_message_length is not None:
--- a/mlserver/grpc/server.py
+++ b/mlserver/grpc/server.py
@@ -64,17 +64,22 @@ class GRPCServer:
         return self._server
 
     def _get_options(self) -> List[Tuple[str, Any]]:
-        options = []
+        options_dict = {}
+
+        if self._settings.custom_grpc_server_settings:
+            options_dict.update(self._settings.custom_grpc_server_settings)
 
         max_message_length = self._settings.grpc_max_message_length
         if max_message_length is not None:
-            options += [
-                ("grpc.max_message_length", max_message_length),
-                ("grpc.max_send_message_length", max_message_length),
-                ("grpc.max_receive_message_length", max_message_length),
-            ]
+            options_dict.update(
+                {
+                    "grpc.max_message_length": max_message_length,
+                    "grpc.max_send_message_length": max_message_length,
+                    "grpc.max_receive_message_length": max_message_length,
+                }
+            )
 
-        return options
+        return list(options_dict.items())
 
     async def start(self):
         self._create_server()
--- a/mlserver/grpc/server.py
+++ b/mlserver/grpc/server.py
@@ -67,6 +67,9 @@ class GRPCServer:
         options_dict = {}
 
         if self._settings._custom_grpc_server_settings:
+            logger.warning(
+                "gRPC custom configuration is out of support. Use as your own risk"
+            )
             options_dict.update(self._settings._custom_grpc_server_settings)
 
         max_message_length = self._settings.grpc_max_message_length
--- a/mlserver/metrics/server.py
+++ b/mlserver/metrics/server.py
@@ -1,5 +1,4 @@
 import uvicorn
-import logging
 
 from fastapi import FastAPI
 from starlette_exporter import handle_metrics
--- a/mlserver/metrics/server.py
+++ b/mlserver/metrics/server.py
@@ -7,8 +7,6 @@ from starlette_exporter import handle_metrics
 from ..settings import Settings
 from .logging import logger
 
-logger = logging.getLogger("mlserver.metrics")
-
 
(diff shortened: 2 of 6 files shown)


mlserver/settings.py
```diff
@@ -146,6 +146,11 @@ class Config:
     kafka_topic_input: str = "mlserver-input"
     kafka_topic_output: str = "mlserver-output"
 
+    # Custom server settings
+    custom_rest_server_settings: Optional[dict] = None
```
adriangonz, line 150: Should we prefix these fields with an underscore (i.e. as `_custom_rest_server_settings`)? This would mainly be to highlight that these settings are provided "as-is" to users and not supported.

Happy to hear other suggestions though! Perhaps we could also add a warning message when they are used?

mlserver/rest/server.py
```diff
@@ -61,17 +61,24 @@ async def start(self):
...
-            "port": self._settings.http_port,
-            "root_path": self._settings.root_path,
-            "access_log": self._settings.debug,
-        }
+        kwargs = {}
+
+        if self._settings.custom_rest_server_settings:
+            kwargs.update(self._settings.custom_rest_server_settings)
+
+        kwargs.update(
```
adriangonz, line 69: In this case (same as gRPC and the metrics server), should it be one or the other? As in, either the custom one or the other one? My main concern is future issues trying to reconcile both.

mlserver/grpc/server.py
```diff
@@ -64,17 +64,22 @@ def _create_server(self):
...
-        return options
+            options_dict.update(
+                {
+                    "grpc.max_message_length": max_message_length,
+                    "grpc.max_send_message_length": max_message_length,
+                    "grpc.max_receive_message_length": max_message_length,
+                }
+            )
+
+        return list(options_dict.items())
```
adriangonz, line 82: Nice one! Pretty clean :+1:

adriangonz (commented):
Hey @pablobgar ,

Thanks a lot for taking the lead on this one! :+1: 

This is a great first stab at this issue. I've added a couple comments below, but all of them are open for discussion. It will be great to hear your thoughts!

mlserver/rest/server.py
pablobgar: In my opinion, explicitly defined settings should override the custom configuration. This allows users to configure extra parameters while they're still using default values for host, http_port, grpc_port, etc.

mlserver/settings.py
pablobgar: I think both are good ideas!! I've added the prefix and a warning message.

mlserver/metrics/server.py
```diff
@@ -37,6 +35,14 @@ async def start(self):
...
     def _get_config(self):
+        kwargs = {}
+
+        if self._settings._custom_metrics_server_settings:
+            logger.warning(
+                "REST custom configuration is out of support. Use as your own risk"
+            )
+            kwargs.update(self._settings._custom_metrics_server_settings)
+
         kwargs = {
```
adriangonz, line 46: Shouldn't the `kwargs` be merged instead of reassigned?

adriangonz (commented):
Changes look great! Added a small comment on the metrics server, but everything else looks good to go :+1:

mlserver/metrics/server.py
pablobgar: You're right. Fixed

adriangonz:
Nice one @pablobgar! Thanks a lot for taking the time to work on this one. Changes look great, tests are green, and this should be ready to go! :rocket:

lizzzcai:
Hi @adriangonz, is there a prerelease for this feature? The latest [release](https://pypi.org/project/mlserver/#history) doesn't have this feature. As `_custom_grpc_server_settings` is a `dict`, in this case is it possible to pass it via `env`? Thanks.

adriangonz:
Hey @lizzzcai, 

We haven't yet cut a pre-release including this one. You should however be able to use `master` directly. 

Regarding the latter point, I imagine that you should be able to pass this one as a JSON-encoded string in an environment variable (i.e. `MLSERVER__CUSTOM_GRPC_SERVER_SETTINGS='{"foo": "bar"}'`) - but I'm not sure if it has been tested. @pablobgar have you tried out this use case by any chance?

Kolajik:
For anyone wondering:

I've tried to set an env variable `MLSERVER__CUSTOM_GRPC_SERVER_SETTINGS='{"grpc.keepalive_timeout_ms": 5000,"grpc.keepalive_permit_without_calls": 1,"grpc.keepalive_time_ms": 600000}'`, but it doesn't seem it is being picked up by the server.

dtpryce:
And to add to @Kolajik I have also fixed all this by removing the `_` from the custom settings, since pre v2 pydantic treats these not as fields but as standard python attributes. Hence, you can ONLY set them within the class itself not via environment variables or from a file so they are entirely hidden from the user. To ensure they are exposed you just have to remove that underscore everywhere (settings, rest/server, grpc/server and metrics/server) then they are usable.

PR #757, opened 2022-10-01
Clean _is_newer function and enable semantic versioning

Salehbigdeli:


adriangonz (approved):
This is brilliant @Salehbigdeli ! That's a really elegant implementation! Thanks a lot for taking the lead on this one. :rocket: 

Changes look great from my side, so once tests are green this should be good to go :+1:

mlserver/registry.py
```diff
@@ -59,7 +50,7 @@ def __init__(
         on_model_reload: List[ModelReloadHook] = [],
         on_model_unload: List[ModelRegistryHook] = [],
     ):
-        self._versions: Dict[str, MLModel] = {}
+        self._versions: Dict[Union[str, None], MLModel] = {}
```
adriangonz, line 53: Oh, I see what you did there! That's a pretty neat idea - makes for a way simpler implementation :+1:

mlserver/registry.py
```diff
@@ -225,20 +196,15 @@ async def get_model(self, version: str = None) -> MLModel:
...
         # NOTE: `.values()` returns a "view" instead of a list
-        models = list(self._versions.values())
-
-        # Add default if not versioned (as it won't be present on the
-        # `_versions` dict
-        if self.default and not self.default.version:
-            models.append(self.default)
-
-        return models
+        return list(self._versions.values())
```
adriangonz, line 199: nit: would this return the default model twice if it also has a version?

adriangonz (approved):
Thanks for addressing the test errors @Salehbigdeli!

I've added a small nitpick below, but I think even with that this should be good to go. It would still be great to hear your thoughts on that point though :+1:

mlserver/registry.py
Salehbigdeli: I believe this isn't possible since values are stored in `dict`.

adriangonz: Wouldn't they get saved under different keys though? As in,

```python
self._versions = {
	"v1.1.1": model1,
	None: model1
}

self._versions.values() # then returns -> [model1, model1]
```

Salehbigdeli: What about this case?

```python
self._versions = {
	"v1.1.1": model1,
        "v2.1.1": model1,
	None: model1
}

self._versions.values() # then returns -> [model1, model1, model1]```
```
I don't understand the use-case, so I don't have any alternative.

adriangonz: Well, a model object can only have one version, so `model1` couldn't have two different versions. However, a model can have a version **and** be the "default" model at the same time.

So I guess the question is, following your changes, is the previous case possible? (i.e. having something like `{"v1": model1, None: model1}`)

Salehbigdeli: Thanks for explaining.
I tried to preserve previous behavior, In previous version of code if there were a default it append the default to the end of `models` and return it.
Unfortunately I didn't understand the code enough to see any problem in this case.

adriangonz: Yeah, totally understand that. The previous code was a bit convoluted, and wasn't easy to understand. The new version is way cleaner.

Regarding the potential duplicated model, that was essentially the main reason of this `self.default.version` check (i.e. if the default model also has a version, don't include it, as it will already be present on the `self._versions` dictionary). 

https://github.com/SeldonIO/MLServer/pull/757/files#diff-8d68e1b9bf57de7aa43eaed4717d3fc4c14241afef15576891e1ad27d74322a5L230-L233

Perhaps what would be best is to add a test case covering that it doesn't happen after the changes? (i.e. that `registry.get_models()` doesn't return the same model twice)

mlserver/registry.py
```diff
@@ -196,7 +196,7 @@ async def get_model(self, version: str = None) -> MLModel:
 
     async def get_models(self) -> List[MLModel]:
         # NOTE: `.values()` returns a "view" instead of a list
-        return list(self._versions.values())
+        return list(v for k, v in self._versions.items() if k)
```
adriangonz, line 199: Should we also add the previous `if self._default and not self._default.version` check to add the default model if it doesn't have a version?

tests/handlers/test_model_repository.py
```diff
@@ -101,11 +101,7 @@ async def test_load_removes_stale_models(
 
     # Validate that the stale test models have been loaded
     registry_models = await model_registry.get_models(sum_model_settings.name)
-    stale_length = (
-        len(to_load)
-        + 1  # Count the (stale) default model
-        + 1  # Count the previous (non-stale) model
-    )
+    stale_length = len(to_load) + 1  # Count the previous (non-stale) model
```
adriangonz, line 104: I think the previous test case was correct, and was just flagging that the default model is no longer present - this may be because it doesn't have a version, therefore it's not present on the `self._versions` dict (see my other comment above).

tests/test_registry.py
```diff
@@ -251,3 +256,30 @@ async def test_rolling_reload(
...
+
+
+@pytest.mark.parametrize(
+    "versions",
+    [
+        ["v0", "v1"],
+        ["v0", "v1", "v2"],
+    ],
+)
+async def test_get_models_dont_return_default_twice(
```
adriangonz, line 268: Nice one adding a test to cover the other behaviour! :+1:

tests/test_registry.py
```diff
@@ -212,7 +213,11 @@ async def test_find_default(
 
 
 async def test_model_not_ready(model_registry: MultiModelRegistry):
-    slow_model_settings = ModelSettings(name="slow-model", implementation=SlowModel)
+    slow_model_settings = ModelSettings(
+        name="slow-model",
+        implementation=SlowModel,
+        parameters=ModelParameters(version="v1"),
```
adriangonz, line 219: Similar to above, I think the previous test case was correct, and just started failing because the default model without a version was not present.

adriangonz (changes requested):
Nice one with the changes!

I think the only point left is to ensure that the default model is **also** returned when it **doesn't have a version**.

adriangonz:
Hey @Salehbigdeli,

Following up on this one, since there hasn't been any activity for a month, we'll be closing this PR for now. 

Having said that, we're still super keen on getting this merged in, as we think it would be a great addition. So let us know when you get a chance to revisit this work, and we'd be more than happy to help out with any questions or reviews! :+1:

Issue #758, opened 2022-10-01
HuggingFace speech models not supported


saeid93:
MLserver HuggingFace runtime cannot work with speech models in the batched mode as the pipeline accepts a list of arrays [(request1), (request2), (request3), (request4), (request5)] which the type of each request is a NumPy array. However, MLServer stacked the NumPy data as an array of arrays of shape (batch_size, input_data) which will result in the following error when sending to the HuggingFace pipeline. It thinks the batched inputs are multi-channel inputs rather than batched single-channel inputs.

```bash
raise ValueError("We expect a single channel audio input for AutomaticSpeechRecognitionPipeline")
ValueError: We expect a single channel audio input for AutomaticSpeechRecognitionPipeline
```

adriangonz:
Hey @saeid93 ,

For this one, isn't this a problem of shapes and expected dimensions? 

Looking at the [implementation for AutomaticSpeechRecognitionPipeline](https://huggingface.co/transformers/v4.8.0/_modules/transformers/pipelines/automatic_speech_recognition.html), it seems it expects `inputs` to have a single dimension. Therefore, wouldn't it be enough to ensure that the individual input's shape is always 1-dimensional?

If I'm not mistaken, that will ensure the batching is done across that single dimension, right?

saeid93:
Hey @adriangonz,

Indeed it is. The code you sent is for before sending them as a batch model (single inputs). However when a batch is sent the inputs to the speech model on the MLServer client side like the following:
```python
    payload = {
        "inputs": [
            {
            "name": "array_inputs",
            "shape": [1, len(input_data)],
            "datatype": "FP32",
            "data": input_data.tolist(),
            "parameters": {
                "content_type": "np"
            }
            }
        ]
    }
```
it is received in the MLServer as concatenated as an np.array of shape `(size_of_the_bach, size_of_input_data)`. This won't go through the HuggingFace model since as you said it expects a one-dimensional array. To send batches to a HugggingFace model instead of sending it as a stacked array we need to send it as a list of single dimension `[(1, size_of_input), (1, size_of_input), (1, size_of_input), ...]`. However, in the current implementation sending the above request will result in the error I mentioned in the first comment as it batches in the stacked np.array which the HuggingFace does not accept. In the pull request, I have added support for such cases. It basically turns the stacked np.array format in the batched version into the list of array inputs that is acceptable by the HugggingFace when the input is a NumPy array.

Please let me know if more information is needed.

[...]

PR #759, opened 2022-10-01
added support for huggingface runtime speech models

saeid93:
fixes: #758


adriangonz:
Following the discussion in #758, we'll be merging this one as an temporary workaround to unblock this use case. 

As always, thanks a lot for providing this fix @saeid93 !

Issue #728, opened 2022-09-15
Get metrics from the Queues in the inference in parallel/pool


alvarorsant:
Hi, 
we would like the number of the elements within the request queue in pool inside of a metric, for performance issues. It's a good idea to get this data to tune the system increasing or decreasing parallel workers and pods in Openshift.
What do you think?

adriangonz:
Hey @alvarorsant ,

That's a pretty cool idea. 

Right now, there are a number of queues which would be valuable to expose through metrics. From the top of my head, I could think of the following points (the order represents when they are hit in the processing flow):

1. AsyncIO "task list" within the main process. This is technically not a queue, but it's a point where requests may get held up initially while the main process gets a free slot to process them.  
2. Queue for adaptive batching, waiting for requests to form a full batch. This will only be present if adaptive batching is enabled though.
3. Queue for inference pool workers, waiting for a worker to pick up a new request. 
4. AsyncIO "task list" within the workers itself. Again, this is technically not a queue, but it's also another place where requests may get held up waiting for the worker to be free to run inference.

Ideally, it would be useful to instrument all 4 points, so that we know how many requests are currently being held on each one. Additionally, we could also have a 5th overarching count of many requests are currently being processed within MLServer (i.e. all requests currently between step `1.` and `4.`).

It would be great to hear your thoughts on that approach @alvarorsant .

alvarorsant:
We have already found point 2 and 3, but could you specify the exactly location of point 1 and 4 in code? We are preparing a PR for you indeed :) and we would need accurate indications.

Thanks in advance.

adriangonz:
Hey @alvarorsant ,

That's great to hear! Thanks a lot for taking the lead in this one! :rocket: 

For points `1.` and `4.`, this would be within Python's own runtime. When called, AsyncIO methods in Python get all converted into "tasks" and scheduled for execution into a list, which gets iterated through by an event loop. The longer this task list becomes, the longer it will take to execute each scheduled task. 

I haven't looked much into this one TBH, but there seem to be a few suggestions here of how to approach this one:

https://splunktool.com/how-can-i-measure-the-length-of-an-asyncio-event-loop

i.e. mainly doing something like:

```
len(asyncio.all_tasks(asyncio.get_running_loop()))
```

alvarorsant:
Hi @adriangonz,
[...]

PR #769, opened 2022-10-06
Add batch and pool queue metrics

miguelopind:
Related to points 2 and 3 of next issue: https://github.com/SeldonIO/MLServer/issues/728


adriangonz (commented):
Nice one @miguelopind! Massive thanks for taking the lead on this one! :rocket: 

I've added a couple comments below, about what would be the right metric type to use here. It would be great to hear your thoughts on those points!

Besides that, it seems the PR has been opened against one of the `release/*` branches. Could you change the base to `master` instead?

mlserver/batching/adaptive.py
```diff
@@ -52,6 +57,10 @@ async def _queue_request(
     ) -> Tuple[str, Awaitable[InferenceResponse]]:
         internal_id = generate_uuid()
 
+        with self.batch_queue_request_count.time():
+            batch_queue_size = self._requests.qsize()
+            self.batch_queue_request_count.observe(batch_queue_size)
```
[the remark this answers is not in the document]
miguelopind: We used Histogram because we needed more sampling items. Otherwise with Gauge we didn't have enough sampling points between two values. The number of elements of the queue depends on time on such a way.

mlserver/batching/adaptive.py
```diff
@@ -52,6 +57,10 @@ async def _queue_request(
     ) -> Tuple[str, Awaitable[InferenceResponse]]:
         internal_id = generate_uuid()
 
+        with self.batch_queue_request_count.time():
```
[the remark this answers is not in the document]
miguelopind: This is related with Histogram. I understand that Gauge doen't need time but with Histogram it is needed to see the trend.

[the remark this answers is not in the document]
adriangonz: Got it! Thanks for clarifying that :+1:

mlserver/parallel/worker.py
```diff
@@ -29,6 +30,10 @@ def __init__(self, requests: Queue, responses: Queue):
 
         self.__executor = None
 
+        # self.model_queue_request_count = Histogram(
```
adriangonz, line 33: nit: should we remove these comments if they are not used?

mlserver/parallel/pool.py
```diff
@@ -92,6 +97,10 @@ async def _process_responses(self):
     async def _process_response(self, response: InferenceResponseMessage):
         internal_id = response.id
 
+        queue_size = self._requests.qsize()
```
adriangonz, line 100: This one has changed a bit on the `master` branch. There's no longer a single `_requests` queue - instead, there's a separate queue per worker. This could be just added as a label on the metric though (i.e. instead of diff metric names for each worker).

adriangonz (changes requested):
Hey @miguelopind ,

Thanks for clarifying those points about using histograms instead of gauges. 

Besides that, it looks like you've based your changes on the `release/1.1.x` branch, instead of `master`. The `InferencePool` class in particular has a couple changes around how it manages the request queues. Therefore, it would be great if you could rebase from `master` and update those bits.

I've added below a high-level summary of these changes. Hopefully it shouldn't be too complex to get this PR up to speed with `master`.

mlserver/parallel/pool.py
miguelopind: Perfect thank you. Just one question, how can I add each worker queue in only one label, just with one metric in the class Worker?

adriangonz: If I understood that right, I guess the options would be:

1. Create the metric in the `Worker.__init__()` method. Main question is whether the `prometheus_client` library is OK with having the same metric name created multiple times. If that works, we could just track each request on the `Worker.send_request` method, and the label could be something like `worker=self.pid`. 
2. Create a single metric in the (new) `Dispatcher.__init__()` class, and track each request on the `Dispatcher.dispatch` method. The label could also be set like `worker=worker.pid`. 

I think `2.` is more similar to what you currently have in this PR, so it's probably an easier option. 

PS: While writing this comment, I also realised that TBH it's not that big of a deal that we label the metrics with the worker index / pid. So we could even go with `2.` and don't label the metrics.

miguelopind: We ended up choosing option 1

miguelopind: Thank you for the guide!

adriangonz (commented):
Hey @miguelopind ,

Apologies for the delay. 

Thanks for making those changes and rebasing from `master`!

Have you tested this one out BTW? My main concern is that the `worker._process_request` method runs on a separate process (the worker's process), so it would be good to validate whether this works as expected (i.e. the main process can see these values, and the metrics get aggregated across the different worker processes). We could actually just add a test similar to this one: 

https://github.com/SeldonIO/MLServer/blob/c8369919a3bad5cd0fe1779771963de6b597655b/tests/metrics/test_rest.py#L19-L45

PR #770, opened 2022-10-07
Add support for raw_input_contents and raw_output_contents in gRPC

### LINKED ISSUE #48: Add support for raw_input_contents field
Add support for top-level `raw_input_contents` in gRPC spec. This field was initially introduced to work-around some performance issues within gRPC.

You can find more details on these issues: 

- https://github.com/triton-inference-server/server/issues/1821
- https://github.com/kubeflow/kfserving/pull/998

adriangonz:
Fixes #48

## Changelog
- Update proto to include `raw_input_contents` and `raw_output_contents`
- Include new `mlserver.raw` module with some utilities to unpack / pack data into raw format
- Update gRPC converters module to take into account the new raw fields

### DIFF
--- a/mlserver/raw.py
+++ b/mlserver/raw.py
@@ -36,7 +36,7 @@ def _tensor_format(elem: InputOrOutput) -> str:
     return f"{size}{ctype}"
 
 
-def unpack_bytes(raw: bytes) -> List[bytes]:
+def _unpack_bytes(raw: bytes) -> List[bytes]:
     """
     From Triton's implementation:
         https://github.com/triton-inference-server/client/blob/6cc412c50ca4282cec6e9f62b3c2781be433dcc6/src/python/library/tritonclient/utils/__init__.py#L246-L273
@@ -55,7 +55,7 @@ def unpack_bytes(raw: bytes) -> List[bytes]:
     return elems
 
 
-def pack_bytes(unpacked: List[ListElement]) -> bytes:
+def _pack_bytes(unpacked: List[ListElement]) -> bytes:
     packed = []
     for elem in unpacked:
         as_bytes = _ensure_bytes(elem)
@@ -77,28 +77,28 @@ def _ensure_bytes(elem: ListElement) -> bytes:
     return elem
 
 
-def unpack_tensor(elem: InputOrOutput, raw: bytes) -> list:
+def _unpack_tensor(elem: InputOrOutput, raw: bytes) -> list:
     tensor_format = _tensor_format(elem)
     return list(struct.unpack(tensor_format, raw))
 
 
-def pack_tensor(elem: InputOrOutput) -> bytes:
+def _pack_tensor(elem: InputOrOutput) -> bytes:
     tensor_format = _tensor_format(elem)
     return struct.pack(tensor_format, *elem.data)
 
 
 def unpack(elem: InputOrOutput, raw: bytes) -> list:
     if elem.datatype == "BYTES":
-        return unpack_bytes(raw)
+        return _unpack_bytes(raw)
 
-    return unpack_tensor(elem, raw)
+    return _unpack_tensor(elem, raw)
 
 
 def pack(elem: InputOrOutput) -> bytes:
     if elem.datatype == "BYTES":
-        return pack_bytes(elem.data)  # type: ignore
+        return _pack_bytes(elem.data)  # type: ignore
 
-    return pack_tensor(elem)
+    return _pack_tensor(elem)
 
 
 def inject_raw(
--- a/tests/test_raw.py
+++ b/tests/test_raw.py
@@ -5,10 +5,10 @@ from typing import List
 
 from mlserver.types import RequestInput, ResponseOutput
 from mlserver.raw import (
-    pack_bytes,
-    pack_tensor,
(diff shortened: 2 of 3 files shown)


cliveseldon:
Looks good. One question when you pack back into raw contents to you remove the explicit data contents as with Triton you get one set but never both? Just wasn't clear on first glance.

mlserver/grpc/converters.py
```diff
@@ -310,6 +324,10 @@ def from_types(cls, type_object: types.InferenceResponse) -> pb.ModelInferRespon
             ],
         )
 
+        if use_raw:
+            # If using raw outputs, ensure it's set on the final object
+            model_infer_response.raw_output_contents.extend(raw)
```
sakoush, line 329: The assumption here is that raw contents is always the last field in the grpc message, which is fine with the current spec. I just wonder is the spec guarantees this is always the case? it makes sense to be always the case but just checking.

mlserver/grpc/converters.py
```diff
@@ -135,6 +136,12 @@ def to_types(cls, pb_object: pb.ModelInferRequest) -> types.InferenceRequest:
                 for out in pb_object.outputs
             ]
 
+        if pb_object.raw_input_contents:
+            # Unpack and inject raw contents into `data` fields if present
+            inject_raw(
+                inference_request.inputs, pb_object.raw_input_contents  # type: ignore
```
sakoush, line 142: it wasnt very clear that grpc takes care of returning a list of bytes in `raw_input_contents` one for each input request. I thought initially this is one big stream of bytes but maybe it is only me.

anyway perhaps add a note to make it clearer?

sakoush (approved):
Very neat! I have (general ) questions:
- how safe is it if we are using raw_contents when we are chaining services from different architecture, os (e.g. unix and windows) and/or python versions? is `struct` taking care of all of that? is it better or worse with triton?
- is it possible to combine codecs and raw_contents? i.e. how pandas dataframes are being sent if raw_contents is enabled? as numpy arrays and there is a further transform to convert to pandas df on the receiver side perhaps?
- a minor question whether the spec guarantees that raw_contents is the last field in the grpc message?

tests/test_raw.py
```diff
@@ -0,0 +1,157 @@
...
+
+def test_unpack_bytes():
+    raw = b"\x03\x00\x00\x00one\x03\x00\x00\x00two\x05\x00\x00\x00three"
+    expected = [b"one", b"two", b"three"]
+
+    unpacked = unpack_bytes(raw)
+    assert unpacked == expected
+
+
+@pytest.mark.parametrize(
```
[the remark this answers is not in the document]
adriangonz: In the test case, these are not strings but bytes arrays already.

mlserver/raw.py
```diff
@@ -0,0 +1,121 @@
...
+    return reduce(mul, shape, 1)
+
+
+def _tensor_format(elem: InputOrOutput) -> str:
+    size = _tensor_length(elem.shape)
+    ctype = _DatatypeToCtype[elem.datatype]
+    return f"{size}{ctype}"
+
+
+def unpack_bytes(raw: bytes) -> List[bytes]:
```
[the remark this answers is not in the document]
adriangonz: Maybe? I think initially these were on a separate module, and then I moved them all here. 

I could make them private though :+1:

mlserver/grpc/converters.py
sakoush: oh yes I just misread the code. it is fine.

mlserver/grpc/converters.py
sakoush: ok  fine if there is a note somewhere else.

adriangonz:
Thanks for your review @sakoush !

To answer the points on the main comment:

    - The implementation is very similar to Triton's. Bytes are actually encoded on the same way. The only difference is with tensors, where we use `struct` and they use the `tobytes()` and `frombuffer()` helpers from `numpy`. Both should be equivalent though. Architecture-wise, the `BYTES` encoding sets explicitly the endianness to ensure there're no issues. Perhaps we could investigate what `numpy` does for tensors to ensure we do the same thing? 
    -  Yes, it should. The `raw -> InferenceRequest` conversion occurs before any codecs are applied. So after that, the normal MLServer flow should kick off. Regarding how to write the request, same principle applies. You could use the `PandasCodec.encode_request()` helper and then call the `mlserver.raw.extract_raw` method to convert the normal data to raw (or you could do it using `tritonclient`, or manually). 
    - Not sure what you mean about this one. On both Triton and MLServer, the `raw_input_contents` / `raw_output_contents` fields are set on the same position of their respective protobuf objects (i.e. `7` and `6`).

sakoush:
> * `tobytes()` and `frombuffer()` helpers from `numpy`.

the ultimate test is to have mlserver -> triton inference pipeline with raw_contents perhaps which we can do with your change :)

> fields are set on the same position of their respective protobuf objects (i.e. 7 and 6).

I am not saying they are different I am just trying to think a bit in the future and consider if there are any further optimisations that we could do, especially in SCv2 pipelines if we know that the raw_contents happen to be the last thing. I dont have anything concrete yet thought that we could explore.

PR #776, opened 2022-10-10
Fix install dev

Salehbigdeli:
We need to add `pip install --editable .` just before `pip install --editable .[all]` because `mlserver` is listed in requirement of other sub-modules and this cause to install `mlserver` non-editable and from pypi.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,8 +7,6 @@ IMAGE_NAME := seldonio/mlserver
 
 install-dev:
 	pip install -r requirements/dev.txt
-	pip install --editable .
-	pip install --editable .[all]
 	for _runtime in ./runtimes/*; \
 	do \
 		pip install --editable $$_runtime; \
@@ -17,6 +15,7 @@ install-dev:
 			pip install -r $$_runtime/requirements-dev.txt; \
 		fi \
 	done
+	pip install --editable .[all]
 
 _generate: # "private" target to call `fmt` after `generate`
 	./hack/generate-types.sh
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,7 @@ IMAGE_NAME := seldonio/mlserver
 
 install-dev:
 	pip install -r requirements/dev.txt
+	pip install --editable .
 	pip install --editable .[all]
 	for _runtime in ./runtimes/*; \
 	do \
--- a/Makefile
+++ b/Makefile
@@ -7,8 +7,6 @@ IMAGE_NAME := seldonio/mlserver
 
 install-dev:
 	pip install -r requirements/dev.txt
-	pip install --editable .
-	pip install --editable .[all]
 	for _runtime in ./runtimes/*; \
 	do \
 		pip install --editable $$_runtime; \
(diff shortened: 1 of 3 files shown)


adriangonz:
Good catch! This one has been nagging me for a while.

I think that the main "culprits" here are the individual inference runtime packages, which list `mlserver` there. Therefore, would it be enough to just have a single `pip install -e .[all]` after the runtimes have been installed (i.e. just move it towards the end)?

Salehbigdeli:
> 

I agree, that's better solution.

adriangonz (approved):
Nice one @Salehbigdeli ! This one should be ready to go in (once tests are green).

adriangonz (commented):
Hey @Salehbigdeli , 

Just noticed you undid the latest changes. Did you find any issue with installing the editable version at the end?

Salehbigdeli:
> Hey @Salehbigdeli ,
> 
> Just noticed you undid the latest changes. Did you find any issue with installing the editable version at the end?

The main reason was the pipeline and I couldn't reproduce the behavior on my local machine. Other than that:
1. they aren't that different.
2. placing `pip install --editable .` in the beginning installs `mlserver` once.

adriangonz:
Hey @Salehbigdeli ,

My main concern is that the approach suggested in this PR installs MLServer twice, one after the other:

```
pip install --editable .
pip install --editable .[all]
```

So I guess the main questions from my side would be:

1. How does this fix the issue if the individual runtimes still get installed afterwards? And why is it required to install it twice?
2. Related to `1.` - to avoid someone else wondering the same in the future, could you add a comment to explain why it's done that way? 
3. What issues did you find when you changed it to a single `pip install -e .[all]` and moved it at the end?

Salehbigdeli:
> Hey @Salehbigdeli ,
> 
> My main concern is that the approach suggested in this PR installs MLServer twice, one after the other:
> 
> ```
> pip install --editable .
> pip install --editable .[all]
> ```
> 
> So I guess the main questions from my side would be:
> 
>     1. How does this fix the issue if the individual runtimes still get installed afterwards? And why is it required to install it twice?
> 
>     2. Related to `1.` - to avoid someone else wondering the same in the future, could you add a comment to explain why it's done that way?
> 
>     3. What issues did you find when you changed it to a single `pip install -e .[all]` and moved it at the end?

I tried installing everything from scratch several time to see what exactly happens. here is what I found:
First, The `[all]` in `pip install --editable .[all]` is obsolete according to log `WARNING: mlserver 1.2.0.dev6 does not provide the extra 'all'`
Second, The root of the problem is one of requirements in `huggingface` when we want to install `huggingface` runtime there is other requirement that is compatible with `mlserever==1.1.0` and pip decides to uninstall current version of `mlserver==1.2.0.dev6` and install older. this conclusion come from this log: `INFO: pip is looking at multiple versions of mlserver to determine which version is compatible with other requirements. This could take a while.
Collecting mlserver`
Third, I couldn't reproduce pipeline error on my local machine.
My conclusion:
You are right about placing `pip install --editable .` after installing runtimes.
I'll move `pip install --editable .` after installing runtimes, and if pipeline fail I can't do anything about it.

adriangonz:
Hey @Salehbigdeli ,

Thanks for those insights and thanks for taking the time to explore this one! 

Good spot about the `all` extras not being there anymore. It seems that was removed a couple weeks ago (which went under the radar). RE the pipeline, I must have missed the previous pipeline error - I thought all tests were green? Now that a new run has been kicked off, let's see what the error is (it may have been an unrelated one).

Issue #777, opened 2022-10-10
Adaptive batching not working properly with kafka


lincoln2491:
I tried to use adaptive batching for kafka request, but it's not working as I expected. I deployed pure sklearn model with below parameters:
- `max_batch_size` set for 32
- `max_batch_time` set for 1

I expected that adaptive batching should collect up to 32 requests, but no longer than one second (event which came earlier starts prediction).

I send 100 messages one after one through kafka and start listening responses. Turns out, that mlserver respect only condition (wait one second), ignore another (collect up to 32 requests) and process one kafka's request per second. This behavior is visible in logs:

Logs for sending (lot of requests send one after another):
```
2022-10-10 17:25:14.106564  0
2022-10-10 17:25:14.109390  1
2022-10-10 17:25:14.111003  2
2022-10-10 17:25:14.111354  3
2022-10-10 17:25:14.111992  4
2022-10-10 17:25:14.112228  5
2022-10-10 17:25:14.112436  6
2022-10-10 17:25:14.112651  7
2022-10-10 17:25:14.112889  8
2022-10-10 17:25:14.113128  9
2022-10-10 17:25:14.113334  10
2022-10-10 17:25:14.113532  11
2022-10-10 17:25:14.113737  12
2022-10-10 17:25:14.113941  13
2022-10-10 17:25:14.114147  14
2022-10-10 17:25:14.114345  15
2022-10-10 17:25:14.115613  16
2022-10-10 17:25:14.124184  17
2022-10-10 17:25:14.125083  18
2022-10-10 17:25:14.127668  19
2022-10-10 17:25:14.128573  20
2022-10-10 17:25:14.131943  21
2022-10-10 17:25:14.132962  22
2022-10-10 17:25:14.135086  23
2022-10-10 17:25:14.135400  24
2022-10-10 17:25:14.138327  25
2022-10-10 17:25:14.141321  26
2022-10-10 17:25:14.142113  27
2022-10-10 17:25:14.144030  28
2022-10-10 17:25:14.146169  29
2022-10-10 17:25:14.146913  30
2022-10-10 17:25:14.147137  31
2022-10-10 17:25:14.147353  32
2022-10-10 17:25:14.151361  33
2022-10-10 17:25:14.153370  34
2022-10-10 17:25:14.155445  35
```
Logs from mlserver (we can see, that mlserver have little more than one second pause after processing each request:
 ```
INFO:     Application startup complete.
[...]

PR #778, opened 2022-10-11
Fix async consuming requests

Salehbigdeli:
I tried #777 and had same problem. my suggested solution to this issue is in this PR.
Maybe fixes: #777

### DIFF
--- a/mlserver/kafka/server.py
+++ b/mlserver/kafka/server.py
@@ -1,4 +1,4 @@
-import asyncio
+from asyncio import Task
 
 from mlserver.errors import MLServerError
 from enum import Enum
@@ -12,6 +12,7 @@ from ..model import MLModel
 from .logging import logger
 from .handlers import KafkaHandlers
 from .message import KafkaMessage
+from ..utils import schedule_with_callback
 
 
 # TODO: Explore implementing custom handler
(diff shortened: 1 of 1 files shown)


mlserver/kafka/server.py
```diff
@@ -59,12 +61,15 @@ async def start(self):
     async def _consumer_loop(self):
         logger.info("Reading messages from consumer")
         async for request in self._consumer:
-            try:
-                await self._process_request(request)
-            except MLServerError as err:
-                logger.exception(f"ERROR {err.status_code} - {str(err)}")
-            except Exception as err:
-                logger.exception(f"ERROR 500 - {str(err)}")
+            asyncio.Task(self.consume_request(request))
```
adriangonz, line 64: Should we use the `mlserver.utils.schedule_with_callback` helper instead? 

That way is aligned with other async kick offs of tasks on other parts of the code base. Note that you will need to provide a callback, but this is good practice (and will also get rid of other warnings complaining that a task wasn't awaited). 

Callbacks will generally accept the resolved `Task` object as an argument. Therefore, you could write something along these lines:

```python
def _process_request_cb(self, process_request_task: Task):
	try:
	    process_request_task.result()
	except MLServerError as err:
	    logger.exception(f"ERROR {err.status_code} - {str(err)}")
	except Exception as err:
	    logger.exception(f"ERROR 500 - {str(err)}")
```

adriangonz (changes requested):
Good spot @Salehbigdeli! This should definitely fix #777 :rocket:

mlserver/kafka/server.py
Salehbigdeli: Thanks,
Good suggestion, I'll make the changes.

PR #779, opened 2022-10-11
Add ssl_verify_path to explainers

cliveseldon:
This allows server certificates to be verified in `requests` calls when MLServer might need to custom ca.crt file. The path to this ca.crt file can now be specified in the model-settings.json.

This is needed for correct TLS usage where black box explainers may call an endpoint whose certificate chain MLServer needs

### DIFF
--- a/runtimes/alibi-explain/mlserver_alibi_explain/common.py
+++ b/runtimes/alibi-explain/mlserver_alibi_explain/common.py
@@ -53,7 +53,7 @@ def convert_from_bytes(output: ResponseOutput, ty: Optional[Type] = None) -> Any
 
 # TODO: add retry and better exceptions handling
 def remote_predict(
-    v2_payload: InferenceRequest, predictor_url: str, ssl_verify_path: str
+    v2_payload: InferenceRequest, predictor_url: str, ssl_verify_path: Optional[str]
 ) -> InferenceResponse:
     verify: Union[str, bool] = True
     if ssl_verify_path != "":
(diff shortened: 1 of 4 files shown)


runtimes/alibi-explain/mlserver_alibi_explain/common.py
```diff
@@ -53,17 +53,23 @@ def convert_from_bytes(output: ResponseOutput, ty: Optional[Type] = None) -> Any
...
 
 
-def remote_metadata(url: str) -> MetadataModelResponse:
+def remote_metadata(url: str, ssl_verify_path: str) -> MetadataModelResponse:
     """Get metadata from v2 endpoint"""
-    response_raw = requests.get(url)
+    verify = True
+    if ssl_verify_path != "":
+        verify = ssl_verify_path
+    response_raw = requests.get(url, verify=verify)
```
adriangonz, line 72: Wouldn't this always verify the SSL certificate when sending a request? Would it make sense to provide a way to disable the verification?

Although, thinking about it, this is already the case atm (I imagine the default is `verify=True`) - so it doesn't change the previous behavior much.

adriangonz (approved):
Added a small comment, but I think changes look good! This should be good to go once the tests pass.

runtimes/alibi-explain/mlserver_alibi_explain/common.py
cliveseldon: yeah - `requests` defaults to verify=True.
We could allow not to but maybe in a future PR?

adriangonz: Yeah, definitely out of scope of this PR.

PR #790, opened 2022-10-17
Micro batching

RafalSkolasinski:
Add micro-batching to `mlserver infer` in order to align with v1 batch processor.

Noticeable changes:
- input/output queues now holds `List[BatchInputItem]` and `List[BatchOutputItem]`
- added separate preprocess / postprocess functions
- added `TritonRequest` class (one step closer towards Triton codecs)

### DIFF
--- a/mlserver/batch_processing.py
+++ b/mlserver/batch_processing.py
@@ -9,6 +9,7 @@ import click
 import json
 import orjson
 
+from time import perf_counter as timer
 from typing import Dict, List, Optional, Tuple
 from mlserver.batching.requests import BatchedRequests
 
@@ -16,9 +17,9 @@ from mlserver.types import InferenceRequest, InferenceResponse, Parameters
 from mlserver.codecs import NumpyCodec
 from mlserver.logging import get_logger
 
-from time import perf_counter as timer
 
 from mlserver.utils import generate_uuid
+from mlserver.batching.requests import _merge_parameters
 
 
 CHOICES_TRANSPORT = ["rest", "grpc"]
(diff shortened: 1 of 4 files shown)


mlserver/batch_processing.py
```diff
@@ -81,87 +82,155 @@ def get_headers(request_id: str) -> Dict[str, str]:
...
-    outputs = []
-    for request_output in inference_request.outputs or []:
-        new_output = httpclient.InferRequestedOutput(
-            request_output.name, binary_data=binary_data
-        )
-        outputs.append(new_output)
+        outputs = []
+        for request_output in inference_request.outputs or []:
+            new_output = httpclient.InferRequestedOutput(
+                request_output.name, binary_data=binary_data
```
adriangonz, line 110: Out of curiosity, do you know what the `binary_data` kwarg does here?

mlserver/batch_processing.py
```diff
@@ -81,87 +82,155 @@ def get_headers(request_id: str) -> Dict[str, str]:
...
+    outputs = []
+    for output_item in response["outputs"]:
+        output_name = output_item["name"]
+        output = NumpyCodec.encode_output(output_name, item.as_numpy(output_name))
+        output_parameters = output_item["parameters"]
+        if output_parameters is not None:
+            # Create copy in case this code will ever be used in another context as
+            # we are modifying the original parameters by removing unwanted key
+            output_parameters = output_parameters.copy()
+            if "binary_data_size" in output_parameters:
```
adriangonz, line 129: No harm in leaving extra parameters here BTW - although I can see how it would look weird to have many random ones on the response.

adriangonz (approved):
Changes look good! If you want to make any extra changes, let me know, and I can give it another pass. However, it all looks good as-is to me! :+1:

mlserver/batch_processing.py
RafalSkolasinski: They make sense only in the initial response from Triton when this is really encoded as binary data. 
When we call `item.as_numpy(name)` we decode that so the parameters is not longer relevant nor valid.

mlserver/batch_processing.py
RafalSkolasinski: Yes, I do! It instructs Triton to return responses as `binary_data` or not. However, [this will only work if we specify the actual parameters we want back](https://github.com/triton-inference-server/client/blob/84756de4f8245d4d333bfb91d2879a922c0b64e7/src/python/library/tritonclient/http/__init__.py#L100-L107), otherwise `tritonclient` will assume we always want binary data back (no harm as it will know how deal with them).

mlserver/batch_processing.py
```diff
@@ -81,87 +82,155 @@ def get_headers(request_id: str) -> Dict[str, str]:
...
+@dataclass
+class TritonRequest:
+    id: str
+    inputs: List[httpclient.InferInput]
+    outputs: List[httpclient.InferRequestedOutput] = None
+
+    @classmethod
+    def from_inference_request(
+        cls, inference_request: InferenceRequest, binary_data: bool
+    ) -> Tuple[str, List[httpclient.InferInput], List[httpclient.InferRequestedOutput]]:
```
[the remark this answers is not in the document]
RafalSkolasinski: I had to set it like that as `mypy` was rightfully complaining - I actually just forgot to set it properly

mlserver/batch_processing.py
adriangonz: Interesting! That's good to know.

PR #792, opened 2022-10-18
Add mlserver infer retry mechanism and few other improvements

RafalSkolasinski:
- adds retry mechanism via `--retries` flag and `MLSERVER_INFER_RETRIES` env variable
- exposes `tritonclient`'s `conn_timeout` option via  `--timeout` (`MLSERVER_INFER_CONNECTION_TIMEOUT` env variable)
- logs every 100 processed entries (not configurable) that allows to decrease verbosity in normal operation
- log errors on reading entries from input files (improved in logs + into the output file)
- log errors from failed requests (improved in logs + into the output file)
- adds parameters for minimum batch interval `--batch-interval` and random jiter `--batch-jiter`
- add support for request headers via `-H` or `MLSERVER_INFER_REQUEST_HEADERS` env variable
- add check if output file path is writable

Note: multiple headers can be set as 
```
-H 'Header1: Val1' -H 'Header2: Val2'
```
or with environmental variable
```
MLSERVER_INFER_REQUEST_HEADERS="Header1:Val1 Header2:Val2"
```

~Note: this is based on https://github.com/SeldonIO/MLServer/pull/790 so after https://github.com/SeldonIO/MLServer/pull/790 is merged I will rebase for cleaner review.~


mlserver/batch_processing.py
```diff
@@ -203,44 +249,80 @@ def postprocess_items(
...
+            triton_request,
+            headers,
+            retries,
+            worker_id,
         )
         logger.debug(f"consumer {worker_id}: received response")
     except Exception as e:
-        logger.error(f"consumer {worker_id}: failed to process task: {repr(e)}")
-        raise
+        logger.error(f"consumer {worker_id}: failed to process task: {e}")
```
adriangonz, line 309: As the above, should we change this for `logger.exception()`?

mlserver/batch_processing.py
```diff
@@ -371,10 +502,11 @@ async def process_batch(
 
     for consumer in consumers:
         consumer.cancel()
+    await asyncio.gather(*consumers, return_exceptions=True)
 
     finalizer.cancel()
-
-    await asyncio.gather(*consumers, finalizer, return_exceptions=True)
+    processed_instances = await finalizer
```
adriangonz, line 508: Good catch! I think `asyncio.gather` doesn't guarantee the order.

adriangonz (commented):
Changes look good! Just added a couple minor nitpicks - but feel free to ignore them if you don't think they're important.

mlserver/batch_processing.py
RafalSkolasinski: It actually did not matter here as this was after:
- `await` on queues' `join` returned
- these coroutines where cancelled 

I only made this change to actually get the output of of `finalizer` so I can print number of processed instances

RafalSkolasinski:
Thanks @adriangonz, I will tackle the `logger.exception()` suggestion as the follow up as I need to give it a bit more thought in which places they make sense. Mainly because some of these should be just logged as errors (like for example 500s from the server) for which stack trace do not make much sense.

PR #797, opened 2022-10-21
Move base image to UBI9

### LINKED ISSUE #789: Move to UBI base images
In order to get certified to run in OpenShift clusters, we need to use an UBI-base image (instead of the current Python base image).

adriangonz:
Fixes #789 

## Changelog
- Change base image to `ubi9/ubi-minimal`
- Remove `mlops-tempo` from base image
- Add Conda to base image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 FROM python:3.8-slim AS wheel-builder
-SHELL ["/bin/bash", "-c"]
+SHELL ["/bin/bash", "-l", "-c"]
 
 COPY ./hack/build-wheels.sh ./hack/build-wheels.sh
 COPY ./mlserver ./mlserver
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -13,35 +13,58 @@ COPY \
 # /opt/mlserver/dist folder
 RUN ./hack/build-wheels.sh /opt/mlserver/dist
 
-FROM python:3.8-slim
+FROM registry.access.redhat.com/ubi9/ubi-minimal
 SHELL ["/bin/bash", "-c"]
 
+ARG PYTHON_VERSION=3.8.13
+ARG CONDA_VERSION=4.13.0
+ARG MINIFORGE_VERSION=${CONDA_VERSION}-1
```
RafalSkolasinski, line 21: What does `-1` mean ?

Dockerfile
```diff
@@ -13,35 +13,58 @@ COPY \
...
 ARG RUNTIMES="all"
 
 ENV MLSERVER_MODELS_DIR=/mnt/models \
     MLSERVER_ENV_TARBALL=/mnt/models/environment.tar.gz \
-    PATH=/opt/mlserver/.local/bin:$PATH
+    MLSERVER_PATH=/opt/mlserver \
+    CONDA_PATH=/opt/conda \
+    PATH=/opt/mlserver/.local/bin:/opt/conda/bin:$PATH
```
RafalSkolasinski, line 28:
```suggestion
    PATH=${MLSERVER_PATH}/.local/bin:${CONDA_PATH}/bin:$PATH
```

Dockerfile
```diff
@@ -13,35 +13,62 @@ COPY \
...
+RUN mkdir $MLSERVER_PATH 
 WORKDIR /opt/mlserver
 
 # Create user and fix permissions
 # NOTE: We need to make /opt/mlserver world-writable so that the image is
 # compatible with random UIDs.
-RUN useradd -u 1000 -s /bin/bash mlserver -d /opt/mlserver && \
-    chown -R 1000:0 /opt/mlserver && \
-    chmod -R 776 /opt/mlserver
+RUN useradd -u 1000 -s /bin/bash mlserver -d $MLSERVER_PATH && \
```
RafalSkolasinski, line 63: hope we will not get any issues when OpenShift will be running the images using random user

RafalSkolasinski (approved):
LGTM

Dockerfile
adriangonz: I tried this at first, but it seemed the Dockerfile didn't recognise them? Easiest was to just hardcode it and figure it out later.

Dockerfile
adriangonz: I tried it out manually, and it seems to be OK. The tests also run the images with random user IDs to verify just this.

Dockerfile
RafalSkolasinski: Cool, make sense, thanks for confirmation!

Dockerfile
RafalSkolasinski: Awesome, good to know!

PR #799, opened 2022-10-23
Alibi Explain Update to work with Hugging face runtime

cliveseldon:
1. Exclude unset fields when sending inference requests to limit issues on deserialization on the server end
2. Set use_bytes when creating input to send to ensure it can be JSON serialized
3. Exclude SCV2 pipelines when asking for metadata as this is not presently possible with Pipelines
4. Use `args` not `*args` in Huggingface runtime as `args` is always a single argument or list. Allows batch inputs to work.

### DIFF
--- a/runtimes/huggingface/mlserver_huggingface/runtime.py
+++ b/runtimes/huggingface/mlserver_huggingface/runtime.py
@@ -90,11 +90,17 @@ class HuggingFaceRuntime(MLModel):
         # TODO: convert and validate?
         kwargs = self.decode_request(payload, default_codec=MultiStringRequestCodec)
 
-        args = kwargs.pop("args", [])
+        args = kwargs.pop("args", None)
+
         X = kwargs.pop("array_inputs", None)
         if X is not None:
             args = [list(X)] + args
-        prediction = self._model(args, **kwargs)
+        if args is None:
+            prediction = self._model(**kwargs)
+        else:
+            prediction = self._model(args, **kwargs)
+
+        logger.debug("Prediction %s", prediction)
 
         # TODO: Convert hf output to v2 protocol, for now we use to_json
         if isinstance(prediction, dict):
--- a/runtimes/huggingface/mlserver_huggingface/runtime.py
+++ b/runtimes/huggingface/mlserver_huggingface/runtime.py
@@ -84,6 +84,9 @@ class HuggingFaceRuntime(MLModel):
         TODO
         """
 
+        # Adding some logging as hard to debug given the many types of input accepted
+        logger.debug("Payload %s",payload)
+
         # TODO: convert and validate?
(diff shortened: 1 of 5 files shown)


runtimes/huggingface/mlserver_huggingface/runtime.py
```diff
@@ -84,14 +84,17 @@ async def predict(self, payload: InferenceRequest) -> InferenceResponse:
...
+
         # TODO: convert and validate?
         kwargs = self.decode_request(payload, default_codec=MultiStringRequestCodec)
 
         args = kwargs.pop("args", [])
         X = kwargs.pop("array_inputs", None)
         if X is not None:
             args = [list(X)] + args
-        prediction = self._model(*args, **kwargs)
+        prediction = self._model(args, **kwargs)
```
axsaucedo, line 97: Actually the *args is necessary as there are some parameters that need to be passed as arg as opposed to as a list. From memory this is the case for the sentiment analysis model, where it would fail if a list is passed https://mlserver.readthedocs.io/en/latest/examples/huggingface/README.html#sentiment-analysis

axsaucedo: Are there currently errors due to the args not being passed as `*args` instead of a list?

cliveseldon: So allowing it to be a list allows batch which is needed for alibi-explain

adriangonz (commented):
Changes look good!

Not sure what's the right thing to do RE `args` vs `*args` though...

cliveseldon:
> Changes look good!
> 
> Not sure what's the right thing to do RE `args` vs `*args` though...

I think the arg is addressed above and is probably better than it was until we have the full PR to handle all Huggingface types.

runtimes/huggingface/mlserver_huggingface/runtime.py
cliveseldon: See [here](https://github.com/huggingface/transformers/blob/674f750a57431222fa2832503a108df3badf1564/src/transformers/pipelines/base.py#L1019-L1022) which shows if args has multiple values they are ignored.

cliveseldon:
I have had a look at all [Pipelines](https://github.com/huggingface/transformers/tree/674f750a57431222fa2832503a108df3badf1564/src/transformers/pipelines) and all support a single argument other arguments are passed as kwargs or for more complex use cases the single argument can be a dict or list of dict. So as far as I can see the change to a single arg does not limit the use of any Pipeline.

The most complex I could find is the [Table Question Answering Pipeline](https://huggingface.co/docs/transformers/main_classes/pipelines#transformers.TableQuestionAnsweringPipeline) which still allows a dict or list of dict as the single argument.

PR #826, opened 2022-11-02
Add worker_queue_request, batch_queue and worker_processes list metrics

alvarorsant:
Add worker_queue_request, batch_queue and worker_processes_list metrics


mlserver/batching/adaptive.py
```diff
@@ -75,9 +83,8 @@ def _start_batcher_if_needed(self):
                 # If task hasn't finished yet, let it keep running
                 return
 
-        self._batching_task = schedule_with_callback(
-            self._batcher(), self._batching_task_callback
-        )
+        self._batching_task = asyncio.create_task(self._batcher())
```
adriangonz, line 86: What was the reason to replace the `schedule_with_callback` call? Under the hood, `schedule_with_callback` will essentially call the same things (it's essentially a helper to avoid duplicating these two lines).

mlserver/parallel/dispatcher.py
```diff
@@ -93,7 +106,19 @@ def _get_worker(self) -> Worker:
...
+    def _workers_queue_monitor(self, worker: Worker, worker_pid: int):
+        """Get metrics from every worker request queue"""
+        queue_size = worker._requests.qsize()
+        
+        queue_request_count.labels(workerpid=str(worker_pid)).observe(
+            float(queue_size)
+        )
+    
+    def _workers_processes_monitor(self, loop: AbstractEventLoop):
+        process_request_count.observe(float(len(asyncio.all_tasks(loop))))
```
adriangonz, line 120: This `loop` would only represent the main process' loop. So it wouldn't capture the workers' queued asyncio tasks. 

For this first iteration, I'd keep it simple and focus just on tracking the queue.

adriangonz (changes requested):
Hey @alvarorsant ,

Thanks for picking back up this effort :+1: 

I've added a couple comments below. Overall, I feel we should keep these changes small, and focus just on tracking the requests queues. 

Besides that, and the review comments, could you also add a couple tests to ensure the new metrics get exposed correctly? You can model these off the tests under `tests.metrics`, like the following one:

https://github.com/SeldonIO/MLServer/blob/f5297d42f00be647954bbed8b12b0bea29c00994/tests/metrics/test_rest.py#L19-L45

mlserver/parallel/dispatcher.py
alvarorsant: So I will focus on only in queue requests in this PR. I will do a couple of tests.

mlserver/batching/adaptive.py
alvarorsant: I didn't touch that part. It must be a matter of updating issue to the master.

mlserver/parallel/dispatcher.py
alvarorsant: Regarding to the tests, I noticed you mount an aiohttp server in a fixture however I'd need to test the metrics inside the code in mlserver folder and I don't know how to fit real metrics in a mocked server.

mlserver/batching/adaptive.py
adriangonz: Is there any chance this was a leftover from the previous PR? Either way, could you try to rebase from `master` to see if it goes away?

mlserver/parallel/dispatcher.py
adriangonz: Wouldn't it be the same as in the current metrics tests? Those spin up an actual MLServer instance, send a couple requests, and then scrape the metrics endpoint to compare the metrics before and after (and assert that they've increased).

PR #828, opened 2022-11-02
Set explicitly a two-dimensional shape on some codecs

### LINKED ISSUE #818: Return explicit batch shape in codecs
Currently, codecs will return a very minimal shape (e.g. `[2]` for `["foo", "bar"]`). This makes the "data-dimension" itself implicit. 

To make it more clear, it would be better if codecs returned a shape (e.g. `[2, 1]`) with both shapes (i.e. batch and data dimensions).

adriangonz:
Fixes #818 

## Changelog
- To avoid room for ambiguity, where a shape is set to `[N]`, convert it to `[N, 1]` (which is MLServer's assumption). This will be switched on in some codecs (`StringCodec`, `Base64Codec`, `DatetimeCodec` and `PandasCodec`).

### DIFF
--- a/docs/user-guide/content-type.md
+++ b/docs/user-guide/content-type.md
@@ -206,13 +206,13 @@ Therefore, we can leverage this to override the model's metadata when needed.
 Out of the box, MLServer supports the following list of content types.
 However, this can be extended through the use of 3rd-party or custom runtimes.
 
-| Python Type                           | Content Type | Request Level | Request Codec                        | Input Level | Input Codec                     |
-| ------------------------------------- | ------------ | ------------- | ------------------------------------ | ----------- | ------------------------------- |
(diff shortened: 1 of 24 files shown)


RafalSkolasinski:
What this means for examples that does not specify / talk about codecs? For example [sklearn](https://mlserver.readthedocs.io/en/latest/examples/sklearn/README.html) or [xgboost](https://mlserver.readthedocs.io/en/latest/examples/xgboost/README.html) runtimes?

adriangonz:
@RafalSkolasinski all runtimes should use codecs under the hood. Although in those cases, they'll use the `NumpyCodec`, which I haven't changed.

My main concern is that Numpy arrays already specify a shape explicitly, so I'm wary of changing that. So they'll keep doing whatever they do at the moment.

RafalSkolasinski:
But this would mean that artefacts deployed using these runtime would still return ambiguous outputs shaped `[N]` which is the issue we wanted to solve in the first place.

PR #839, opened 2022-11-13
Allow MLflow URI with scheme

regen100:
MLflow can load [many schemes](https://github.com/mlflow/mlflow/blob/aa9366483e5e5f2fd933fe479a837d219958a1bc/mlflow/store/artifact/artifact_repository_registry.py#L78-L94). However, `get_model_uri` accepts only local file paths.

This PR enables `MLflowRuntime` to load models by URIs such as `s3://...` and `file:...`.

### DIFF
--- a/mlserver/utils.py
+++ b/mlserver/utils.py
@@ -1,6 +1,7 @@
 import os
 import uuid
 import asyncio
+import urllib.parse
 
 from asyncio import Task
 from typing import Callable, Dict, Optional, List, Type
@@ -12,7 +13,9 @@ from .errors import InvalidModelURI
 
 
 async def get_model_uri(
-    settings: ModelSettings, wellknown_filenames: List[str] = []
+    settings: ModelSettings,
+    wellknown_filenames: List[str] = [],
+    allowed_schemes: List[str] = [],
 ) -> str:
     if not settings.parameters:
         raise InvalidModelURI(settings.name)
@@ -20,6 +23,12 @@ async def get_model_uri(
     model_uri = settings.parameters.uri
     if not model_uri:
         raise InvalidModelURI(settings.name)
+
+    scheme = urllib.parse.urlparse(model_uri).scheme
+    if scheme != "":
+        if scheme in allowed_schemes:
+            return model_uri
+        raise InvalidModelURI(settings.name, model_uri)
 
     full_model_uri = _to_absolute_path(settings._source, model_uri)
     if os.path.isfile(full_model_uri):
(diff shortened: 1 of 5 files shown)


runtimes/mlflow/tests/conftest.py
```diff
@@ -96,6 +96,15 @@ def model_settings(model_uri: str) -> ModelSettings:
     )
 
 
+@pytest.fixture
+def model_settings_scheme(model_uri: str) -> ModelSettings:
+    return ModelSettings(
+        name="mlflow-model",
+        implementation=MLflowRuntime,
+        parameters=ModelParameters(uri="file:" + model_uri),
```
adriangonz, line 104: Rather than adding a new fixture, should we parametrise one of the tests to run with `file://` or no-schema. That way we add less boilerplate code to maintain later.

runtimes/mlflow/mlserver_mlflow/runtime.py
```diff
@@ -151,7 +151,10 @@ async def invocations(
 
     async def load(self) -> bool:
         # TODO: Log info message
-        model_uri = await get_model_uri(self._settings)
+        model_uri = await get_model_uri(
+            self._settings,
+            allowed_schemes=mlflow.store.artifact.artifact_repository_registry._artifact_repository_registry._registry,
+        )
         self._model = mlflow.pyfunc.load_model(model_uri)
```
adriangonz, line 158: I don't have much experience downloading things from `dbfs://` / `s3://` using MLflow, but would `load_model()` just handle everything else? How would we configure the DBFS / S3 host, credentials, etc?

mlserver/utils.py
```diff
@@ -21,6 +24,12 @@ async def get_model_uri(
     if not model_uri:
         raise InvalidModelURI(settings.name)
 
+    scheme = urllib.parse.urlparse(model_uri).scheme
```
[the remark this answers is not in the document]
regen100: Great idea! I rewrote the code to use `file` as the default scheme.

runtimes/mlflow/tests/conftest.py
regen100: I refactored the tests by pytest parameterization.

runtimes/mlflow/mlserver_mlflow/runtime.py
regen100: In the case of S3 URLs, MLflow retrieves the artifact through `boto3`, which retrieves the credentials from an environment variables `AWS_ACCESS_KEY_ID` or the default credential store `~/.aws`.

regen100:
> we try to shy away from downloading artefacts in MLServer, as we see it as more of an infrastructure concern

Your policy is reasonable. Downloading artifacts introduces complexity such as cache control.

adriangonz (approved):
Changes look good! Nice one defaulting the scheme on the `urlparse()` method :+1: 

Once the tests are green, this should be good to go.

Issue #821, opened 2022-10-31
Custom Model Repository


jgallardorama-itx:
Hello!

We would like to extend the model repository to enable fetch models from other sources (database, services, etc...).
Have you thought how to enable this customation or extension?
For example, if ModelRepository was a interface, and an injection mechanishm like with a model is provided, it could enable this feature. What is your roadmap about this feature?

Thank you in advance.

Javi

adriangonz:
Hey @jgallardorama-itx,

Thanks for opening this one.

In general, we prefer to implement the download of external artefacts through upstream orchestration engines, like Seldon Core or KServe. Main reason is that, in production setups, you generally don't want the inference server to spend time performing the download, and is best to delegate that on other infra pieces. 

Having said that, I'd be keen to allow the user to inject other implementations of the `ModelRepository` class. I could see this being configured on the `settings.json` file, which could read a Python import path (e.g. `mlserver.repository.ModelRepository`) and use that as the repository class. If these repository implementations inferface with external sources, they would probably go on a diff package though.

jgallardorama-itx:
Hi team!

I opened a PR about this issue https://github.com/SeldonIO/MLServer/pull/849 

Please, could you reopen this issue, sorry for the inconveniences

adriangonz:
Just saw that PR - sure thing, will reopen right ahead :+1:  Thanks a lot for your work on the PR!

adriangonz:
Fixed by #849

PR #849, opened 2022-11-15
Feature/custom model repository

jgallardorama:
Custom model repository:
- By parameter in the ML Server constructor
- By settings with the implementation and parameters settings

Both ways to extend the behaviors allow the custom model repository to handle the persistence model settings externally to mlserver.


mlserver/repository/repository.py
```diff
@@ -13,7 +14,17 @@
...
+    @abc.abstractmethod
+    async def list(self) -> List[ModelSettings]:
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    async def find(self, name: str) -> List[ModelSettings]:
+        raise NotImplementedError
+
+
+class ImplModelRepository(ModelRepository):
```
adriangonz, line 27: Should we rename `ImplModelRepository` to something more descriptive about the repository implementation? The main feature of the current implementation is that it doesn't enforce any folder layout or schema, so I was thinking `SchemalessModelRepository`? Or maybe just `DefaultRepository`? Names are hard...

mlserver/cli/serve.py
```diff
@@ -34,7 +36,7 @@ async def load_settings(
 
     models_settings = []
     if settings.load_models_at_startup:
-        repository = ModelRepository(settings.model_repository_root)
+        repository = ImplModelRepository(settings.model_repository_root)
```
adriangonz, line 39: Should this use the `ModelRepository` implementation defined in the `settings` instead?

mlserver/repository/repository.py
```diff
@@ -13,7 +14,17 @@
 DEFAULT_MODEL_SETTINGS_FILENAME = "model-settings.json"
 
 
-class ModelRepository:
+class ModelRepository(metaclass=abc.ABCMeta):
```
adriangonz, line 17: Nice one using `abc`! :+1:

mlserver/server.py
```diff
@@ -22,7 +22,9 @@
 
 
 class MLServer:
-    def __init__(self, settings: Settings):
+    def __init__(
+        self, settings: Settings, model_repository: Optional[ModelRepository] = None
```
adriangonz, line 26: To keep the interface simple, I would remove the `model_repository` kwarg and just rely on what comes from the `settings` object. This is also consistent with the rest of parameters and flags.

mlserver/settings.py
```diff
@@ -87,6 +87,10 @@ class Config:
     model_repository_root: str = "."
     """Root of the model repository, where we will search for models."""
 
+    model_repository_implementation: Optional[PyObject]
```
adriangonz, line 90: We could make it non-`Optional` and set it by default to `mlserver.repository.DefaultRepository` (or however we decide to call it). 

BTW to make it less verbose we could also call the field just `model_repository` and `model_repository_args`? This is a bit nitpicky so feel free to ignore it though.

jgallardorama-itx:
Hi team!

I opened a PR about this issue https://github.com/SeldonIO/MLServer/pull/849 

Please, could you reopen this issue, sorry for the inconveniences

tests/custom_md_repo/dummymdrepo.py
```diff
@@ -0,0 +1,38 @@
...
+    DEFAULT_MODEL_SETTINGS_FILENAME,
+)
+from mlserver.settings import ModelSettings
+from mlserver.errors import ModelNotFound
+
+from ..fixtures import SumModel, ErrorModel, SimpleModel
+
+
+class DummyModelRepository(ModelRepository):
+    def __init__(self, extra_params: Optional[dict] = None) -> None:
```
adriangonz, line 14: To make it more of an "alternative reference implementation", should we just unpack the `files` kwarg directly here? As in,

```python
def __init__(self, root: str, files: List[str]):
```

tests/custom_md_repo/conftest.py
```diff
@@ -0,0 +1,75 @@
+import os
```
adriangonz, line 1: I was wondering, should we move the `tests/custom_md_repo` tests to the existing `tests/repository` folder? They seem to fit "semantically" well there (as in, that's where I'd look for a test of a custom repo). 

We can probably add the `conftest.py`, the `dumymdrepo.py` fixture and the `test_custom_md_repo.py` there, at the top level folder.

tests/custom_md_repo/test_custom_md_repo.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+async def test_custom_model_repo_by_param(
+    settings: Settings, prometheus_registry: CollectorRegistry
+):
+    model_repository = DummyModelRepository(
+        {"files": ["tests/custom_md_repo/testdata/model-settings.json"]}
+    )
+
+    aserver = get_ml_server(settings, prometheus_registry, model_repository)
```
adriangonz, line 44: Wouldn't it be possible to just use the fixture from `conftest.py`?

tests/custom_md_repo/testdata/model-settings.json
```diff
@@ -0,0 +1,34 @@
+{
```
adriangonz, line 1: We don't need this one, right?

tests/custom_md_repo/testdata/settings-custom-md-repo.json
```diff
@@ -0,0 +1,18 @@
...
+    "allow_origins": [
+      "*"
+    ]
+  },
+  "model_repository_implementation": "tests.custom_md_repo.dummymdrepo.DummyModelRepository",
+  "model_repository_implementation_args": {
+    "files": [
+      "tests/custom_md_repo/testdata/model-settings.json"
+    ]
+  }
```
adriangonz, line 17: That looks pretty cool!

mlserver/server.py
```diff
@@ -54,7 +56,18 @@ def __init__(self, settings: Settings):
             on_model_reload=on_model_reload,  # type: ignore
             on_model_unload=on_model_unload,  # type: ignore
         )
-        self._model_repository = ModelRepository(self._settings.model_repository_root)
+
+        if model_repository:
+            self._model_repository = model_repository
+        elif self._settings.model_repository_implementation:
+            self._model_repository = self._settings.model_repository_implementation(
+                self._settings.model_repository_implementation_args
```
adriangonz, line 64: Given that `model_repository_implementation_args` is a dict, should we just unpack it and pass it as kwargs (i.e. with the `**` thingy)?

adriangonz, line 64: I was thinking that we could move this logic to the `mlserver.repository` submodule, as a sort of factory that takes in the `settings` object and returns a `ModelRepository` implementation?

adriangonz (changes requested):
Hey @jgallardorama ,

Thanks a lot for taking the lead on this one! :rocket: 

This is an awesome job! I've added a couple suggestions and questions below, but the main idea looks great! Would be good to hear your thoughts on those comments. :+1:

adriangonz:
Fixes #821

mlserver/repository/repository.py
jgallardorama-itx: I'll do it. SchemalessModelRepository

mlserver/cli/serve.py
jgallardorama-itx: My bad!, I will fix it

mlserver/repository/repository.py
jgallardorama-itx: thanks

mlserver/settings.py
jgallardorama-itx: I got the idea, but I don't know exactly how you want to code it. I will try it and then you correct me if I'm wrong.

No problem, I'll rename the settings.

mlserver/server.py
jgallardorama-itx: I like it!, thanks.

tests/custom_md_repo/testdata/settings-custom-md-repo.json
jgallardorama-itx: :)

mlserver/server.py
jgallardorama-itx: I can do a first approach, and then we'll see.

mlserver/repository/repository.py
jgallardorama-itx: Done

mlserver/cli/serve.py
jgallardorama-itx: Done

mlserver/server.py
jgallardorama-itx: Done

mlserver/settings.py
jgallardorama-itx: I kept model_repository_implementation and model_repository_implementation_args. I thought It could be good to have different identifiers for different things

tests/custom_md_repo/dummymdrepo.py
jgallardorama-itx: Done

tests/custom_md_repo/conftest.py
jgallardorama-itx: sorry,... I didn't this change because I don't know how to reuse conftest methods using a different settings with custom settings. Any idea how to reuse fixtures "mlserver" "rest_client" with a different settings?

tests/custom_md_repo/test_custom_md_repo.py
jgallardorama-itx: Done

mlserver/server.py
jgallardorama-itx: Done

jgallardorama-itx: Done

tests/custom_md_repo/testdata/model-settings.json
adriangonz: Unless it changes anything over the root one, I would just let the tests reuse the other one.

tests/custom_md_repo/conftest.py
adriangonz: Good point! I think that the `prometheus_registry`, `settings` and `rest_client` fixtures have probably earned their place on the root `conftest.py`, as they're already used in a couple other test modules. That's completely out of the scope of this PR though - it may come in a future one though. 

So I think for now, we can just move both the fixtures and the tests jointly into a `tests/repository/test_custom_repository.py` file (don't worry - once I make the changes above, I'll remove the redundant ones so that it's only the `settings` fixture in there, which is the one that we're interested in modifying).

mlserver/settings.py
adriangonz: A `PyObject` can point to any Python object, so I think that it should be possible to set something like this in the `Settings` object:

```python
from mlserver.repository import SchemalessModelRepository

model_repository_implementation: PyObject = SchemalessModelRepository
```

mlserver/repository/repository.py
```diff
@@ -65,3 +76,17 @@ async def find(self, name: str) -> List[ModelSettings]:
...
+
+
+class ModelRepositoryFactory:
+    @staticmethod
+    def resolve_model_repository(settings: Settings) -> ModelRepository:
+        result: ModelRepository
+        if settings.model_repository_implementation:
+            result = settings.model_repository_implementation(
+                **settings.model_repository_implementation_args
+            )
```
adriangonz, line 88: To respect the current interface (and the current `model_repository_root` setting), would it make sense to always pass the root folder as a first argument, then followed by `model_repository_implementation_args`?

mlserver/settings.py
```diff
@@ -87,6 +87,10 @@ class Config:
...
+
+    model_repository_implementation_args: dict = {}
```
adriangonz, line 92: As an extra note on the settings, would it be possible to add a comment below each of them? That will ensure it shows up on the generated docs.

adriangonz (changes requested):
Hey @jgallardorama,

Thanks a lot for addressing those initial comments! Besides those, I've added a couple new minor ones. We're getting close though - almost there!

tests/custom_md_repo/conftest.py
jgallardorama-itx: Ok, I'll do it

mlserver/repository/repository.py
jgallardorama-itx: Well, I was thinking about how to avoid a compatibility issue with the signature of the default and other modelRepository classes. I mean, now SchemalessModelRepository expects model_repository_root from settings, and on the other hand, model_repository_implementation expects parameters from settings. So how do conciliate both interfaces? I think I shouldn't make a breaking change with the previous implementation. I can write the code but please could you give the expected behavior with the default implementation (previous implementation) and the new one? 
My two cents:
```
if settings.model_repository_root:
  model_repository_implementation_args["file"] = settings.model_repository_root

```
That way, we could handle the default model repository and how to pass args. How do you see it?

mlserver/settings.py
jgallardorama-itx: sure, I'll do it

mlserver/settings.py
jgallardorama-itx: Got it, I have a new problem, about default implementation and signing constructors, I explained below.

tests/custom_md_repo/testdata/model-settings.json
jgallardorama-itx: Done

tests/custom_md_repo/conftest.py
jgallardorama-itx: Done

mlserver/settings.py
jgallardorama-itx: Done

mlserver/repository/repository.py
jgallardorama-itx: Hi @adriangonz ,

My apologies for my delay too, day-to-day!

I see both sides of the solution, if you prefer to keep the parameter, no big deal, I can keep it. IMO it can be confusing to have a parameter that the implementation doesn't always use. I will update code with your choice.

mlserver/settings.py
jgallardorama-itx: Hi 
I don't think I can do this assigment, because SchemalessModelRepository uses the module settings, so I cannot set model_repository_implementation with the Value SchemalessModelRepository, because I got a circular reference. I can move this default assignment to ModelRepositoryFactory implements creation/construction of model_repository object.
Any alternative?

jgallardorama-itx: I upload the option Optional[PyObject], I hope it's a showstopper for the PR, but tell me if you find a better alternative

mlserver/repository/repository.py
jgallardorama-itx: Done

adriangonz:
BTW on the linter errors, it seems to be mainly complaining about the following errors:

```
flake8 .
./mlserver/repository/repository.py:6:1: F401 'typing.Optional' imported but unused
./tests/repository/conftest.py:14:1: F401 '..utils.get_available_ports' imported but unused
./tests/repository/conftest.py:14:1: E402 module level import not at top of file
./tests/repository/dummymdrepo.py:1:1: F401 'os' imported but unused
./tests/repository/test_custom_md_repo.py:1:1: F401 'asyncio' imported but unused
./tests/repository/test_custom_md_repo.py:2:1: F401 'os' imported but unused
./tests/repository/test_custom_md_repo.py:3:1: F401 'time.sleep' imported but unused
```

You can run them locally by calling `make lint`, which may speed up your feedback loop (i.e. as opposed to waiting for the CI to finish each time).

adriangonz:
Hey @jgallardorama ,

It seems there's still one linter check failing:

```
mypy ./mlserver
mlserver/repository/factory.py:12: error: "object" not callable  [operator]
```

Once that's sorted, this one should be good to go :+1:

jgallardorama-itx:
Hi @adriangonz 

My bad, I only executed Flake for the previous comment. This time I have run `make lint` and everything looks good, but let me know if anything is missing.

PR #853, opened 2022-11-18
Fix issue with empty dimensions

adriangonz:
As a follow up to #818, Alibi doesn't seem to like model response tensors shaped as `[N, 1]` (i.e. with an explicit dimensionality shape). To work around these, process response tensors within the Alibi runtime to detect and remove these explicit empty shapes.

### DIFF
--- a/runtimes/alibi-explain/mlserver_alibi_explain/explainers/black_box_runtime.py
+++ b/runtimes/alibi-explain/mlserver_alibi_explain/explainers/black_box_runtime.py
@@ -83,4 +83,12 @@ class AlibiExplainBlackBoxRuntime(AlibiExplainRuntimeBase):
             ssl_verify_path=self.ssl_verify_path,
         )
         # TODO: do we care about more than one output?
-        return NumpyCodec.decode_output(v2_response.outputs[0])
+        decoded = NumpyCodec.decode_output(v2_response.outputs[0])
+
+        if decoded.ndim > 1 and decoded.shape[-1] == 1:
+            # Assume the response tensor is shaped as `[N, 1]` and remove the
+            # explicit dimensionality shape (which Alibi doesn't seem to like)
+            last_dim = decoded.ndim - 1
+            return decoded.squeeze(axis=last_dim)
+
+        return decoded


runtimes/alibi-explain/mlserver_alibi_explain/explainers/black_box_runtime.py
```diff
@@ -83,4 +83,12 @@ def _infer_impl(self, input_data: Union[np.ndarray, List[str]]) -> np.ndarray:
             ssl_verify_path=self.ssl_verify_path,
         )
         # TODO: do we care about more than one output?
-        return NumpyCodec.decode_output(v2_response.outputs[0])
+        decoded = NumpyCodec.decode_output(v2_response.outputs[0])
+
+        if decoded.ndim > 1 and decoded.shape[-1] == 1:
```
sakoush, line 88: is it possible to add a test case?

sakoush (approved):

runtimes/alibi-explain/mlserver_alibi_explain/explainers/black_box_runtime.py
adriangonz: Apologies, browser didn't refresh reviews until I pushed the merged button :disappointed: 

I think this bit should be covered by some of the tests we've got there (not the exact case, but close enough).

sakoush: I am actually curious what the existing alibi runtime tests didnt capture this issue?

adriangonz: What the tests don't seem to capture is cases where the return shape is like `[N, 1]` . These are cases where the ML framework (e.g. sklearn) would return just `[N]`, but the recent change in the codecs makes the data-dimensionality explicit. 

In these cases, Alibi seemed to run into issues when getting the prediction back from the model (as it would return a np array with shape `[N, 1]`). 

Just to clarify, this is still not captured by the tests. It would be great to add one, but don't have the capacity now (still fixing other things).

PR #864, opened 2022-11-22
feat(rest): log HTTP server listening event

rio:
Implemented exactly like the gRPC message.


rio:
No problem. I have some more lined up, sorry in advance ;)

PR #872, opened 2022-11-25
updated grpc example based on recent updates

saeid93:
updates based on https://github.com/SeldonIO/MLServer/pull/785


adriangonz (changes requested):
Nice one @saeid93! 

Changes look good! I think the only ask is whether you would be able to re-generate the  markdown corresponding to the notebook? This can be done as:

```
make -C docs/examples all
```

adriangonz (approved):
Thanks for re-generating the markdown! This looks good to go!

PR #888, opened 2022-12-07
fix(codecs): do not crash when the content type is not registered

rio:
Any unexpected input in the content_type parameter field on either input or request will crash the server otherwise and return a 500.


adriangonz (approved):
Nice spot @rio!

Thanks for fixing this one! Oh and nice one for adding a test as well! :rocket:

PR #912, opened 2022-12-19
do not run benchmark job on forks

RafalSkolasinski:
This job often run on forks producing unnecessary noise 
image


adriangonz:
lol good spot!

PR #914, opened 2022-12-22
Adding conda unpack option to docker build / entrypoint

dtpryce:
Trying to remove as much logging as possible but allowing for development / debugging. I believe this should remove all the unpack logs from the pod start

### DIFF
--- a/hack/activate-env.sh
+++ b/hack/activate-env.sh
@@ -10,6 +10,8 @@ if [ "$#" -ne 1 ]; then
   exit 1
 fi
 
+quiet=${3:-false}
+
 _unpackEnv() {
   local _envTarball=$1
   local _envFolder=$2
@@ -27,6 +29,7 @@ _unpackEnv() {
 _activateEnv() {
   local _envFolder=$1
   local _activate="$_envFolder/bin/activate"
+  local _quiet=$quiet
 
   if ! [[ -f $_activate ]]; then
     echo "Environment not found at '$_envFolder'"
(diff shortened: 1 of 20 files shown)


hack/activate-env.sh
```diff
@@ -40,7 +41,10 @@ _activateEnv() {
   set -u
 
   echo "--> Calling conda-unpack..."
-  conda-unpack
+  if [ $quiet ]; then
+    conda-unpack --quiet
```
adriangonz, line 45: To keep things simple, should we always run `conda-pack` in quiet mode and remove the `_quiet` arg?

dtpryce: True! Would that be ok for everyone? Does anyone use those logs?

dtpryce: @adriangonz ^

adriangonz: It should be alright to remove them :+1:

adriangonz (approved):
Nice one @dtpryce ! Looks great now!

adriangonz:
Hey @dtpryce ,

It seems tests are failing due to an issue that has now been fixed in `master`. Would you be able to rebase?

dtpryce:
> Hey @dtpryce ,
> 
> It seems tests are failing due to an issue that has now been fixed in `master`. Would you be able to rebase?

I think I have done it ... if this fails, I can make a new fork and re-request :D

adriangonz:
Nice one! Looks all it's green! :rocket:

PR #916, opened 2022-12-23
common batch parameters stored in a list

### LINKED ISSUE #915: Issue in aggregating batched requests parameters
When sending batched requests with custom parameters (using extra field in [here](https://github.com/SeldonIO/MLServer/blob/749cd02bc5f3ee0b915151eedeff50075ac946ad/mlserver/types/dataplane.py#L27)) like the following:

```python
# request 1
payload = types.InferenceRequest(
    inputs=[
        types.RequestInput(
            name="parameters-np",
            shape=[1],
            datatype="BYTES",
            data=[],
            parameters=types.Parameters(
                custom-param='value-1,
        )
    ]
)
# request 2
payload = types.InferenceRequest(
    inputs=[
        types.RequestInput(
            name="parameters-np",
            shape=[1],
            datatype="BYTES",
            data=[],
            parameters=types.Parameters(
                custom-param='value-2',
        )
    ]
)
```
the results param field for the batched inference request will only contain a single `custom-param` field with the last request value for that, this is because a [dictionary is used](https://github.com/SeldonIO/MLServer/blob/749cd02bc5f3ee0b915151eedeff50075ac946ad/mlserver/batching/requests.py#L112) to aggregate parameters and for the case of having a same parameter with different values in the same batch only one value will be stored for them. E.g. in the case above we will only have `custom-param` equals to `value-2`. A better solution would be able to store all the values of common request parameter separately in a list e.g. `{custom-param=['value-1', 'value-2']`

saeid93:
fixes #915

aggregate batch requests of:
```python
# request 1
payload = types.InferenceRequest(
    inputs=[
        types.RequestInput(
            name="parameters-np",
            shape=[1],
            datatype="BYTES",
            data=[],
            parameters=types.Parameters(
                custom-param='value-1,
        )
    ]
)
# request 2
payload = types.InferenceRequest(
    inputs=[
        types.RequestInput(
            name="parameters-np",
            shape=[1],
            datatype="BYTES",
            data=[],
            parameters=types.Parameters(
                custom-param='value-2',
        )
    ]
)
```
as follows:
`{custom-param=['value-1', 'value-2']`
instead of overwriting the `value-1` with `value-2`

### DIFF
--- a/mlserver/batching/requests.py
+++ b/mlserver/batching/requests.py
@@ -45,15 +45,22 @@ def _merge_input_parameters(
             "content_type",
             "headers",
         }
-        all_params = {**all_params, **obj_params}
+        uncommon_keys = set(all_params).union(set(obj_params)) - common_keys
+        new_all_params = {}
         for key in common_keys:
             if type(all_params[key]) == list:
-                all_params[key] = all_params[key].append(obj_params[key])
+                new_value = all_params[key] + [obj_params[key]]
+                new_all_params[key] = new_value
             else:
                 new_value = [all_params[key]]
                 new_value.append(obj_params[key])
-                all_params[key] = new_value
-    return all_params
+                new_all_params[key] = new_value
+        for key in uncommon_keys:
+            if key in all_params.keys():
+                new_all_params[key] = all_params[key]
+            if key in obj_params.keys():
+                new_all_params[key] = obj_params[key]
+    return new_all_params
 
 
 def _merge_data(
--- a/mlserver/batching/requests.py
+++ b/mlserver/batching/requests.py
@@ -29,6 +29,33 @@ def _merge_parameters(
     return {**all_params, **obj_params}
 
 
+def _merge_input_parameters(
+    all_params: dict,
+    parametrised_obj: Union[
+        InferenceRequest, InferenceResponse, RequestInput, RequestOutput
+    ],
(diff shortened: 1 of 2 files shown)


adriangonz (approved):
Nice one @saeid93 ! Thanks for scratching your head with this one! :rocket: 

It's quite a tricky problem. I could see issues if some of the batch requests are missing parameters and then the cardinality of some of the lists of different to the batch size. At the same time, there's not much we can do there... 

I think this solution is as much as can be done from the MLServer side, and then it's up to the user to use their own parameters convention consistently :+1:

saeid93:
Thank you @adriangonz ! About the cardinality yes it is problem, I first tried to have a per request dictionary style but since the codecs are dependent on some of the fields that wasn't possible. I think maybe we can clarify this somewhere in example notebooks as **custom parameters** and explaining the caveats, I'll be happy to add it to documentation but I don't know where could be the best fit for explaining this, if you find it necessary please let me know which example notebook you think is best for it and I can add the explanation.

adriangonz:
Rather than a notebook, I think we could just add a note block to the adaptive batching docs. This could just clarify briefly how the parameter merging logic works:

https://github.com/SeldonIO/MLServer/blob/master/docs/user-guide/adaptive-batching.md

PR #922, opened 2022-12-30
bugfix, set a default image format when Pillow image has no format

pepesi:
1. Error happend when a run a `image-segmantation` task. The Pillow Image pipeline output has not format. So add a default one
2. If client give a invalid content_type, the default_codec may be None, caused runtime error


adriangonz (approved):
Nice catch @pepesi ! Looks great to me :+1: 

Once tests are green, this should be good to go.

PR #930, opened 2023-01-03
support huggingface translation task suffix

### LINKED ISSUE #947: The huggingface translation task name requires an extra suffix
When I run the Huggingface translation task, got this raise ValueError('The task defaults can\'t be correctly selected. You probably meant "translation_XX_to_YY"')

pepesi:
Fixes #947

When I run huggingface task translation, got this `raise ValueError('The task defaults can\'t be correctly selected. You probably meant "translation_XX_to_YY"')`

So, I add a new setting item in HuggingfaceSetting, to support task name suffix.

### DIFF
--- a/runtimes/huggingface/mlserver_huggingface/common.py
+++ b/runtimes/huggingface/mlserver_huggingface/common.py
@@ -38,6 +38,9 @@ class HuggingFaceSettings(BaseSettings):
         env_prefix = ENV_PREFIX_HUGGINGFACE_SETTINGS
 
     task: str = ""
+    # Why need this filed?
+    # for translation task, required a suffix to specify source and target
+    # related issue: https://github.com/SeldonIO/MLServer/issues/947
     task_suffix: str = ""
     pretrained_model: Optional[str] = None
     pretrained_tokenizer: Optional[str] = None
@@ -47,7 +50,9 @@ class HuggingFaceSettings(BaseSettings):
 
     @property
     def task_name(self):
-        return f"{self.task}{self.task_suffix}"
+        if self.task == "translation":
+            return f"{self.task}{self.task_suffix}"
+        return self.task
 
 
 def parse_parameters_from_env() -> Dict:
--- a/runtimes/huggingface/mlserver_huggingface/common.py
+++ b/runtimes/huggingface/mlserver_huggingface/common.py
@@ -38,11 +38,16 @@ class HuggingFaceSettings(BaseSettings):
         env_prefix = ENV_PREFIX_HUGGINGFACE_SETTINGS
 
     task: str = ""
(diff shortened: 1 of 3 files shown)


runtimes/huggingface/mlserver_huggingface/common.py
```diff
@@ -38,12 +38,17 @@ class Config:
         env_prefix = ENV_PREFIX_HUGGINGFACE_SETTINGS
 
     task: str = ""
+    task_suffix: str = ""
```
adriangonz, line 41: Wouldn't it be enough to include the suffix in the `task` field?

pepesi: Sure, but for other reasons.

Some default metadatas is associated with task types. 

https://github.com/SeldonIO/MLServer/blob/c07a1702581abd348fb7c30fd7e8c663b3505bb2/runtimes/huggingface/mlserver_huggingface/metadata.py#L179

The task field does not specify whether it represents the task type or the task name. 

Maybe there is a better way. This only solves the problem I have encountered

adriangonz: Right, I see... Thanks for clarifying that.

Do you know whether hugginface uses other variations of the task name besides the suffix? If that's the case, we could go for something more generic, like regexes for the metadata dict keys? 

If, on the other hand, it's only suffixes - then I think it should be alright to just add this extra key to the config.

runtimes/huggingface/mlserver_huggingface/common.py
```diff
@@ -38,12 +38,17 @@ class Config:
...
     pretrained_model: Optional[str] = None
     pretrained_tokenizer: Optional[str] = None
     optimum_model: bool = False
     device: int = -1
     batch_size: Optional[int] = None
 
+    @property
+    def task_name(self):
```
adriangonz, line 49: Could we also add a test for the `task_name` property? This is mainly to ensure no one changes it randomly in the future.

runtimes/huggingface/mlserver_huggingface/common.py
adriangonz: PS: Could we also add a note as to why we need `task_suffix`, perhaps referencing the original issue?

pepesi: From the comment in transforms lib. https://github.com/huggingface/transformers/blob/6767ce71d661ae57afa38b6cd0d533e3dfd463fa/src/transformers/pipelines/__init__.py#L257

This is a special case for task `translation`.

adriangonz (approved):
Nice one! This looks great now @pepesi :+1:

PR #931, opened 2023-01-04
Add rest request gzip decompress

gawsoftpl:
I add gzip decompress for requests if has header Content-Encoding


gawsoftpl:
Fastapi handles GZipMiddleware but only for Response. For a request with gzip payload and header Content-encoding: gzip Fastapi requires additional Middleware to decompress that payload. 

https://fastapi.tiangolo.com/advanced/custom-request-and-route/#create-a-custom-gziprequest-class

mlserver/rest/gzip_request.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class GzipRequest(Request):
+    async def body(self) -> bytes:
+        if not hasattr(self, "_body"):
+            body = await super().body()
+            if "gzip" in self.headers.getlist("Content-Encoding"):
+                body = gzip.decompress(body)
+            self._body = body
+        return self._body
```
adriangonz, line 14: We already have a custom `Request` object defined in https://github.com/SeldonIO/MLServer/blob/master/mlserver/rest/requests.py. Would it be possible to add the `body` method there, and remove the need for the `GzipRoute` router?

adriangonz (changes requested):
Got it! Thanks for clarifying the difference.

I've added a small comment to simplify the implementation. Other than that, this looks good!

mlserver/rest/gzip_request.py
gawsoftpl: I will check that, if that will work I will make a new pull request.

gawsoftpl: I create gzip decompress in Custom Request class (https://github.com/SeldonIO/MLServer/blob/master/mlserver/rest/requests.py.)
I will make new pull requests. I close this one

PR #934, opened 2023-01-09
fixing FastAPI incompatibility

saeid93:
Fixing MLServer blocking incompatibility with the recent update of FastAPI https://github.com/tiangolo/fastapi/releases/tag/0.89.0


adriangonz (approved):
Great spot @saeid93!

PR #941, opened 2023-01-09
Change default for gRPC max message length

adriangonz:


adriangonz:
Changed in https://github.com/SeldonIO/seldon-core/pull/4571 instead.

PR #946, opened 2023-01-11
Only convert pd.DataFrame to numpy array in alibi-explain

sakoush:
Currently we have been converting all inputs to numpy before passing them to alibi `explain` method. This is not true in the case of `AnchorText` as it expects an `str`, check [here](https://docs.seldon.io/projects/alibi/en/latest/api/alibi.explainers.html#alibi.explainers.AnchorText.explain).

Although we should be perhaps more strict in the way we do the conversions between types but for now we know we want to convert pd.DataFrame to numpy which we do explicitly.


adriangonz (approved):
Looks good!

PR #956, opened 2023-01-17
fix(logging): switch eventloop log to debug

rio:
This log statement is executed before the logging settings are applied and can trip up systems that don't expect logs to be formatted differently.


adriangonz (approved):
Good spot!

Issue #139, opened 2021-02-24
Add lockfile to MLServer


adriangonz:
To better support use cases where `mlserver` is used as a library, we shouldn't restrict the dependencies versions too much. Instead, we should look into adding some sort of lockfile (e.g. through Poetry or Conda) that locks the versions in the Docker image (so that the "app-level" environment is kept consistent).

PaulEdwardBrennan:
Hi Adrian. Can you please give some more detailed explanation on what's involved here. 
Thank you.

adriangonz:
Hey @PaulEdwardBrennan , 

This would involve adding a lock file which specifies the exact deps and subdeps that are shipped within each MLServer image. This ensures reproducibility, as well as avoiding scenarios where a subdep of a subdep introduces a breaking change. To achieve this, we'll most likely leverage [Poetry](https://python-poetry.org/). 

Do you have any particular questions around this one?

RafalSkolasinski:
I am gonna look into this one in following days

adriangonz:
Good point! Let me pull that into `v1.1.x`.

PR #970, opened 2023-01-25
Update security pipeline

adriangonz:
Snyk has issues working with complex setup.py files (e.g. https://github.com/snyk/cli/issues/1367).

To account for this, we set up the environment and then dump it with `pip freeze` into a `requirements.txt`. Note that this environment will include both production and development dependencies.

Once we move into Poetry (#139 ), this workaround can be replaced for the lock file.


RafalSkolasinski (commented):
LGTM!

PR #971, opened 2023-01-26
Add support for Prometheus metrics across multiple processes

### LINKED ISSUE #303: Move Tempo's insight logger to MLServer
Tempo currently has an implementation for an Insights Logger, which lets its internal runtime log metrics and payloads. Since this is relative low-level, it could be interesting to move this feature to MLServer, so that every inference runtime can leverage it.

adriangonz:
Part of #303 

## Notes

Note that this work just enables the collection of metrics across multiple workers, which lets users define Prometheus metrics directly on their custom runtimes. However, we may still introduce a simplified way to register metrics within the MLServer package itself (e.g. `self.metrics.register('foo', 1)`) - which is why this PR doesn't close #303 (and also why there is no docs update in this PR). 

## Changelog
- Enable Prom's multiprocess mode when `parallel_workers > 0`, adding a new `metrics_dir` setting.
- Clean up `*.db` files during shutdown.
- Add tests around manual set up of custom metrics in custom runtimes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@
 *.dylib
 *.pyc
 bin
+*.db
 
 # Mac file system
 **/.DS_Store
--- a/mlserver/handlers/dataplane.py
+++ b/mlserver/handlers/dataplane.py
@@ -16,21 +16,6 @@ from prometheus_client import (
 from typing import Optional
 
 
-_ModelInferRequestSuccess = Counter(
-    "model_infer_request_success",
-    "Model infer request success count",
-    ["model", "version"],
-)
-_ModelInferRequestFailure = Counter(
-    "model_infer_request_failure",
-    "Model infer request failure count",
-    ["model", "version"],
-)
-_ModelInferRequestDuration = Summary(
-    "model_infer_request_duration", "Model infer request duration", ["model", "version"]
-)
-
-
 class DataPlane:
     """
     Internal implementation of handlers, used by both the gRPC and REST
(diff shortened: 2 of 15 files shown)


sashman:
This will be a game changer 🥰

sashman:
@adriangonz question about usage

given 

```python
class CustomMetricsModel(MLModel):
    async def load(self) -> bool:
        self.counter = Counter(COUNTER_NAME, "Test custom counter")
        self.ready = True
        return self.ready

    async def predict(self, req: InferenceRequest) -> InferenceResponse:
        self.counter.inc()
        return InferenceResponse(model_name=self.name, outputs=[])
```

`self.counter = Counter(COUNTER_NAME, "Test custom counter")` is a class variable, would it be possible to use a `Counter` variable outside of the the `CustomMetricsModel` scope? e.g.

```
counter = Counter(COUNTER_NAME, "Test custom counter")

class SomethingElse:
    def callback():
        counter.inc()

class CustomMetricsModel(MLModel):
    async def load(self) -> bool:
        self.ready = True
        some_middleware_register(SomethingElse())
        return self.ready

    async def predict(self, req: InferenceRequest) -> InferenceResponse:
        return InferenceResponse(model_name=self.name, outputs=[])
```

adriangonz:
Hey @sashman ,

That's not the final interface. 

This PR only enables usage of raw Prom metrics within models loaded on parallel workers, but there will extra changes before it's out.

Issue #721, opened 2022-09-07
Support multiple environments


adriangonz:
At the moment, MLServer allows you to inject a [custom environment](https://mlserver.readthedocs.io/en/latest/examples/conda/README.html). However, this environment is then used for every loaded model.

Ideally, it would be great if MLServer also supported running different custom environments for each loaded model.

PR #992, opened 2023-02-10
Add support to run Worker with a separate environment

adriangonz:
Part of #721. 

We still need to add top-level support to manage diff pools with diff environments. However, in the interest of keeping PRs small, this will come in future PRs.

## Changelog
- Add new `Environment` class
- Add an "env" parameter to the `Worker` class, which will start the worker using the custom environment

## Notes
One of the limitations we've found is that the version of Python **must match** between the main process and the worker processes. 
There's a chance we could walk around this if we dig deeper into the `multiprocessing` library and stop using some of the primitives it offers (e.g. like the `Process` class) - however that would be a chunky piece of work.


mlserver/env.py
```diff
@@ -0,0 +1,74 @@
...
+from .logging import logger
+
+
+def _extract_env(tarball_path: str, env_path: str) -> None:
+    logger.info(f"Extracting environment tarball from {tarball_path}...")
+    with tarfile.open(tarball_path, "r") as tarball:
+        tarball.extractall(path=env_path)
+
+
+class Environment:
```
sakoush, line 19: question: How extensible is this class? for example if we wanted `Environment` to be defined from a `conda.yaml` or `requrirements.txt` or `name` etc.? It might be for a separate PR but we probably need to have a factory method, similar to `from_tarball` but more general...

mlserver/parallel/worker.py
```diff
@@ -52,11 +58,16 @@ def _executor(self):
...
 
     def run(self):
-        install_uvloop_event_loop()
-        configure_logger(self._settings)
-        configure_metrics(self._settings)
-        self._ignore_signals()
-        asyncio.run(self.coro_run())
+        ctx = nullcontext()
+        if self._env:
+            ctx = self._env
```
sakoush, line 63: nit: should this be inline?

mlserver/env.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def __init__(self, env_path: str):
+        self._env_path = env_path
+
+    @classmethod
+    async def from_tarball(cls, tarball_path: str, env_path: str) -> "Environment":
+        loop = asyncio.get_running_loop()
+        await loop.run_in_executor(None, _extract_env, tarball_path, env_path)
+
+        return cls(env_path)
```
sakoush, line 34: not sure if this is part of this class but we probably want to make sure that this is a valid env?

sakoush (approved):
Great! I left some comments but in general looks good.

mlserver/env.py
```diff
@@ -0,0 +1,74 @@
...
+        return os.path.join(self._env_path, "bin")
+
+    @cached_property
+    def lib_path(self) -> str:
+        pattern = os.path.join(self._env_path, "lib", "python*")
+        matches = glob.glob(pattern)
+
+        for match in matches:
+            if os.path.isdir(match):
+                return match
```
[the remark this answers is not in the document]
adriangonz: Good point! Let me document those a bit better.

mlserver/env.py
```diff
@@ -0,0 +1,74 @@
...
+
+        return ""
+
+    def __enter__(self):
+        self._prev_sys_path = sys.path
+        self._prev_bin_path = os.environ["PATH"]
+
+        sys.path = [*self.sys_path, *self._prev_sys_path]
+        os.environ["PATH"] = os.pathsep.join([self.bin_path, self._prev_bin_path])
```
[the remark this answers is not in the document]
adriangonz: The custom paths will come in  before the existing ones. So the custom version of duplicated packages should come in earlier (which should be covered on the tests).

We could have the option to completely disable the base path, in which case the custom environment would be isolated, but I would leave that for a future PR as the behaviour may be more nuanced than that (e.g. we may want an extra flag that lets the user decide whether the new env should run isolated or not).

mlserver/parallel/worker.py
adriangonz: I've got mixed opinions about Python's inline syntax TBH. In this case, I felt it was clearer to have it separate.

mlserver/env.py
adriangonz: It's hard to check whether something is a valid env. However, the class should already handle gracefully when some of the environment base folders are missing.

mlserver/env.py
```diff
@@ -0,0 +1,74 @@
...
+
+    @classmethod
+    async def from_tarball(cls, tarball_path: str, env_path: str) -> "Environment":
+        loop = asyncio.get_running_loop()
+        await loop.run_in_executor(None, _extract_env, tarball_path, env_path)
+
+        return cls(env_path)
+
+    @cached_property
+    def sys_path(self) -> List[str]:
```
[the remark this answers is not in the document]
sakoush: Actually I meant to ask first whether they should be private methods?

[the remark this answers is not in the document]
adriangonz: That's a great point! We probably just need the constructors as public. Everything else can be private, since it will mainly be used as a context manager.

PR #1007, opened 2023-02-21
Ability to pass extra parameters to the models in huggingface

pascalwhoop:
I want to pass extra parameters like `device_map` and `load_in_8bit`. So I thought this is the lightest way to get that done

settings file is now

```
{
    "name": "opt-125m",
    "parameters": {
        "extra": {
            "model_parameters": {
                "device_map": "auto",
                "load_in_8bit": true
            },
            "task": "text-generation",
            "device": "0",
            "pretrained_model": "facebook/opt-125m"
        }
    }
}
```

thoughts?

### DIFF
--- a/runtimes/huggingface/mlserver_huggingface/common.py
+++ b/runtimes/huggingface/mlserver_huggingface/common.py
@@ -13,6 +13,7 @@ from transformers.models.auto.tokenization_auto import AutoTokenizer
 import transformers
 from mlserver.logging import logger
 
+
 from optimum.pipelines import SUPPORTED_TASKS as SUPPORTED_OPTIMUM_TASKS
 
 
--- a/runtimes/huggingface/mlserver_huggingface/common.py
+++ b/runtimes/huggingface/mlserver_huggingface/common.py
@@ -10,6 +10,8 @@ from mlserver.errors import MLServerError
 from transformers.pipelines import pipeline
 from transformers.pipelines.base import Pipeline
 from transformers.models.auto.tokenization_auto import AutoTokenizer
+import transformers
+from mlserver.logging import logger
 
 from optimum.pipelines import SUPPORTED_TASKS as SUPPORTED_OPTIMUM_TASKS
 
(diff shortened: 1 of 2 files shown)


runtimes/huggingface/mlserver_huggingface/common.py
```diff
@@ -42,10 +45,12 @@ class Config:
     # for translation task, required a suffix to specify source and target
     # related issue: https://github.com/SeldonIO/MLServer/issues/947
     task_suffix: str = ""
+    auto_loader_name: Optional[str] = "AutoModel"
```
adriangonz, line 48: Should we use a `PyObject` instead? You can check out the `Settings` class implementation for an example on how it could be used:

https://github.com/SeldonIO/MLServer/blob/27ca2573c740109e28da69690c8341260161b6aa/mlserver/settings.py#L90

Also, as a separate point, is `auto_loader_name` the best field name? Should it be just `model_loader`?

runtimes/huggingface/mlserver_huggingface/common.py
```diff
@@ -122,6 +132,11 @@ def load_pipeline_from_settings(hf_settings: HuggingFaceSettings) -> Pipeline:
         # Device needs to be set to -1 due to known issue
         # https://github.com/huggingface/optimum/issues/191
         device = -1
+    else:
+        logger.debug("loading model without optimum")
+        model = _get_model_loader(hf_settings).from_pretrained(model, **hf_settings.model_parameters)
+        #model = AutoModel.from_pretrained(model, **hf_settings.model_parameters)
```
adriangonz, line 138: Should we remove the older version without the model loader?

adriangonz (commented):
Hey @pascalwhoop ,

Thanks a lot for the contribution! :rocket: 

I think this could be a great addition. I've added a couple comments, but besides that, the changes look good.

The only point left to add would be some tests around the new logic. E.g. "does changing the model loader work?", "do the model parameters make their way through?".

adriangonz:
Hey @pascalwhoop ,

Following up on this one, since there hasn't been any activity for a few weeks, we'll be closing this PR for now. 

Having said that, we're still super keen on getting this merged in, as we think it would be a great addition. So let us know when you get a chance to revisit this work, and we'd be more than happy to help out with any questions or reviews! :+1:

Issue #359, opened 2021-10-19
Improve docs for Alibi Explain runtime


seldondev:
- [ ] Add example on serving explainers through Alibi Explain runtime
- [ ] Add docs around extra config required to serve explainers.

PR #1017, opened 2023-02-27
Alibi-explain runtime `kernel_shap` fixes

### LINKED ISSUE #1008: Add ability to allow batch explanation
Currently we assume in mlserver `alibi-explain` runtime that the user wants to explain an individual inference prediction (no batch). This is true in the case of anchors. However this is a restriction given that other explainers allow for batch, e.g. `kernel-shap`.

One is idea is to allow the user to specify whether an explainer can handle batch in `model-settings.json`

A related point to `kernel-shap` is that the user is likely to want to explain `predict-proba`, which is sometime not the default predictor in inference models.

A possible example for such `model-settings.json` is as follows:

```
{
    "name": "ks_wine",
    "implementation": "mlserver_alibi_explain.AlibiExplainRuntime",
    "parallel_workers": 0,
    "parameters": {
        "uri": "../artifacts/ks_wine",
        "version": "v0.1.0",
        "extra": {
            "explainer_type": "kernel_shap",
            "infer_uri": "http://localhost:8080/v2/models/wine/infer",
            "infer_output": "predict_proba",
            "explainer_batch": true,
        }
    }
}
```

sakoush:
fixes #1008 

This PR adds two extra optional configuration fields:

- `explainer_batch`: whether an explainer can consume a batch (e.g. in the case of `kernel_shap`)
- `infer_output`: allowing the user to specify which output to explain in the case of multi-output inference models


runtimes/alibi-explain/mlserver_alibi_explain/explainers/white_box_runtime.py
```diff
@@ -37,7 +37,7 @@ async def load(self) -> bool:
             init_parameters["model"] = self._inference_model
             self._model = self._explainer_class(**init_parameters)  # type: ignore
         else:
-            self._model = self._load_from_uri(self._inference_model)
+            self._model = await self._load_from_uri(self._inference_model)
```
adriangonz, line 40: Good catch!

runtimes/alibi-explain/mlserver_alibi_explain/common.py
```diff
@@ -92,6 +91,10 @@ class Config:
 
     infer_uri: str
     explainer_type: str
+    explainer_batch: Optional[bool] = False
```
adriangonz, line 94: Would it be worth opening an issue to document the Alibi runtime settings? We could use the same pattern we use for `Settings` and `ModelSettings` (i.e. leveraging some autodoc utils from Pydantic).

adriangonz (approved):
Looks great! Added a couple comments, but I don't think they're within the scope of this PR.

runtimes/alibi-explain/mlserver_alibi_explain/common.py
sakoush: Should we re-purpose https://github.com/SeldonIO/MLServer/issues/359 to add this kind of documentation?

runtimes/alibi-explain/mlserver_alibi_explain/explainers/black_box_runtime.py
```diff
@@ -53,12 +53,14 @@ async def load(self) -> bool:
...
-        # current implementation we will only explain the first data element.
-        input_data = input_data[0]
+        if not self.alibi_explain_settings.explainer_batch:
+            # if we get a list of strings, we can only explain the first elem and there
+            # is no way of just sending a plain string in v2, it has to be in a list
+            # as the encoding is List[str] with content_type "BYTES"
+            # we also assume that the explain data will contain a batch dimension,
+            # and in current implementation we will only explain the first data element.
+            # this is for explainers that do not support batch, e.g. anchors
+            input_data = input_data[0]
```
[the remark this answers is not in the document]
sakoush: Yeah we have a brief discussion with @jklaise about this loop and although we did not conclude whether it is a runtime or library responsibility, it feels that we certainly do it here and perhaps in parallel. I will create an Issue for this.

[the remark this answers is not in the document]
sakoush: Created an issue: https://github.com/SeldonIO/MLServer/issues/1018 but I am not entirely sure yet at which layer this could be done.

jklaise:
LGTM!

RobertSamoilescu (approved):
LGTM!

PR #1020, opened 2023-03-02
Metrics Interface

### LINKED ISSUE #303: Move Tempo's insight logger to MLServer

adriangonz:
Fixes #303 

Follow-up to #971 to introduce a simplified interface to interact with custom metrics, as well as accompanying docs to show how to use them.


adriangonz:
Hey @sashman!

This PR introduces a simple interface to register and interact with custom metrics. It would be great to get your thoughts on this one ;)

sashman:
Looking sweet @adriangonz ! Couple of questions:

> Under the hood, metrics logged through the {func}`mlserver.log` method will get
exposed to Prometheus as a Histogram.

* Would it be possible to expose a Counter as well? Or override the type somehow?
* Speaking of histograms specifically, would it be possible to control the bucket config for the metric? We currently have some of the default metrics capped at 10 as max value which is not ideal for monitoring.

sashman:
From https://github.com/prometheus/client_python#histogram, 

> The default buckets are intended to cover a typical web/rpc request from milliseconds to seconds. They can be overridden by passing buckets keyword argument to Histogram.

Unfortunately, our "typical" requests vary hugely in duration :)

adriangonz:
Hey @sashman ,

That's a great point. 

Fully agree that Histograms are not a great fit for many cases. We've already got some ideas to introduce support for other types of custom metrics. However, this will most likely get tackled under a diff issue (see #1030). 

There're a few other things on the backlog though, so it's not 100% clear whether that one will be tackled in time for `1.3.0` or will need to wait until `1.4.0`.

PR #1054, opened 2023-03-21
Support batch accumulation within Detect runtime

### LINKED ISSUE #1019: Support drift batch accumulation in Detect runtime
Drift models generally need to operate on a batch of inputs. This batch would usually get accumulated on-the-fly with a "moving window" over past requests that is kept within the runtime. 

Note that this type of batch is different from the adaptive batching support built into MLServer. On the latter, all request connections are held while the batch is getting formed (and only returned after batch inference occurs), whereas in the former requests are answered as they come (with the batch kept up-to-date in the background).

adriangonz:
Fixes #1019

## Changelog

- Add the `batch_size` flag to the Detect settings.
- Use the `BatchedRequests` utility class within the Detect runtime to accumulate batches.

### DIFF
--- a/mlserver/batching/__init__.py
+++ b/mlserver/batching/__init__.py
@@ -1,4 +1,5 @@
+from .requests import BatchedRequests
 from .adaptive import AdaptiveBatcher
 from .hooks import load_batching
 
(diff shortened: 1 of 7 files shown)


SachinVarghese (approved):
LGTM

PR #1056, opened 2023-03-23
Refactor model wrapper

sakoush:
Move the runtime wrapper to a common utility file.

### DIFF
--- a/runtimes/alibi-explain/tests/conftest.py
+++ b/runtimes/alibi-explain/tests/conftest.py
@@ -234,6 +234,7 @@ async def anchor_image_runtime_with_remote_predict_patch(
             ),
         )
     )
+    assert isinstance(rt, MLModel)
     await rt.load()
 
     yield rt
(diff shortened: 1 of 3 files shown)


mlserver/model_wrapper.py
```diff
@@ -0,0 +1,78 @@
...
+from mlserver.types import (
+    MetadataTensor,
+    RequestInput,
+    InferenceRequest,
+    MetadataModelResponse,
+    InferenceResponse,
+)
+
+
+class WrapperMLModel(MLModel):
```
adriangonz, line 15: Do you remember why we needed this instead of just extending the `MLModel` class?

sakoush: It is essentially so that we have different implementations of `MLModel` where inner `MLModel` in this wrapper is created by a factory method and allowing the user to have in the `model-settings.json` one reference in `implementation` to this WrapperMLModel. For example in `mlserver-alibi-explain` runtime it is going to be `AlibiExplainRuntime` for ally types of alibi models.

sakoush: I simplified the code based on your suggestion to use `__new__` instead

adriangonz (approved):
This looks great! I love when there are more lines removed than added :grin: 

image

PR #1058, opened 2023-03-25
merged batch size variables of huggingface runtime

saeid93:
Fixes #740


saeid93:
Hey @adriangonz , the test error seems  seems to be unrelated to the code changes. Is that right?

adriangonz (approved):
Looks great! 

Thanks for following up with this one @saeid93. :rocket:

PR #1064, opened 2023-03-29
Fix signature codec with tuple type annotations

### LINKED ISSUE #1036: decode_args with tuple return value
I'm confused about how to use `decode_args()` when the model returns a tuple of, let's say, a numpy array.

If I have an inference function with the following signature

```python
import numpy as np
from mlserver.codecs.decorator import decode_args

def predict(input: np.ndarray) -> tuple[np.ndarray]:
    # simple identity model
    return (input, )

decode_args(predict)
```
I receive a

```
CodecNotFound: Codec not found for output field with type tuple[numpy.ndarray]
```

error message.

I think the issue lies in the following line that is used to unwrap the tuple signature and convert it to a list of types.

https://github.com/SeldonIO/MLServer/blob/9d7045cc3c50a14201733865ca0c3e8480ebd942/mlserver/codecs/decorator.py#L37

However the line checks if the type hint is a tuple instance, which is not true in the above case: `isinstance(tuple[np.ndarray], tuple) == False)`

To make `decode_args` work with a tuple return signature, I have to change the function signature to

```python
def predict(input: np.ndarray) -> tuple([np.ndarray]):
```
which is an actual tuple instance and not a type hint. This looks like a bug to me.

sauerburger:
Fixes #1036 

I'm unsure how/where to add a test for this as I didn't manage to run the test suite. Locally, I've experimented with the following snippet.

```python
from typing import Tuple
import numpy as np
from mlserver.codecs.decorator import SignatureCodec

def predict_fn_typing_tuple() -> Tuple[np.ndarray, np.ndarray]:
    return (np.array([4]), np.array([2]))


def predict_fn_native_tuple() -> tuple[np.ndarray, np.ndarray]:
    return (np.array([4]), np.array([2]))


def predict_fn_tuple_object() -> (np.ndarray, np.ndarray):
    return (np.array([4]), np.array([2]))


def test_typing_tuple():
    signature_codec = SignatureCodec(predict_fn_typing_tuple)
    assert signature_codec._output_hints == [np.ndarray, np.ndarray]


def test_native_tuple():
    signature_codec = SignatureCodec(predict_fn_native_tuple)
    assert signature_codec._output_hints == [np.ndarray, np.ndarray]


def test_tuple_object():
    signature_codec = SignatureCodec(predict_fn_tuple_object)
    assert signature_codec._output_hints == [np.ndarray, np.ndarray]
```

Please note: The notation for the `predict_fn_native_tuple` test requires Python 3.9. The PR code, however, does not depend on Python 3.9.

### DIFF
--- a/mlserver/codecs/decorator.py
+++ b/mlserver/codecs/decorator.py
@@ -38,6 +38,10 @@ def _as_list(a: Optional[Union[Any, Tuple[Any]]]) -> List[Any]:
         # Split into components
         return list(a)
 
+    if get_origin(a) is tuple:
+        # Convert type arguments into list
+        return list(get_args(a))
+
     # Otherwise, assume it's a single element
     return [a]
 
--- a/tests/codecs/test_decorator.py
+++ b/tests/codecs/test_decorator.py
@@ -2,7 +2,7 @@ import pytest
 import numpy as np
 import pandas as pd
 
-from typing import Any, Callable, Dict, Optional, List
+from typing import Any, Callable, Dict, Optional, List, Tuple
 
 from mlserver.types import (
     InferenceRequest,
@@ -13,7 +13,7 @@ from mlserver.types import (
 )
 from mlserver.codecs.base import InputCodec
 from mlserver.codecs.utils import Codec
-from mlserver.codecs.decorator import SignatureCodec
+from mlserver.codecs.decorator import SignatureCodec, _as_list
 from mlserver.codecs.errors import InputsNotFound, OutputNotFound
 from mlserver.codecs.numpy import NumpyCodec, NumpyRequestCodec
 from mlserver.codecs.string import StringCodec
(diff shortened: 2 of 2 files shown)


adriangonz:
Nice one! Thanks a lot for following up with this PR @sauerburger .

Changes look great! :rocket:  

For the tests, perhaps you can add a new one for the `_as_list` method, covering the different ways to specify the type hint? We run tests in 3.8 - 3.10, so we'll need to skip the `tuple[np.ndarray, np.ndarray]` form unfortunately. Happy to help out writing the test if you've got any questions.

adriangonz (approved):
Nice one @sauerburger ! Thanks for making those changes. 

PR looks great. Once tests are green, it should be good to go ahead. :+1:

sauerburger:
Thanks for the kind feedback. I've fixed the issue triggered the linter.

PR #1102, opened 2023-04-17
warning message for batch corner case

saeid93:
This corner case has happened to me many times. It took some time to root cause analysis. I think a warning message could help to avoid confusion for other users.


adriangonz (approved):
Nice addition @saeid93 !

PR #1108, opened 2023-04-19
Add support for online drift detectors

ascillitoe:
This PR adds support for [online drift detectors](https://docs.seldon.io/projects/alibi-detect/en/stable/cd/methods.html#online) to Alibi Detect runtime. 

## Overview

This involves two primary changes:
1. Online drift detectors are run on one instance at a time, with the time-step `self.t` updated each time. If a request contains a batch of instances, the payload must be unpacked and passed to the detector one instance at a time.
2. Online drift detectors have state, which must be periodically saved for checkpointing. State is saved every `state_save_freq` time-steps, where `state_save_freq` is a new configurable parameter added to `AlibiDetectSettings`.

## Limitations

Parallel inference is NOT currently supported for online detectors (i.e. `parallel_workers=0` is required). This would require careful thought wrt to how state is managed/shared across parallel workers, in addition to considering the implications of distributing non-iid samples between workers. 

## TODO's

- [x] Documentation? - _should be auto-completed pending https://github.com/SeldonIO/MLServer/pull/1109_
- [x] Linting

### DIFF
--- a/runtimes/alibi-detect/mlserver_alibi_detect/runtime.py
+++ b/runtimes/alibi-detect/mlserver_alibi_detect/runtime.py
@@ -144,7 +144,7 @@ class AlibiDetectRuntime(MLModel):
             )
 
         # Add headers
-        y["meta"]["headers"] = {  # TODO - if we want to viz the sliding windows, where should we store window size?
+        y["meta"]["headers"] = {
             "x-seldon-alibi-type": self.alibi_type,
             "x-seldon-alibi-method": self.alibi_method,
         }
--- a/runtimes/alibi-detect/mlserver_alibi_detect/runtime.py
+++ b/runtimes/alibi-detect/mlserver_alibi_detect/runtime.py
@@ -1,7 +1,8 @@
+import os
 import numpy as np
 
 from pydantic.error_wrappers import ValidationError
-from typing import Optional, List
+from typing import Optional, List, Union
 from pydantic import BaseSettings
 from functools import cached_property
 
@@ -41,6 +42,12 @@ class AlibiDetectSettings(BaseSettings):
     inference runs for all of them).
     """
 
+    state_save_freq: Optional[int] = 100  # TODO - need to check != 0
+    """
+    Save the detector state after every `state_save_freq` predictions.
+    Only applicable to detectors with a `save_state` method.
+    """
+
 
 class AlibiDetectRuntime(MLModel):
     """
@@ -58,15 +65,21 @@ class AlibiDetectRuntime(MLModel):
         super().__init__(settings)
 
     async def load(self) -> bool:
-        model_uri = await get_model_uri(self._settings)
+        self._model_uri = await get_model_uri(self._settings)
         try:
-            self._model = load_detector(model_uri)
+            self._model = load_detector(self._model_uri)
+            # Update AlibiDetectSettings according to whether an online drift detector (i.e. has a save_state method)
+            # TODO - in future we may use self._model.meta['online'] here, but must reconsider outlier detectors first
+            if hasattr(self._model, 'save_state'):
+                self._ad_settings.batch_size = None
+            else:
+                self._ad_settings.state_save_freq = None
         except (
-            ValueError,
-            FileNotFoundError,
-            EOFError,
-            NotImplementedError,
-            ValidationError,
+                ValueError,
+                FileNotFoundError,
+                EOFError,
+                NotImplementedError,
+                ValidationError,
         ) as e:
             raise MLServerError(
                 f"Invalid configuration for model {self._settings.name}: {e}"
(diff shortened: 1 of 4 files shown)


runtimes/alibi-detect/mlserver_alibi_detect/runtime.py
```diff
@@ -58,15 +65,21 @@ def __init__(self, settings: ModelSettings):
...
+        self._model_uri = await get_model_uri(self._settings)
         try:
-            self._model = load_detector(model_uri)
+            self._model = load_detector(self._model_uri)
+            # Update AlibiDetectSettings according to whether an online drift detector (i.e. has a save_state method)
+            # TODO - in future we may use self._model.meta['online'] here, but must reconsider outlier detectors first
+            if hasattr(self._model, 'save_state'):
+                self._ad_settings.batch_size = None
+            else:
+                self._ad_settings.state_save_freq = None
```
ascillitoe, line 76: The presence of a `save_state` method implies the detector is an online drift detector, meaning `detector.predict` takes one instance at a time (hence `batch_size = None)`, and we should save state (hence `state_save_freq` isn't set to `None`). Conversely, if the detector isn't an online drift detector, `state_save_freq=None` to prevent attempting to save state.

Note that checking `self._model.meta['online']` might be more elegant/readable. However, the meaning of this meta-field in Alibi Detect is currently confusing, since we set `online=True` for some outlier detectors such as `[Mahalanobis](https://docs.seldon.io/projects/alibi-detect/en/stable/od/methods/mahalanobis.html)`, despite this not having a `save_state` method and `predict` taking a batch of instances.

The above could need reconsidering in the future if we add `save_state` methods to new detectors, but we don't have immediate plans for this...

runtimes/alibi-detect/mlserver_alibi_detect/runtime.py
```diff
@@ -105,23 +118,37 @@ def _detect(self, payload: InferenceRequest) -> InferenceResponse:
...
-            y = self._model.predict(np.array(input_data), **predict_kwargs)
-            return self._encode_response(y)
-        except (ValueError, IndexError) as e:
-            raise InferenceError(
-                f"Invalid predict parameters for model {self._settings.name}: {e}"
-            ) from e
+        # If batch is configured, wrap X in a list so that it is not unpacked
+        X = np.array(input_data)
+        if self._ad_settings.batch_size:
+            X = [X]
```
ascillitoe, line 124: Wrap `X` in a list (of length 1) so that we don't unpack into single instances when running in batch mode (i.e. offline detectors). `pred` is then returned as a list of length 1. 

This slightly wacky logic is intended to avoid having two duplicate try-except blocks for the two (offline vs online) use cases...

runtimes/alibi-detect/mlserver_alibi_detect/runtime.py
```diff
@@ -105,23 +118,37 @@ def _detect(self, payload: InferenceRequest) -> InferenceResponse:
...
+            try:
+                pred.append(self._model.predict(x, **predict_kwargs))
+            except (ValueError, IndexError) as e:
+                raise InferenceError(
+                    f"Invalid predict parameters for model {self._settings.name}: {e}"
+                ) from e
+            # Save state if necessary
+            if self._ad_settings.state_save_freq and \
+                    self._model.t % self._ad_settings.state_save_freq == 0 and self._model.t > 0:
+                self._model.save_state(os.path.join(self._model_uri, 'state'))
```
ascillitoe, line 139: Could nest the saving logic in a `AlibiDetectRuntime._save_state` method if we want to raise custom errors etc

runtimes/alibi-detect/mlserver_alibi_detect/runtime.py
```diff
@@ -58,9 +65,11 @@ def __init__(self, settings: ModelSettings):
         super().__init__(settings)
 
     async def load(self) -> bool:
-        model_uri = await get_model_uri(self._settings)
+        self._model_uri = await get_model_uri(self._settings)
         try:
-            self._model = load_detector(model_uri)
+            self._model = load_detector(self._model_uri)
+            # Check whether an online drift detector (i.e. has a save_state method)
+            self._online = True if hasattr(self._model, 'save_state') else False
```
ascillitoe, line 72: The presence of a `save_state` method implies the detector is an online drift detector, meaning `detector.predict` takes one instance at a time **and** we should save state. Conversely, if the detector isn't an online drift detector, `state_save_freq=None` to prevent attempting to save state.

Note that checking `self._model.meta['online']` might be more elegant/readable. However, the meaning of this meta-field in Alibi Detect is currently confusing, since we set `online=True` for some outlier detectors such as `[Mahalanobis](https://docs.seldon.io/projects/alibi-detect/en/stable/od/methods/mahalanobis.html)`, despite this not having a `save_state` method and `predict` taking a batch of instances.

The above could need reconsidering in the future if we add `save_state` methods to new detectors, but we don't have immediate plans for this...

runtimes/alibi-detect/tests/conftest.py
```diff
@@ -125,6 +127,21 @@ def drift_detector_settings(
...
+@pytest.fixture
+def online_drift_detector_settings(
+    online_drift_detector_uri: str,
+) -> ModelSettings:
+    return ModelSettings(
+        name="alibi-detect-model",
+        implementation=AlibiDetectRuntime,
+        parameters=ModelParameters(
+            uri=online_drift_detector_uri,
+            version="v1.2.3",  # TODO - should this match version.py?
```
ascillitoe, line 139: Question: what version should this be?

runtimes/alibi-detect/mlserver_alibi_detect/runtime.py
```diff
@@ -105,23 +114,37 @@ def _detect(self, payload: InferenceRequest) -> InferenceResponse:
...
         outputs = []
         for key in y["data"]:
             outputs.append(
-                NumpyCodec.encode_output(name=key, payload=np.array([y["data"][key]]))
+                NumpyCodec.encode_output(name=key, payload=np.array(y["data"][key]))
             )
 
         # Add headers
-        y["meta"]["headers"] = {
+        y["meta"]["headers"] = {  # TODO - if we want to viz the sliding windows, where should we store window size?
```
ascillitoe, line 147: Placeholder: will discuss offline

runtimes/alibi-detect/mlserver_alibi_detect/runtime.py
adriangonz: Yuup! That would make sense - we can also re-use that method in the (future) `unload()` handler. We could also move the `if` check to a `_should_save_state()` method to keep the logic cleaner.

runtimes/alibi-detect/tests/conftest.py
adriangonz: This version is arbitrary (it can be anything the user wants). We could even set it to `None`.

adriangonz (approved):
This looks great @ascillitoe! :rocket: 

I've added a minor comment around the state save logic and check, but everything else should be good to go once the linter and tests pass!

runtimes/alibi-detect/mlserver_alibi_detect/runtime.py
ascillitoe: Refactored in https://github.com/SeldonIO/MLServer/pull/1108/commits/4790d5f8c49aa1d49d993f5ff92bb2d0740b8435. Note that I haven't added  unit test to check that `_save_state` raises an error, since it is difficult to cause an error! Alibi Detect automatically creates a directory, so a `FileNotFoundError` isn't possible. I was thinking about forcing a `PermissionError` but it doesn't look easy to change the permissions of the pytest `tmp_path`...

PR #1120, opened 2023-04-28
Add default LD_LIBRARY_PATH env var

adriangonz:
Set a default value for the `LD_LIBRARY_PATH` env var to ensure the CUDA libraries injected at run-time by GKE can be found.


adriangonz:
Issue in tests should be fixed by #1121

PR #1143, opened 2023-05-11
Added option to configure the grpc histogram

cristiancl25:


mlserver/settings.py
```diff
@@ -190,6 +191,9 @@ class Config:
     grpc_max_message_length: Optional[int] = None
     """Maximum length (i.e. size) of gRPC payloads."""
 
+    grpc_response_latency_histogram : bool = False
```
adriangonz, line 194: Would there be any caveat to removing this setting and always keeping it enabled? It seems reasonable (and aligns it with its REST counterpart - where I think the latency histogram is also always enabled).

adriangonz (approved):
Thanks for the changes @cristiancl25 !

PR looks great now. Once tests are green, we can go ahead with this one. :rocket:

cristiancl25:
Done!

adriangonz:
Thanks for rebasing it! It should be ready to go now  :+1:

PR #1152, opened 2023-05-16
Unreachable line

saeid93:
The last return line of the `_wait_response` method seems to be unreachable (?)


adriangonz (approved):
Good spot! :+1:  Probably a remnant of a previous version

PR #1158, opened 2023-05-23
Move to Poetry to manage deps

### LINKED ISSUE #139: Add lockfile to MLServer

adriangonz:
Fixes #139

## Changelog

- Move to poetry to manage deps
- Move Tox config from `setup.cfg` to individual `tox.ini` files (a global one, and one per runtime)


RafalSkolasinski:
Worth to link the related issue https://github.com/SeldonIO/MLServer/issues/139

PR #1166, opened 2023-05-23
Replace pip freeze for mlserver + tensorflow deps

adriangonz:


edshee (approved):
Looks good. thanks for fixing

PR #1174, opened 2023-05-25
Fix small typo in Python API docs

krishanbhasin-gc:
`a python packaged` -> `a python package`


adriangonz (approved):
Good spot! Thanks a lot for the typo fix @krishanbhasin-gc!

PR #1182, opened 2023-05-30
Only run tests on modified runtimes

adriangonz:
After adding support for MacOS in the tests, the test matrix has grown too large - which means tests are taking a while to finish. This PR ensures that we only execute tests for modified runtimes to try and make the PR tests a bit lighter (in `master`, it should still test all of them).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -68,18 +68,21 @@ jobs:
       - name: Check Runtime Changes
         id: changes
         run: |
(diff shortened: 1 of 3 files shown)


adriangonz:
Even after disabling runtime tests, macos tests still take a while (~30m to finish, compared to ~15m from the Linux ones) so we will disable them and will only run them on `master` (i.e. similar to the `all-runtimes` tests).

PR #1184, opened 2023-05-31
Add workaround for segfault in MacOS tests

adriangonz:
Recent versions of OpenMP (i.e. >= 12) cause segfaults in MacOS when training LightGBM / XGBoost models (but only when Torch is present). This PR downgrades OpenMP to an older version (i.e. == 11). 

See https://github.com/microsoft/LightGBM/issues/4229 and https://github.com/autogluon/autogluon/issues/1442 for more details.


adriangonz:
The only failure in `all-runtimes` is due to a flaky test in the MacOS setup - not related to this PR.

Issue #1210, opened 2023-06-12
got unexpected datatype None from numpy array, expected BYTES" from mlserver infer


RafalSkolasinski:
`mlserver infer ...` subcommand does not support BYTES inputs - reproduced by mock v2 echo model [here](https://github.com/RafalSkolasinski/seldon-models/tree/master/v2-echo-model).

I used a simple Echo Model for testing 

```python
import json

from typing import Dict, Any
from mlserver import MLModel, types
from mlserver.codecs import StringCodec


class EchoModel(MLModel):
    async def load(self) -> bool:
        print("Echo Model Initialized")
        return await super().load()

    async def predict(self, payload: types.InferenceRequest) -> types.InferenceResponse:
        return types.InferenceResponse(
            id=payload.id,
            model_name=self.name,
            model_version=self.version,
            outputs=[
                types.ResponseOutput(
                    name=input.name,
                    shape=input.shape,
                    datatype=input.datatype,
                    data=input.data,
                    parameters=input.parameters,
                )
                for input in payload.inputs
            ],
        )
```
with `settings.json`
```json
{
    "debug": "true"
}
```
and `model-settings.json`
```json
{
    "name": "echo-model",
    "implementation": "model.EchoModel"
}
```

the model served locally with `mlserver start` does properly process request of form
```json
{
  "inputs": [
    {
      "name": "fp32-input",
      "shape": [1, 1],
      "datatype": "FP32",
      "data":  [ 1.23 ]
    },
    {
      "name": "int32-input",
      "shape": [1, 1],
      "datatype": "INT32",
      "data":  [ 1.23 ]
    },
    {
      "name": "bytes-input",
      "shape": [1, 1],
      "datatype": "BYTES",
      "data":  [ "some bytes" ]
    }
  ],
  "outputs": [
    {
      "name": "predict"
    }
  ]
}
```

however when trying to put it as input batch with
```
{"inputs":[{"name":"fp32-input","shape":[1,1],"datatype":"FP32","data":[1.23]},{"name":"int32-input","shape":[1,1],"datatype":"INT32","data":[1.23]},{"name":"bytes-input","shape":[1,1],"datatype":"BYTES","data":["some bytes"]}],"outputs":[{"name":"predict"}]}
[...]

PR #1213, opened 2023-06-12
fix mlserver infer with BYTES

RafalSkolasinski:
Closes https://github.com/SeldonIO/MLServer/issues/1210

### DIFF
--- a/mlserver/batch_processing.py
+++ b/mlserver/batch_processing.py
@@ -5,6 +5,7 @@ from random import random
 import tritonclient.http.aio as httpclient
 
 import asyncio
+import numpy as np
 import aiofiles
 import logging
 import click
@@ -101,17 +102,21 @@ class TritonRequest:
         cls, inference_request: InferenceRequest, binary_data: bool
     ) -> "TritonRequest":
         inputs = []
-
         for request_input in inference_request.inputs or []:
             new_input = httpclient.InferInput(
                 request_input.name, request_input.shape, request_input.datatype
             )
+            request_input_np = NumpyCodec.decode_input(request_input)
+
+            # Change datatype if BYTES to satisfy Tritonclient checks
+            if request_input.datatype == "BYTES":
+                request_input_np = request_input_np.astype(np.object_)
+
             new_input.set_data_from_numpy(
-                NumpyCodec.decode_input(request_input),
+                request_input_np,
                 binary_data=binary_data,
             )
             inputs.append(new_input)
-
         outputs = []
         for request_output in inference_request.outputs or []:
             new_output = httpclient.InferRequestedOutput(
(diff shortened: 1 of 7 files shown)


mlserver/batch_processing.py
```diff
@@ -101,17 +102,21 @@ def from_inference_request(
...
     ) -> "TritonRequest":
         inputs = []
-
         for request_input in inference_request.inputs or []:
             new_input = httpclient.InferInput(
                 request_input.name, request_input.shape, request_input.datatype
             )
+            request_input_np = NumpyCodec.decode_input(request_input)
+
+            # Change datatype if BYTES to satisfy Tritonclient checks
```
RafalSkolasinski, line 111: I have the feeling this should be something handled in the `NumpyCodec` but it may be too `tritonclient` specific...

tests/conftest.py
```diff
@@ -294,6 +301,30 @@ async def mlserver(
     assert not prom_files
 
 
+@pytest.fixture
+async def echo_mlserver(
```
adriangonz, line 305: Instead of creating a new one, I think we could just load the echo model on a `batch_processing/conftest.py` fixture? Or even just in the new test. That way we avoid having to maintain multiple similar fixtures.

mlserver/batch_processing.py
adriangonz: Good point - do you know what's returned by `NumpyCodec.decode_input` at the moment?

adriangonz (commented):
Nice one @RafalSkolasinski! Added a small comment around tests - I think we can just reuse one of the existing fixtures, but everything else looks great. :+1:

tests/conftest.py
RafalSkolasinski: I may not read it correct, but I understood that `mlserver` fixture was having hard-coded `sum-model` to be used. I may have missed the way to customize it though?

RafalSkolasinski: Fixed

mlserver/batch_processing.py
adriangonz: I think what's going is that the NumpyCodec encodes that list as a tensor of strings (i.e. using `np.dtype(str)`). However, Triton doesn't know how to handle that case. 

https://github.com/SeldonIO/MLServer/blob/d283fc0c6f07f731ed94139847da9e70d9be2497/mlserver/codecs/numpy.py#L41-L43

I can't remember why we had to add that `np.dtype(str)` special case, but I would be keen to keep it as is to avoid any potential side effects. Having said that, it's unclear whether changing the `dtype` here may cause any issues downstream (although I'd expect that strings in np arrays are quite an edge case).

RafalSkolasinski:
@adriangonz done - should be fine now

PR #1230, opened 2023-06-14
Silent logging

dtpryce:
- Removed the v option from tar
- Sent output of activate to dev/null 

Was hard to test with the mlserver image being built elsewhere


hack/activate-env.sh
```diff
@@ -36,7 +36,7 @@ _activateEnv() {
   echo "--> Sourcing new environment at $_envFolder..."
   # Need to disable unbound errors for activate
   set +u
-  source $_activate
+  source $_activate &> /dev/null 2>&1
```
adriangonz, line 39: I would keep the error output visible - in case it fails to activate the environment somehow

adriangonz (commented):
Nice one @dtpryce! Added a small comment there, but the approach make sense :+1:

hack/activate-env.sh
dtpryce: Updated to be only `1> /dev/null` so that should silence all non-error logging

adriangonz (approved):

dtpryce:
Awesome @adriangonz feel free to merge into 1.4.0 or earlier :P

dtpryce:
> Awesome @adriangonz feel free to merge into 1.4.0 or earlier :P

Actually wait I have another small update for it

adriangonz:
Nice one! This should be ready to go. Thanks for the changes @dtpryce ! :rocket:

PR #1234, opened 2023-06-16
Restart dead workers

### LINKED ISSUE #707: Server appears healthy even when all workers are down
I experienced an issue, when all parallel workers crash for foreseeable reasons, mlserver continues to run as if it was healthy but it is no longer able to process any (rest api) requests.
After looking a bit into the code base it seems that `_get_worker()` in `dispatcher.py` does not check if the worker is still alive before sending the request to it.
It would be helpful if mlserver could catch and report crashed workers.

adriangonz:
Fixes #707 

## Changelog

- Detect when a worker stopped unexpectedly through `SIGCLD`
- Cancel all in-flight requests (added new `AsyncResponses` class to keep track of in-flight requests)
- Load all models on new worker (added new `WorkerRegistry` class to keep track of models per worker)

### DIFF
--- a/mlserver/parallel/dispatcher.py
+++ b/mlserver/parallel/dispatcher.py
@@ -121,18 +121,46 @@ class Dispatcher:
     def __init__(self, workers: Dict[int, Worker], responses: Queue):
         self._responses = responses
         self._workers = workers
-        self._workers_round_robin = cycle(self._workers.keys())
+        self._workers_round_robin = self._reset_round_robin()
+        self._worker_starting_lock = asyncio.Lock()
         self._active = False
         self._process_responses_task = None
         self._executor = ThreadPoolExecutor()
         self._async_responses = AsyncResponses()
 
+    def _reset_round_robin(self) -> list:
+        worker_pids = list(self._workers.keys())
+        self._workers_round_robin = cycle(worker_pids)
+        return self._workers_round_robin
+
+    async def on_worker_start(self, worker: Worker):
+        """
+        Handler for workers who have just started but are still not ready to
+        receive traffic.
+        This is used for workers that got restarted and need to reload all
+        models.
+        """
+        # Lock while worker is coming up to ensure no model updates get lost in
+        # translation
+        async with self._worker_starting_lock:
+            self._workers[worker.pid] = worker
+
+    def on_worker_ready(self, worker: Worker):
+        """
+        Handler for workers who are now ready to receive traffic.
+        """
+        self._reset_round_robin()
+
     def on_worker_stop(self, worker: Worker, exit_code: int):
+        """
+        Handler used for workers who stopped unexpectedly and there need to be
+        removed from the round robin rotation.
+        """
         pid = worker.pid
         if pid in self._workers:
             del self._workers[pid]
 
-        self._workers_round_robin = cycle(self._workers.keys())
+        self._reset_round_robin()
         self._async_responses.cancel(worker, exit_code)
 
     def start(self):
@@ -185,14 +213,15 @@ class Dispatcher:
     async def dispatch_update(
         self, model_update: ModelUpdateMessage
     ) -> List[ModelResponseMessage]:
-        return await asyncio.gather(
-            *[
-                self._dispatch_update(worker, model_update)
-                for worker in self._workers.values()
-            ]
-        )
+        async with self._worker_starting_lock:
+            return await asyncio.gather(
+                *[
+                    self.dispatch_update_to_worker(worker, model_update)
+                    for worker in self._workers.values()
+                ]
+            )
 
-    async def _dispatch_update(
+    async def dispatch_update_to_worker(
         self, worker: Worker, model_update: ModelUpdateMessage
     ) -> ModelResponseMessage:
         # NOTE: Need to rewrite the UUID to ensure each worker sends back a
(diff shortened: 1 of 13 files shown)


mlserver/parallel/registry.py
```diff
@@ -56,6 +57,24 @@ def __init__(
...
+            lambda *args: asyncio.create_task(self._handle_worker_stop(*args)),
+        )
+
+    async def _handle_worker_stop(self, signum, frame):
+        pid, exit_code = os.waitpid(-1, os.WNOHANG)
+        if pid == 0 or exit_code == 0:
+            # Worker terminated gracefully, nothing to do
+            return
+
+        await self._default_pool.on_worker_stop(pid, exit_code)
```
sakoush, line 73: question: whats the difference between default_pool and other pools?

mlserver/parallel/pool.py
```diff
@@ -23,6 +24,35 @@
 InferencePoolHook = Callable[[Worker], Awaitable[None]]
 
 
+class WorkerRegistry:
```
sakoush, line 27: nit: should this be in `registry.py`?

mlserver/parallel/pool.py
```diff
@@ -66,19 +93,80 @@ def env_hash(self) -> Optional[str]:
...
+        self._dispatcher.on_worker_stop(worker, exit_code)
+        if pid in self._workers:
+            # NOTE: worker may be removed by dispatcher
+            del self._workers[pid]
+
+        # Call attached on_worker_stop hooks
+        await asyncio.gather(*[callback(worker) for callback in self._on_worker_stop])
+
+        # Start a new worker
+        await self._start_worker()
```
sakoush, line 123: is there any race condition when we start a new worker and any new model load calls?

tests/parallel/test_pool.py
```diff
@@ -31,6 +34,78 @@ def test_workers_start(inference_pool: InferencePool, settings: Settings):
...
+async def test_start_worker_new_model(
+    settings: Settings,
+    inference_pool: InferencePool,
+    sum_model: MLModel,
+    simple_model: MLModel,
+):
+    # Ensure pool has some loaded models
+    await inference_pool.load_model(sum_model)
+
+    # Assert new models make their way throug to the new worker
```
sakoush, line 88:
```suggestion
    # Assert new models make their way through to the new worker
```

sakoush (approved):
Great PR! I left some minor comments.

mlserver/parallel/dispatcher.py
```diff
@@ -44,6 +48,128 @@ def _get_or_create_metric(self) -> Histogram:
...
+        Schedule a response to be resolved in the future.
+        """
+        loop = asyncio.get_running_loop()
+        future = loop.create_future()
+        message_id = message.id
+        self._futures[message_id] = future
+
+        # Keep track of allocation for in-flight requests
+        self._futures_map[message_id] = worker.pid  # type: ignore
+        self._workers_map[worker.pid].add(message_id)  # type: ignore
```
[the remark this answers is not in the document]
adriangonz: Good point - let me do that

mlserver/parallel/dispatcher.py
```diff
@@ -44,6 +48,128 @@ def _get_or_create_metric(self) -> Histogram:
...
+        self, message: Message, worker: Worker
+    ) -> ModelResponseMessage:
+        """
+        Schedule a response and wait until it gets resolved.
+        """
+        message_id = message.id
+        self.schedule(message, worker)
+        return await self.wait(message_id)
+
+    def schedule(self, message: Message, worker: Worker) -> Future:
```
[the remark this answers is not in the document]
adriangonz: Yuuup, we can make this and `wait()` private

mlserver/parallel/dispatcher.py
```diff
@@ -44,6 +48,128 @@ def _get_or_create_metric(self) -> Histogram:
...
+        in_flight = self._workers_map[worker.pid]  # type: ignore
+        if in_flight:
+            logger.info(
+                f"Cancelling {len(in_flight)} in-flight requests for "
+                "worker {worker.pid} which died unexpectedly with "
+                "exit code {exit_code}..."
+            )
+        for message_id in in_flight:
+            # TODO: Can we use worker.exitcode here?
+            err = WorkerStop(exit_code)
```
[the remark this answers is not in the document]
adriangonz: Probably it can, but decided to play it safe in case we run into issues by re-using the same exception object across coroutines

mlserver/parallel/registry.py
adriangonz: The default pool uses the same environment as the main MLServer process and it's always running. When a model doesn't specify a custom environment, it will just go to the default pool.

mlserver/parallel/pool.py
adriangonz: I'd say it's a different type of registry. The registry in `registry.py` keeps track of the different worker pools running, whereas this registry just keeps track of which models have each worker loaded.

mlserver/parallel/dispatcher.py
```diff
@@ -44,6 +48,128 @@ def _get_or_create_metric(self) -> Histogram:
...
+            future = self._futures[message_id]
+            loop = future.get_loop()
+            loop.call_soon_threadsafe(future.set_exception, err)
```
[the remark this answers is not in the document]
adriangonz: Opened a separate issue to track this: #1266

PR #1263, opened 2023-06-26
Publish is_drift metric to Prom

joshsgoldstein:
added drift metrics to be scraped by prometheus

### DIFF
--- a/runtimes/alibi-detect/mlserver_alibi_detect/runtime.py
+++ b/runtimes/alibi-detect/mlserver_alibi_detect/runtime.py
@@ -130,6 +130,7 @@ class AlibiDetectRuntime(MLModel):
             try:
                 current_pred = self._model.predict(x, **predict_kwargs)
                 pred.append(current_pred)
+                print("in the current prediction")
                 print(current_pred['data']['is_drift'])
                 mlserver.log(seldon_model_drift=current_pred['data']['is_drift'])
             except (ValueError, IndexError) as e:
--- a/runtimes/alibi-detect/mlserver_alibi_detect/runtime.py
+++ b/runtimes/alibi-detect/mlserver_alibi_detect/runtime.py
@@ -8,6 +8,7 @@ from functools import cached_property
 
 from alibi_detect.saving import load_detector
 
+import mlserver
 from mlserver.batching import BatchedRequests
 from mlserver.types import InferenceRequest, InferenceResponse
 from mlserver.settings import ModelSettings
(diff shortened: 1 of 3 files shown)


runtimes/alibi-detect/mlserver_alibi_detect/runtime.py
```diff
@@ -136,6 +141,18 @@ def _detect(self, payload: InferenceRequest) -> InferenceResponse:
 
         return self._encode_response(self._postproc_pred(pred))
 
+    def _log_metrics(self, current_pred: dict) -> None:
+        if "data" not in current_pred:
+            return
+
+        data = current_pred["data"]
+        if "is_drift" in data:
+            # NOTE: is_drift will hold an array which could be larger than 1
```
ascillitoe, line 150: To give context here, for univarite detectors (like `KSDrift`) `is_drift` will be an array like `'is_drift': array([1, 0])` if `predict(..., drift_type='feature')`, otherwise an int if `predict(..., drift_type='batch')`. Whereas for multivariate detectors like `MMDDrift` it will always be an int...

So above looks good, but just noting that currently the tests only test the `drift_type='feature'` behaviour...

runtimes/alibi-detect/mlserver_alibi_detect/runtime.py
```diff
@@ -136,6 +141,18 @@ def _detect(self, payload: InferenceRequest) -> InferenceResponse:
 
         return self._encode_response(self._postproc_pred(pred))
 
+    def _log_metrics(self, current_pred: dict) -> None:
+        if "data" not in current_pred:
+            return
+
+        data = current_pred["data"]
+        if "is_drift" in data:
```
ascillitoe, line 149: Question for @adriangonz and @joshsgoldstein. What is the motivation for this PR? 

Asking as wondering if we are definitely only concerned about `is_drift` here? Often `p_val` and `distance` are of equal (or greater) interest. As is `threshold`, and `distance_threshold` (dependent on detector).

ascillitoe (commented):
Two initial questions before final review.

runtimes/alibi-detect/mlserver_alibi_detect/runtime.py
joshsgoldstein: @ascillitoe we are enabling scv2 alerting in deploy

ascillitoe (approved):
LGTM!

PR #1268, opened 2023-06-30
Rename HF codec to `hf`

adriangonz:
## Changelog

- Rename multi-input HF codec's content type to `hf` (the previous `str` caused conflicts with MLServer's base one)
- Always use `HuggingfaceRequestCodec` in HF runtime (i.e. ignore top-level request `content_type` annotation)

### DIFF
--- a/runtimes/huggingface/README.md
+++ b/runtimes/huggingface/README.md
@@ -13,6 +13,16 @@ pip install mlserver mlserver-huggingface
 For further information on how to use MLServer with HuggingFace, you can check
 out this [worked out example](../../docs/examples/huggingface/README.md).
 
+## Content Types
+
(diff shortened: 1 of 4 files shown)


cliveseldon:
I don't think this fixes thins as for all  cases:

The issue is this seldon core v2 huggingface demo explainer is calling a pipeline so we can't get the metadata, so the blackbox explainer adds content-type='str' by default which no longer works as we need content_type="hf".

if content_type is provided in payload it always has precedene over any in model-settings.json.

adriangonz:
As far as I can see, that metadata only applies to the input-level metadata but leaves the request-level metadata unset - so we can leave it as it is.

cliveseldon (approved):

PR #1279, opened 2023-07-03
Add support for white-box explainers to alibi-explain runtime

ascillitoe:
## Summary

This PR adds support for the sklearn API white-box explainers (`TreeShap`, `TreePartialDependence`, `PartialDependenceVariance`) to the `AlibiExplainRuntime`. 

## Implementation details

- **New runtime:** Support is implemented via the new `AlibiExplainSKLearnAPIRuntime` subclass. For the remaining white-box explainers two further subclassed runtimes will be required (one being the `IntegratedGradientsWrapper`). See [the notion doc](https://www.notion.so/Scoping-of-enhanced-MLServer-explainer-support-24261abcdee143f083a77a9a2d95f143?pvs=25) for more details.

- **Tests:** Some explainer's have various "nitty-gritty" details wrt to serving, for example the `target` kwarg for `IntegratedGradients`, calling `fit` with no data for `TreeShap`, and black-box/white-box mode of `PartialDependenceVariance` being determined by `type(predictor)`. Therefore, the PR is proposing to perform an E2E test (`test_end_2_end`) on every explainer (with relevant parametrisation such as testing uri vs init_param's based loading). This would provide more confidence that all explainers can actually be served. However, it will increase the number of tests being run somewhat, so needs discussion...

## TODO

- [x] Linting etc
- [ ] ~~Docs (leave until all white box explainers supported...)~~
- [x] Add tests to check loading of all supported tree models e.g. lightgbm models etc - _done in [afae1e6](https://github.com/SeldonIO/MLServer/pull/1279/commits/afae1e6fa7e951487b4b39820d306c6126e5317c)_


runtimes/alibi-explain/mlserver_alibi_explain/explainers/sklearn_api_runtime.py
```diff
@@ -0,0 +1,36 @@
...
+class AlibiExplainSKLearnAPIRuntime(AlibiExplainWhiteBoxRuntime):
+    """
+    Runtime for white-box explainers that require access to a tree-based model matching the SKLearn API, such as
+    a sklearn, XGBoost, or LightGBM model. Example explainers include TreeShap and TreePartialDependence.
+    """
+    async def _get_inference_model(self) -> Any:
+        inference_model_path = self.alibi_explain_settings.infer_uri
+        # Attempt to load model in order: XGBoost, LightGBM, sklearn
+        # TODO - add support for CatBoost (would require model_type = 'classifier' or 'regressor' in settings)
+        try:
```
ascillitoe, line 22: This is pretty ugly, but can't see an obvious alternative unless we add `model_flavour` and `model_type` options to `AlibiExplainSettings`?

runtimes/alibi-explain/mlserver_alibi_explain/explainers/white_box_runtime.py
```diff
@@ -29,17 +29,23 @@ def __init__(self, settings: ModelSettings, explainer_class: Type[Explainer]):
...
+        # white box explainers requires access to the full inference model
         self._inference_model = await self._get_inference_model()
 
         if self.alibi_explain_settings.init_parameters is not None:
+            # Instantiate explainer with init parameters (and give it full inference model)
             init_parameters = self.alibi_explain_settings.init_parameters
-            # white box explainers requires access to the inference model
-            init_parameters["model"] = self._inference_model
-            self._model = self._explainer_class(**init_parameters)  # type: ignore
+            self._model = self._explainer_class(self._inference_model, **init_parameters)  # type: ignore
```
ascillitoe, line 38: Passing `self._inference_model` as unnamed arg instead of kwarg, because some explainers name this arg differently (i.e. `predict_fn`, `model` or `predictor`).

runtimes/alibi-explain/mlserver_alibi_explain/explainers/white_box_runtime.py
```diff
@@ -29,17 +29,23 @@ def __init__(self, settings: ModelSettings, explainer_class: Type[Explainer]):
...
         else:
+            # Load explainer from URI (and give it full inference model)
             self._model = await self._load_from_uri(self._inference_model)
 
         return True
 
+    def _explain_impl(self, input_data: Any, explain_parameters: Dict) -> Explanation:
```
ascillitoe, line 45: Moving to base `AlibiExplainWhiteBoxRuntime` class because two out of three subclasses won't need specialised `_explain_impl` (only integrated gradients one needs it).

runtimes/alibi-explain/pyproject.toml
```diff
@@ -10,6 +10,9 @@ packages = [{include = "mlserver_alibi_explain"}]
 [tool.poetry.dependencies]
 python = "^3.8.1,<3.12"
 mlserver = "*"
+mlserver_sklearn = "*"
```
ascillitoe, line 13: Adding `sklearn`, `xgboost` and `lightgbm` as deps via mlserver runtimes only because `_load_sklearn_interface` from `mlserver_xgboost` is being used inside `_get_inference_model`. Might need more discussion...

runtimes/alibi-explain/mlserver_alibi_explain/explainers/black_box_runtime.py
```diff
@@ -44,8 +44,7 @@ async def load(self) -> bool:
         # TODO: use init explainer field instead?
         if self.alibi_explain_settings.init_parameters is not None:
             init_parameters = self.alibi_explain_settings.init_parameters
-            init_parameters["predictor"] = self._infer_impl
```
ascillitoe, line 47: See https://github.com/SeldonIO/MLServer/pull/1279/files#r1250878077

runtimes/alibi-explain/tests/test_white_box_sklearn_cases.py
```diff
@@ -0,0 +1,66 @@
...
+    # Set kwargs
+    init_kwargs = {
+        'predictor': sk_income_model,
+        'feature_names': income_data['feature_names'],
+        'categorical_names': income_data['category_map'],
+        'target_names': income_data['target_names'],
+    }
+    explain_kwargs = {'features': [0, 2]}
+
+    save_dir = tmp_path if save else None
```
ascillitoe, line 44: Bit ugly, but can't seem to parametrize `save_dir` with the `tmp_path` fixture directly...

runtimes/alibi-explain/mlserver_alibi_explain/alibi_dependency_reference.py
```diff
@@ -57,6 +61,21 @@ class ExplainerDependencyReference:
...
+        explainer_name=_TREE_SHAP_TAG,
+        runtime_class=f"{_WHITEBOX_SKLEARN_MODULE}.AlibiExplainSKLearnAPIRuntime",
+        alibi_class="alibi.explainers.TreeShap",
+    ),
+    _TREE_PARTIAL_DEPENDENCE_TAG: ExplainerDependencyReference(
+        explainer_name=_TREE_PARTIAL_DEPENDENCE_TAG,
+        runtime_class=f"{_WHITEBOX_SKLEARN_MODULE}.AlibiExplainSKLearnAPIRuntime",
+        alibi_class="alibi.explainers.TreePartialDependence",
+    ),
+    _TREE_PARTIAL_DEPENDENCE_VARIANCE_TAG: ExplainerDependencyReference(
```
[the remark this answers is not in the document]
adriangonz: Sounds good!

runtimes/alibi-explain/mlserver_alibi_explain/explainers/black_box_runtime.py
adriangonz: Nice one! Good spot :+1:

runtimes/alibi-explain/mlserver_alibi_explain/explainers/sklearn_api_runtime.py
adriangonz: haha I agree... when you see it written out is a bit weird. 

Happy to add a `model_type` field to the settings? 

A cleaner alternative IMO would be to expect models to provide their own `model-settings.json`. We could then just load the models as usual with their relevant runtimes (and then adding some accessor to retrieve the actual model instance). That way, framework-specific stuff would just fall onto each's runtime implementation. However, assuming that the model's `model-settings.json` will always be present may be too big of an assumption?

runtimes/alibi-explain/mlserver_alibi_explain/explainers/white_box_runtime.py
adriangonz: Nice one on the `Explanation` object - I hadn't realised that was now available in Alibi!

runtimes/alibi-explain/pyproject.toml
adriangonz: I think this makes sense - it also opens the door to loading the underlying model's `model-settings.json` file if that's something we want. 

I would also add the local path as a dev dep to ensure they use the local version during tests (i.e. similar to what we do for `mlserver` here: https://github.com/SeldonIO/MLServer/pull/1279/files#diff-004b0ec921eca570278c30340365c92c93b4db77b758655e0d67acaa69f99371R20)

runtimes/alibi-explain/tests/helpers/sk_model.py
```diff
@@ -0,0 +1,66 @@
...
+
+from sklearn.ensemble import GradientBoostingClassifier
+from alibi.datasets import fetch_adult
+
+from mlserver import MLModel
+from mlserver.codecs import NumpyCodec
+from mlserver.types import InferenceRequest, InferenceResponse
+
+
+_MODEL_PATH = Path(os.path.dirname(__file__)).parent / ".data" / "sk_income" / "model.joblib"
```
adriangonz, line 13: If the model is quick to train, there's no need for caching. I think we did initially that for the pytorch model bc it took a while, but generally the approach is to train on every round if it's quick enough to simplify the fixtures.

runtimes/alibi-explain/tests/helpers/sk_model.py
```diff
@@ -0,0 +1,66 @@
...
+        input_data = np_codec.decode_input(model_input)
+        output_data = self._model(input_data)
+        return InferenceResponse(
+            model_name=self.name,
+            outputs=[np_codec.encode_output("predict", output_data)]
+        )
+
+    async def load(self) -> bool:
+        self._model = joblib.load(get_sk_income_model_uri())
+        return True
```
adriangonz, line 35: Do we use this one?

runtimes/alibi-explain/tests/test_white_box_sklearn.py
```diff
@@ -0,0 +1,59 @@
...
+    and explain kwargs. The pytest-cases doesn't work too well yet with AsyncIO, therefore we need to treat the
+    fixture as an Awaitable and await it in the tests (See https://github.com/smarie/python-pytest-cases/issues/286).
+    The `explainer`, `request` and `explain_kwargs` are simply passed through.
+    """
+    runtime = AlibiExplainRuntime(model_settings)
+    runtime.ready = await runtime.load()
+    return runtime, explainer, request, explain_kwargs
+
+
+async def test_end_2_end(test_case):
```
adriangonz, line 37: I would just call it `test_explain` TBH, as that seems more clear (i.e. it's just testing the `.explain()` method for whitebox sklearn explainers, and the scaffolding just trains the explainer, the model, etc.

runtimes/alibi-explain/tests/test_white_box_sklearn_cases.py
adriangonz: It would probably be a bit fiddly to do that - so this workaround is good enough! :+1:

adriangonz (commented):
This is a great PR @ascillitoe !!

I love the test thoroughness - the more tests the better IMO (as long as they are not massively slow!). Added a few comments here-and-there, but they are mainly questions and suggestions - nothing biggie.

runtimes/alibi-explain/mlserver_alibi_explain/explainers/sklearn_api_runtime.py
```diff
@@ -0,0 +1,36 @@
...
+import lightgbm as lgb
+from xgboost.core import XGBoostError
+from lightgbm.basic import LightGBMError
+
+from mlserver_xgboost.xgboost import _load_sklearn_interface as load_xgb_model
+from mlserver.errors import InvalidModelURI
+from mlserver_alibi_explain.explainers.white_box_runtime import AlibiExplainWhiteBoxRuntime
+
+
+class AlibiExplainSKLearnAPIRuntime(AlibiExplainWhiteBoxRuntime):
```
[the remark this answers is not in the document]
ascillitoe: Renamed to `SKLearnRuntime" in ed0a1b9f880aabf1069e16e705e249bbf8bab0fd.

runtimes/alibi-explain/mlserver_alibi_explain/explainers/sklearn_api_runtime.py
```diff
@@ -0,0 +1,36 @@
...
+            # Try to load as sklearn model first
+            model = joblib.load(inference_model_path)
+        except (IndexError, KeyError, IOError):
+            try:
+                # Try to load as XGBoost model
+                model = load_xgb_model(inference_model_path)
+            except XGBoostError:
+                try:
+                    # Try to load as LightGBM model (do this last as raises warning if not successful)
+                    model = lgb.Booster(model_file=inference_model_path)
```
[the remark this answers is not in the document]
ascillitoe: Thanks, looking again I think you're right about this. The above would result in a raw `Booster` not an sklearn api wrapped model. 

I initially went with `lgb.Booster(...)` because that is what `mlserver_lightgbm` uses:

https://github.com/SeldonIO/MLServer/blob/3aee441aa8f0d22732c97ee879b46a3a7cbd086c/runtimes/lightgbm/mlserver_lightgbm/lightgbm.py#L22

On the other hand, `mlserver_xgboost` results in sklearn API models being loaded:

https://github.com/SeldonIO/MLServer/blob/3aee441aa8f0d22732c97ee879b46a3a7cbd086c/runtimes/xgboost/mlserver_xgboost/xgboost.py#L47

This difference is tricky for the white-box explainers, since we want the latter (i.e. sklearn models). I see a number of options:
1. See if we can just expect all of these white-box models to be given as `joblib` files (sklearn deffo works, as does `lightgbm`. Would need to test `xgboost`.). This results in a pretty problematic difference to the model runtime's themselves.
2. Use `mlserver_xgboost` etc (related to the discussion of assuming `model-settings.json` exists etc). However, would need to modify `mlserver_lightgbm` to wrap models in the SKLearn API (potentially problematic, requires access to `_` attributes.
3. Just stick to sklearn and xgboost models for now. Leave `lightgbm` until later...

runtimes/alibi-explain/tests/test_white_box_sklearn.py
ascillitoe: Renamed in 53ce0b02e7886cc82a75bc52af27da1516207c6b

runtimes/alibi-explain/mlserver_alibi_explain/explainers/sklearn_api_runtime.py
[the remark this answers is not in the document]
adriangonz: Thanks for that context @ascillitoe!

As far as I can see, the ideal solution would be to support both methods in the LightGBM runtime (we'd need both because since the format is different, it would be a breaking change otherwise). We could then use some defaults (and / or settings flags) to decide which method to use in the LGBM runtime (e.g. if file is `model.bst` - use LGBM's `load()`; if file is `model.joblib` - use `joblib`).

The White Box Explainer runtime can then look into that to decide whether it can explain a LGBM model or not (e.g. it could look for a `LGBMRuntime._sklearn_interface: bool` flag that we could add to the LGBM runtime), and raise an error if the LGBM model was saved as a booster?

runtimes/alibi-explain/tests/helpers/sk_model.py
ascillitoe: 75ab2e4c35343f599e944a3f2816355af833c944 removed caching and moved training to fixture

runtimes/alibi-explain/tests/helpers/sk_model.py
ascillitoe: Nope. I'd included as anticipating using it when I add black-box explainer tests requiring tabular data (and a model) e.g. `PartialDependence` test etc (similar to the `TFMNISTModel` in `helper/tf_model.py`). But https://github.com/SeldonIO/MLServer/commit/75ab2e4c35343f599e944a3f2816355af833c944 removes it for now, and can add back in later if needed...

runtimes/alibi-explain/tests/conftest.py
```diff
@@ -323,3 +327,14 @@ def _train_anchor_image_explainer() -> None:
 
     _ANCHOR_IMAGE_DIR.mkdir(parents=True)
     anchor_image.save(_ANCHOR_IMAGE_DIR)
+
+
+@pytest.fixture(scope="module")
+def sk_income_model() -> BaseEstimator:
```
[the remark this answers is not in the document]
ascillitoe: Done in 5f14616d906ba0704857ed40bf203871806acb39!

runtimes/alibi-explain/pyproject.toml
ascillitoe: @adriangonz just to confirm, do you mean like this?

```
mlserver_sklearn = {path = "../../runtimes/sklearn", develop = true}
mlserver_xgboost = {path = "../../runtimes/xgboost", develop = true}
mlserver_lightgbm = {path = "../../runtimes/lightgbm", develop = true}
```

runtimes/alibi-explain/mlserver_alibi_explain/explainers/sklearn_api_runtime.py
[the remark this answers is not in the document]
ascillitoe: I agree that adding support for `.bst` vs `joblib` to `mlserver_lightgbm` might be nice. Not least because I think there might be a potentially confusing difference between `mlserver_xgboost` and `mlserver_lightgbm` right now. If a user  serialised sklearn api compatible classifier's from each library (as `'xgb_model.json'` and `lightgbm_model.bst`), the `mlserver_xgb` runtime would support `infer_output='predict_proba`, whilst the `mlserver_lightgbm` runtime would not (since the loaded `lgb.Booster` does not). 

Above is unrelated to explainers! Wrt to explainers, I think the main point now is deciding if we want to enforce this `model-setting.json` being present in the model dir, or decide which runtime to use for loading via a `model_type` setting in the explainer `model-setting.json`. 

Also, worth noting that if the model is saved via `joblib`, we don't even need to know what library (`xgboost`, `lightgbm` etc) the model belongs to. We can just `joblib.load()` and use the model directly...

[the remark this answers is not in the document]
ascillitoe: Notes from offline discussion:
- Having a separate `model-settings.json` for the model e.g.
   ```
   explainer/
      explainer_artefact/
      model-settings.json
      model/
         model_artefact.bst
         model-settings.json
   ```
   
   would be problematic, since mlserver would recursively search `explainer/` and deploy the model itself too, due to the presence of the extra `model-settings.json`.

- With above in mind, decision to stick with the current `try/except` strategy for now. However, will remove the `lgb.Booster` block since this doesn't return an sklearn api compatible model (user should use `joblib` for `lightgbm` models).

- Open issue wrt discrepancy between `mlserver_xgboost` and `mlserver_lightgbm`.
- Open issue wrt improving loading of white-box models in `SKLearnRuntime`.

runtimes/alibi-explain/mlserver_alibi_explain/explainers/sklearn_api_runtime.py
ascillitoe: Discussed in https://github.com/SeldonIO/MLServer/pull/1279#discussion_r1252798610. Decision to stick with `try/except` strategy for now.

runtimes/alibi-explain/pyproject.toml
adriangonz: Yuuup, exactly

adriangonz (approved):
Nice one @ascillitoe! Thanks for making those changes.

Once the linter passes, this looks good to go from my side! Conscious it's still marked as a draft though, so happy to re-review once you think it's ready.

ascillitoe:
Thanks @adriangonz, I'll make the linter changes now and turn off draft once done!

PR #1281, opened 2023-07-04
Introduce tracing with OpenTelemetry

### LINKED ISSUE #39: Add support for tracing
Trace each inference step within MLServer. These traces can be pushed to Jaeger or similar OpenTracing backends.

vtaskow:
Resolves #39 
The intention is the whole Seldon Core v2 ecosystem to have enabled distributed tracing. Currently, MLServer is missing this and this PR tracing features via the OpenTelemetry Python SDK. Other components of the system already have tracing enabled and exporting traces to an OTel collector service. By having it enabled in settings, it will provide transparency on requests, helping us to get a better picture of the journey of a request and identify any potential or already existing bottlenecks.
This PR adds basic tracing functionality by instrumenting both the REST and the gRPC servers on start. And adds a few tests to verify spans are being sent out. This is enough to give us basic information about incoming requests and their life-cycle in MLServer.


tests/tracing/conftest.py
```diff
@@ -0,0 +1,47 @@
...
+    tracer_provider = TracerProvider(resource=resource)
+    span_processor = SimpleSpanProcessor(span_exporter)
+    tracer_provider.add_span_processor(span_processor)
+
+    mocker.patch("mlserver.rest.app.get_tracer_provider", return_value=tracer_provider)
+    mocker.patch(
+        "mlserver.grpc.server.get_tracer_provider", return_value=tracer_provider
+    )
+    yield tracer_provider
+    tracer_provider.shutdown()
```
adriangonz, line 39: Out of curiosity, do we need to also shutdown the regular OTLP tracer?

adriangonz (commented):
Nice one @vtaskow! PR looks great! :rocket: 

I've added a couple questions, but it's pretty much ready to go.

tests/tracing/conftest.py
vtaskow: When the OTLP tracer is shut down, it sends a signal for every span processor to flush outstanding spans to their corresponding exporters, and for them in turn to stop handling any further incoming spans. I believe we don't need to shut it down manually, because the app is instrumented to run with tracing for its whole duration, if tracing is enabled. If at some point, we decide to implement a feature to disable tracing, via API for example, then I think it would make sense to do that basically to tell the tracer that we don't want to process and handle new spans anymore.

adriangonz: Got it - as long as it cleans itself when MLServer shuts down we should be OK

adriangonz (approved):

PR #1286, opened 2023-07-06
Fix build-wheels.sh error when copying to output path

lc525:
Fixes hack/build-wheels.sh failing when trying to copy the dist over itself (source and output path the same). This occured when doing `make build` for example, where build-wheels.sh is called with an output path of ./dist and tries to copy the mlserver wheels.

### DIFF
--- a/hack/build-wheels.sh
+++ b/hack/build-wheels.sh
@@ -15,24 +15,29 @@ fi
 _buildWheel() {
   local _srcPath=$1
   local _outputPath=$2
-  local _currentDir=$PWD
 
   # Poetry doesn't let us send the output to a separate folder so we'll `cd`
   # into the folder and them move the wheels out
   # https://github.com/python-poetry/poetry/issues/3586
   pushd $_srcPath
   poetry build
-  cp ./dist/* $_outputPath
+  # Only copy files if destination is different from source
+  local _currentDistPath=$PWD/dist
+  if ! [[ "$_currentDistPath" = "$_outputPath" ]]; then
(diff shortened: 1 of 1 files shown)


hack/build-wheels.sh
```diff
@@ -15,24 +15,29 @@ fi
...
 
 _main() {
   # Convert any path into an absolute path
   local _outputPath=$1
+  mkdir -p $_outputPath
   if ! [[ "$_outputPath" = /* ]]; then
-    _outputPath="$PWD/$_outputPath"
+    pushd $_outputPath
+    _outputPath="$PWD"
+    popd
```
adriangonz, line 39: Is this to make the absolute path consistent between diff relative paths?

adriangonz (approved):
Good spot and great fix @lc525 ! :rocket: 

I've added a small clarification question, but changes look good!

hack/build-wheels.sh
lc525: It gets an absolute path for `$_outputPath` that we can more easily compare to `$_srcPath` in `_buildWheel()`. Previously, if the user provided `./dist`, `$PWD/$_outputPath` would end up looking like `/opt/mlserver/./dist` which is more difficult to compare with the source (something like `/opt/mlserver/dist`).

Here, I'm first creating the output dir if it doesn't exist, enter it, and `$PWD` is already set to the absolute path we need.

adriangonz: Thanks for clarifying that!

PR #1288, opened 2023-07-07
System tracing provider for tracing via native probes (BPF)

lc525:
This PR introduces a new tracing provider, enabling dynamic attachment of native (BPF, Systemtap) probes at runtime.
The goal of the provider is to allow correlation between MLServer-specific events and operating system behaviour (system load, performance, resource usage, etc).

## Approach
The provider exposes a number of tracepoints (hook functions without any code attached), which can be triggered (fired)
when particular application-level events happen (e.g. a model gets loaded/unloaded, an inference request joins a queue, etc).

At runtime, external probes (BPF programs, Systemtap scripts) can be attached to those tracepoints and to perform tracing actions (measurements, in-linux-kernel data aggregations, correlating MLServer context with OS context, etc).

The underlying implementation creates and dynamically links a native shared library where the tracepoint hooks exist as functions containing just a couple of nop instructions. When external probes are attached, the code of those tracepoints is modified at runtime to jump into the tracing code.

## Features
- Completely optional feature, with the ability to enable it via the `tracepoints` extra
- The exact tracepoints exposed for external probing configurable via settings
- Near-zero overhead when not in use (no external probes attached). Because tracepoints look like normal functions from the perspective of python code, and receive arguments which might need to be computed, the provider offers a way of conditionally computing the tracepoint arguments based on whether an external probe has been attached to a particular tracepoint or not.

## Introduced dependencies
- When the `tracepoints` extra is enabled, MLServer will require additional dependencies, as follows:
  * [stapsdt](https://github.com/linux-usdt/python-stapsdt)
  * [libstapsdt](https://github.com/linux-usdt/libstapsdt) (a native library, in turn requiring `libelf`)
- MLServer will continue to work as normal even if the extra is enabled but the dependencies are not met
- The Dockerfile has been updated to allow for container images in which system tracing dependencies are installed

## TODOs
- [ ] Add instrumentation to the inference request path (including for batched requests)
- [ ] Stabilise tracepoint arguments
- [ ] Write documentation example for simple usage via `bpftrace`
- [ ] Move to "off-by-default" settings in settings/Dockerfile (left to enabled atm for testing)


mlserver/types/__init__.py
```diff
@@ -45,4 +47,8 @@
     "RepositoryIndexResponse",
     "RepositoryLoadErrorResponse",
     "RepositoryUnloadErrorResponse",
+    # Tracing
+    "Tracepoint",
+    "ArgStatus",
+    "MAX_TRACEPOINT_ARGS",
```
adriangonz, line 53: I'd move these types into the `mlserver.sys_tracing` module to keep them closer to their own scope.

mlserver/types/tracepoints.py
```diff
@@ -0,0 +1,186 @@
...
+ArgTypeMap._prototypes = {
+    Tracepoint.model_load_begin: ArgTypeMap.model_args,
+    Tracepoint.model_load_end: ArgTypeMap.model_args,
+    Tracepoint.model_reload_begin: ArgTypeMap.model_args_reload,
+    Tracepoint.model_reload_end: ArgTypeMap.model_args,
+    Tracepoint.model_unload: ArgTypeMap.model_args,
+    Tracepoint.inference_enqueue_req: ArgTypeMap.queue_args,
+    Tracepoint.inference_begin: ArgTypeMap.infer_args,
+    Tracepoint.inference_end: ArgTypeMap.infer_args,
+}
```
adriangonz, line 186: Thinking out loud here, but the tracepoints and args seem very coupled with the implementation - would it be possible to make those more generic? 

My main worry would be whether a small change into any of those implementations (e.g. changing the args for inference) could then cause any side effects around system tracing.

tests/testdata/settings-tracing.json
```diff
@@ -0,0 +1,13 @@
...
+  "tracepoint_settings": {
+    "enable_tracepoints": true,
+    "configured_tracepoints": [
+        "model_load_begin",
+        "model_load_begin",
+        "model_load_end",
+        "model_reload_begin",
+        "model_reload_end",
+        "model_unload"
+    ]
```
adriangonz, line 11: Would there be any reason not to enable all the tracepoints? Or put another away, is there any trade-off to always enable all of them to simplify the settings?

mlserver/sys_tracing/stapsdt_stub.py
```diff
@@ -0,0 +1,31 @@
+class Provider(object):
```
adriangonz, line 1: Are these stubs just used when the underlying library is not present?

mlserver/registry.py
```diff
@@ -155,6 +156,7 @@ async def _load_model(self, model: MLModel):
             # appears as a not-ready (i.e. loading) model
             self._register(model)
 
+            sys_tracer.tp_model_load_begin(model.name, model.version)
```
adriangonz, line 159: Thinking out loud here as well - it looks like this would fit really well as a context manager (i.e. as `with sys_tracer.tp_model_load(): .....`)

Dockerfile
```diff
@@ -53,7 +102,16 @@ RUN microdnf update -y && \
         libgomp \
         mesa-libGL \
         glib2-devel \
-        shadow-utils
+        shadow-utils && \
+        if [[ "${OPT_MLSERVER_SYS_TRACING}" == "sdt-sys-tracing" ]]; then \
+            microdnf install -y \
+                elfutils-libelf; \
+        fi
```
adriangonz, line 109: Would there be any trade-off to always installing it on the image? 

Mainly thinking that, if we want to make this an opt-in feature within MLServer, then it makes sense it's always possible to enable it on the official set of images. Otherwise, we would need to build a second set of images which have these installed.

Dockerfile
```diff
@@ -107,7 +169,12 @@ RUN . $CONDA_PATH/etc/profile.d/conda.sh && \
             pip install $_wheel --constraint ./dist/constraints.txt; \
         done \
     fi && \
-    pip install $(ls "./dist/mlserver-"*.whl) --constraint ./dist/constraints.txt && \
+    if [[ ${#_extras[@]} -gt 0 ]]; then \
+        extras_list=$(IFS=, ; echo "${_extras[@]}") && \
```
adriangonz, line 173: Following the same reasoning as the other comment, unless we have some reason to avoid it I'd install all the extras in the Dockerfile - that way the `seldonio/mlserver` image is always ready as-is to enable tracepoints.

Dockerfile
```diff
@@ -18,19 +28,57 @@ COPY \
...
         --without-hashes \
+        "${_extras[@]}" \
         --format constraints.txt \
         -o /opt/mlserver/dist/constraints.txt && \
     sed -i 's/\[.*\]//g' /opt/mlserver/dist/constraints.txt
 
+
+# Build native dependencies for tracepoints; 
+# Almalinux is binary-compatible with rhel ubi images but contains repositories
+# with additional devel packages (elfutils-libelf-devel needed here)
```
adriangonz, line 45: Thanks for the note! I was just wondering about the need for `almalinux`!

adriangonz (commented):
This looks amazing @lc525! :rocket: 

Conscious this is still a WIP draft, but I've added a few comments and some questions below. Happy to chat about how you foresee this being used!

mlserver/types/__init__.py
lc525: Those types were in `mlserver.sys_tracing` initially. The problem is that it creates a circular dependency; I'll see if I can sort it out more elegantly.

Depending on what we decide to do on the Settings side (and whether to simplify or not the available settings) this could be solved trivially.

mlserver/types/tracepoints.py
lc525: In a way, they do end quite coupled indeed. However, the prototypes essentially become the public interface that external probes use to run their code. Any changes to arguments, besides simply adding new ones at the end of the list, would require changes for whatever external BPF probes people write against the interface. So we would need to stabilise them at some point, after we write more actual probing code.

The idea is to pass to probes the minimal information that is _required_ for meaningful tracing applications. Hopefully, this information remains relatively stable even as MLServer evolves. For example, the actual inference functions might change without us changing the data passed for tracing via the same Tracepoints.

That being said, the `_prototypes` defined here do not end up being "enforced". Say we add a new argument to the `inference_begin` tracepoint without changing any of the `tp_*` functions in the provider. All existing code will continue to work because the function for triggering the tracepoint (`Probe.fire`) receives a variable number of arguments. We can then selectively call the tracepoint with additional arguments _sometimes_ by directly using the `SystemTracingProvider.__call__(tracepoint, *args)` (I don't necessarily advocate that it's a sane thing to do, as code might end up being confusing in the long term wrt to what arguments should be passed to which tracepoints).

We can discuss further, perhaps there is a better design to be found here.

tests/testdata/settings-tracing.json
lc525: Some groups of tracepoints will inevitably add more overhead then others (when active). For example, tracepoints that execute per inference will inevitably add more overhead because they will fire a lot more often than the ones related to models.

This would be a way to stop a group of tracepoints from the "source" side (never expose them for tracing tools). The other view, which would simplify things quite a bit, is that those groups of tracepoints should be disabled on the tracing tool side (for example, instruct it to never attach any per-inference probes).

Another thing to consider is whether some MLServer users will want to disable sets of tracepoints because they consider them side-channels  for sensitive information (for example, if things could be deduced from the relative timing of inference requests, their latency, etc). I imagine this is rare though and such users would want to disable tracing entirely.

mlserver/sys_tracing/stapsdt_stub.py
lc525: The stubs are used both when the underlying library is not present and for any tracepoints disabled via configuration.

mlserver/registry.py
lc525: Cool idea! This could be quite useful for tracepoints that are often fired in pairs within the same function. 

Sometimes the "begin" tracepoint will be fired in a different function than the "end", so we'll need to retain the ability to fire them individually.

Dockerfile
lc525: The main reason I wanted to offer this as an option is in case there are users who see `libelf` and `libstapsdt` as increasing their attack surface in the container. So perhaps all images provided by Seldon would have those installed, but users building their own have the ability of not adding them in.

If that represents just a couple of users requiring extreme security, perhaps the way forward would be to just offer a simple way of removing those from an existing image rather than installing extras conditionally.

Dockerfile
lc525: Happy to do that if the decision (given above comments) is to install all extras. I don't know if pip has a simple way of doing that here, I might still have to build a list.

Dockerfile
lc525: I'll also investigate if I can easily add one of the `almalinux` repositories to `ubi9/ubi-minimal` so that we can reuse the cached ubi-minimal image.

mlserver/types/tracepoints.py
adriangonz: Thanks for adding that extra context. 

Do you have an example of how it looks on the "other end", i.e. where those probes are used?

tests/testdata/settings-tracing.json
adriangonz: Got it! In that case, to keep settings simple, I'd consider just having a single settings flag to either disable or enable all of them (we can always extend configurability later on if there's a need for it).

Dockerfile
adriangonz: Got it - I'd then just remove the option from the `Dockerfile` and always install those. This simplifies maintenance of the Dockerfile, and if we need to we can always add this back later on.

Dockerfile
adriangonz: We can also see if Poetry supports an `all` extras target? Otherwise we can always add one.

adriangonz: (and / or just hardcode a `pip install mlserver[extras1, extras2, etc.]`)

Dockerfile
adriangonz: Probably not worth it TBH - having tried it in the past, it tends to be a pain adding repos to UBI images (some of them are also behind "paywalls" and can only be used if you have a RHEL subscription).

Dockerfile
lc525: Poetry currently supports --all-extras on install but not export (this is tracked in python-poetry/poetry-plugin-export#45). I've added an `everything` extra. I've intentionally avoided the name `all` as that might be reserved in the future by poetry/pip, but we can move to it once there is enough support.

adriangonz: TBH I would just use `all` as it's more standard. If poetry decides to always add one automatically we can just remove it and use Poetry's built-in target.

PR #1296, opened 2023-07-13
Conda pack unsilent

Kolajik:
Partially reverting #1230 PR. We would love to see the logging from conda pack actually, but keep conda-unpack silent. This then ensures the user to see if conda env was created successfully, which is important.


dtpryce:
@adriangonz this is one of my team ... we realised we actually want build logs just not inference logs from conda / pack or unpack.

Kolajik:
Change made in #1298 in the end. So it's going to be merged with that PR.

Issue #1297, opened 2023-07-13
Move conda resolver to libmamba


dtpryce:
Building custom runtimes using `mlserver build` is fantastic but also leverages conda and some of our builds have been known to take up to 8 hours due to the conda standard resolver taking so long to resolve the environment. An improvement has already been developed by conda: the libmamba resolver.

Details can be found here: https://www.anaconda.com/blog/a-faster-conda-for-a-growing-community

But the highlights:
- make sure to be using conda 22.11 or higher (`conda update -n base conda`)
- install the new solver: `conda install -n base conda-libmamba-solver`
- set as default: `conda config --set solver libmamba`

dtpryce:
Discovered that the conda version has to be 22.11 or higher ... PR in progress

PR #1298, opened 2023-07-14
 Bump in conda version and mamba solver 

dtpryce:
- Bump conda to 22.11.1
- Install and set solver to libmamba
- Tested with a custom runtime build - much much faster
- Tested running the server based on docker image in V2
- Removed the quiet from the build after devs complained they wanted it back there but not from server starting like #1296 

Should solve #1297

### DIFF
--- a/mlserver/cli/constants.py
+++ b/mlserver/cli/constants.py
@@ -9,7 +9,7 @@ ARG MLSERVER_ENV_NAME="mlserver-custom-env" \\
 RUN conda config --add channels conda-forge && \\
     conda install conda-libmamba-solver && \\
     conda config --set solver libmamba && \\
-    conda install conda-pack 
+    conda install conda-pack
 
 # The `[]` character range will ensure that Docker doesn't complain if the
 # files don't exist:
--- a/mlserver/cli/constants.py
+++ b/mlserver/cli/constants.py
@@ -1,13 +1,15 @@
 DockerfileName = "Dockerfile"
 DockerfileTemplate = """
-FROM continuumio/miniconda3:4.12.0 AS env-builder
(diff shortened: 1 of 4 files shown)


mlserver/cli/constants.py
```diff
@@ -1,12 +1,14 @@
 DockerfileName = "Dockerfile"
 DockerfileTemplate = """
-FROM continuumio/miniconda3:4.12.0 AS env-builder
+FROM continuumio/miniconda3:22.11.1 AS env-builder
```
adriangonz, line 3: I just noticed that the main `Dockerfile` actually uses `23.1.0`, so should we use that instead to make sure they're aligned?

adriangonz (commented):
Nice one @dtpryce! :rocket: 

Added a small comment around the Conda version, but everything else looks good! Once this is in, we may also look into enabling libmamba at the base `Dockerfile` image.

mlserver/cli/constants.py
dtpryce: I wasn't able to test with that version of the docker image on a linux host machine and got some awful c++ issues locally with our custom runtimes (might be me trying to test on a Mac host but not entirely sure) ... still want me to bump to highest conda image?

adriangonz: Fair enough - in that case let's go ahead with the PR as-is :+1:

dtpryce: I would add one thing to consider for the future here ... image CVEs are becoming a big part of our work so having both build and inference image at the highest python and even debian as well as library versions is becoming something we focus on. I can see you guys being super proactive on lib CVEs via dependabot but this is a level below that to consider also.

PR #1317, opened 2023-07-27
Remove fastapi dependency upper limit

wdhorton:
This is messing up my ability to install mlserver alongside other packages, and I can't see the reason why the fastapi version should have an upper limit. Almost all the other dependencies are either "*" or "^..."

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,7 +45,7 @@ addopts = "--import-mode=importlib"
 [tool.poetry.dependencies]
 python = "^3.8.1,<3.12"
 click = "*"
-fastapi = ">=0.88.0,!=0.89.0,<=0.97.0"
+fastapi = ">=0.88.0,!=0.89.0"
 python-dotenv = "*"
 grpcio = "*"
 numpy = "*"


adriangonz:
Hey @wdhorton ,

Unfortunately, we've been bitten in the past by breaking changes in FastAPI. Therefore, we can't remove the upper ceiling there.

We can expand the range though. Do you know what package is incompatible with the current set of constraints?

adriangonz:
Hey @jlopezpena ,

Sure thing - do you know which version of FastAPI enables that version of Starlette?

adriangonz:
Current aim is to bump the version of Pydantic to `2.x` (and FastAPI to latest). There are quite a few breaking changes in Pydantic `2.x` so it's taking a bit longer than expected.

In the meantime, we're currently thinking of bumping the ceiling in the `3.5.x` branch of MLServer. 

If I'm not mistaken, latest `master` should already be OK though?

adriangonz:
Thanks for confirming that @jlopezpena .

In that case, we'll probably go ahead cutting a `1.4.0` release with that fix, as well as other improvements and CVEs.

PR #1319, opened 2023-07-28
feat(huggingface): Load local artefacts in HuggingFace runtime

### LINKED ISSUE #1076: Load local artefacts in HuggingFace runtime
Support loading artifacts from provided model-URI

vtaskow:
# Why 
Currently, it is not convenient to load local model artefacts in the HuggingFace runtime. In the scenario when users would like to specify a custom model or perhaps would like to avoid downloading a model while mlserver operates, there must be a convenient way to specify a location to the desired model artefacts folder.
Furthermore, there is a way to specify such a location now by using the `parameters.extra.pretrained_model` in `model-settings.json` config. However, the intention of this field is to be used for ready models from the HuggingFace hub.

# What and How
This PR provides a new way for a user to use a local HuggingFace model, not having to download a ready one from the HuggingFace hub or to limit themselves to those available there. Essentially, if the `parameters.extra.pretrained_model` is not specified in `model-settings.json`, then `parameters.uri` will be used to locate and load a custom model that is already downloaded by rclone. Furthermore, `parameters.uri` if not set, it's automatically filled by the HuggingFace runtime to point to the default location for downloading models in MLServer.

### Background
PREDICTIVE_UNIT_PARAMETERS is an environment variable containing a list of parameters describing a HuggingFace model. It is the predecessor or `model-settings.json`. In SCv1, if this variable is not empty, it is used instead of `model-settings.json`(even if supplied).

### PREDICTIVE_UNIT_PARAMETERS is empty AND `model-settings.json` is missing
- [x] provided gs path to HF model artefacts folder not containing `model-settings.json`; no UI params set
Result: raises MissingHuggingFaceSettings exception because there are no predictive params set or `model-settings.json` EVEN THOUGH MLSERVER_MODEL_URI is auto set to /mnt/models

### PREDICTIVE_UNIT_PARAMETERS is not empty AND `model-settings.json` is missing
- [x] PREDICTIVE_UNIT_PARAMETERS contained: task = text-generation; MLSERVER_MODEL_URI is auto set to /mnt/models
Result: able to initialise HF settings, params has task = text-generation; HF settings -> pretrained_model is not set but falls back to MLSERVER_MODEL_URI - **this is the change I am making in this PR**, Able to predict :heavy_check_mark:  
 
- [x] PREDICTIVE_UNIT_PARAMETERS contained: task = text-generation, pretrained_model = /mnt/models(the same as MLSERVER_MODEL_URI which is auto set to that value)
Result: able to initialise HF settings, params has task = text-generation and pretrained_model = /mnt/models which is being used as a location for the custom model. Able to predict :heavy_check_mark: 

### PREDICTIVE_UNIT_PARAMETERS is empty and  using `model-settings.json`
- [x] MLSERVER_MODEL_URI is auto set to /mnt/models and `model-settings.json` contained:
```
{
  "name": "transformer",
  "parameters": {
    "extra": {
      "task": "text-generation"
    }
  }
}
```
Result: able to initialise HF settings, pretrained model is not used because it's empty, correctly falls back to using MLSERVER_MODEL_URI instead which points to /mnt/models where rclone has downloaded the model artefact folder. :heavy_check_mark:
**Important to note: no models were downloaded on demand from the HuggingFace hub.**

- [x] MLSERVER_MODEL_URI is auto set to /mnt/models and `model-settings.json` contained:
```
{
  "name": "transformer",
  "parameters": {
    "extra": {
      "pretrained_model": "/mnt/models",
      "task": "text-generation"
    }
  }
}
```
Result: able to initialise HF settings, pretrained model IS used and correctly points to /mnt/models where rclone has downloaded the model artefact folder :heavy_check_mark:
**Important to note: no models were downloaded on demand from the HuggingFace hub.**

- [x] MLSERVER_MODEL_URI is auto set to /mnt/models and `model-settings.json` contained:
```
{
  "name": "transformer",
  "parameters": {
    "uri": "/mnt/models",
    "extra": {
      "task": "text-generation"
    }
  }
}
```
Result: able to initialise HF settings, uri IS used and correctly points to /mnt/models where rclone has downloaded the model artefact folder :heavy_check_mark:
**Important to note: no models were downloaded on demand from the HuggingFace hub.**

### `model-settings.json` is missing
Message: ModelFailed
Reason: Can't create model-settings from requirements [huggingface]
Result is expected :heavy_check_mark: 

### `model-settings.json` doesn't have `parameters.extra`
```
{
  "name": "transformer",
  "parameters": {
    "extra": {}
  }
}
```
mlserver_huggingface.errors.MissingHuggingFaceSettings: Missing HuggingFace Runtime settings.                                                                                
Result is expected :heavy_check_mark: 

### `model-settings.json` doesn't have `parameters.extra` but has `parameters.uri`
```
{
  "name": "transformer",
  "uri": "/mnt/models",
  "parameters": {
    "extra": {}
  }
}
```
mlserver_huggingface.errors.MissingHuggingFaceSettings: Missing HuggingFace Runtime settings.                                                                                
Result is expected :heavy_check_mark: 

### `model-settings.json` has only `task` field in `parameters.extra`
```
{
  "name": "transformer",
  "parameters": {
    "extra": {
      "task": "text-generation"
    }
  }
}

```
Result: It's correctly using the downloaded model in /mnt/agent/models because even though neither `parameters.uri` is set nor `parameters.extra.pretrained_model`, the runtime model registry by default fills in `parameters.uri` to be the `model-settings.json` path. Able to predict correctly :heavy_check_mark: 
**This is possible because of the fallback to `parameters.uri` if `parameters.extra.pretrained_model` is not set.**

### `model-settings.json` has `task` and `pretrained_model` fields in `parameters.extra` but `pretrained_model` points to a HuggingFace hub model
```
{
  "name": "transformer",
  "parameters": {
    "extra": {
     "pretrained_model": "distilgpt2",
      "task": "text-generation"
    }
  }
}
```
Result: The specified pretrained model is correctly downloaded from the HuggingFace hub. Able to predict correctly :heavy_check_mark: 

## Further testing with model artefacts folder created from `transformers.pipeline.save_pipeline()`
```
{
  "parameters": {
    "extra": {
      "task": "text-generation"
    }
  }
}
```
 - [x] Works for SCv1 with `model-settings.json`
 - [x] Works for SCv1 without `model-settings.json` but with PREDICTIVE_UNIT_PARAMS
 - [x] Works for SCv1 with `model-settings.json` but with PREDICTIVE_UNIT_PARAMS
 - [x] Works for SCv2 with `model-setting.json`

Resolves #1076


mlserver/cli/serve.py
```diff
@@ -23,7 +23,7 @@ async def load_settings(
     if folder:
         sys.path.insert(0, folder)
 
-    settings = None
+    settings: Settings
```
adriangonz, line 26: Is this valid as-is? I would add back the `= None` for explicitness (i.e. `settings: Optional[Settings] = None`).

runtimes/huggingface/README.md
```diff
@@ -55,6 +55,36 @@ MLSERVER_MODEL_HUGGINGFACE_OPTIMUM_MODEL=true
...
+  "implementation": "mlserver_huggingface.HuggingFaceRuntime",
+  "parameters": {
+      "uri": "./models/your-model-artefacts-folder",
+      "version": "v1.0",
+      "extra": {
+          "task": "question-answering",
+      }
+  }
+}
+```
```
adriangonz, line 79: I think this should be already covered on other sections of the docs (i.e. both around MMS and also around relative paths, etc.) so I'd suggest removing it in favour of the other sections.

runtimes/huggingface/README.md
```diff
@@ -55,6 +55,36 @@ MLSERVER_MODEL_HUGGINGFACE_OPTIMUM_MODEL=true
...
+
+````{note}
+If a `parameters.extra.pretrained_model` parameter is specified in `parameters.extra` section, it will take precedence 
+over `parameters.uri`. Furthermore, in the case of MMS, `parameters.extra.pretrained_model` and `parameters.uri` can 
+be used interchangeably only if the paths provided are absolute. This is because the `parameters.extra.pretrained_model` 
+expects a relative path to itself, i.e. relative to the `model-settings.json`.
+````
```
adriangonz, line 86: Conscious that `pretrained_model` can also be a model name, doesn't need to be a filesystem path. Therefore, I'd prepend the second sentence with that.

Also, MMS should be orthogonal to this so I'd probably remove that reference. You can use the same model settings loading one or multiple models.

adriangonz (changes requested):

runtimes/huggingface/README.md
vtaskow: Good point about the model name!

adriangonz (approved):
Nice one! Changes look good from my side - this should be good to go once tests are green. :rocket:

runtimes/huggingface/tests/test_common.py
```diff
@@ -1,6 +1,8 @@
+from unittest.mock import patch, MagicMock
```
agrski, line 1: :question: Is there a particular order defined for imports?  I can't see grouping by standard library vs. third-party groupings,  so not sure if there is an expectation around this

runtimes/huggingface/tests/test_common.py
```diff
@@ -45,3 +47,48 @@ def test_load_pipeline(optimum_model: bool, expected):
...
+    [
+        (None, True, None, None),
+        (None, True, "", ""),
+        (None, True, "/some/folder/model-artefacts", "/some/folder/model-artefacts"),
+        (None, False, "N/A", None),
+        ("", True, None, None),
+        ("", True, "", ""),
+        ("", True, "/some/folder/model-artefacts", "/some/folder/model-artefacts"),
+        ("", False, "N/A", ""),
+        ("./some-pretrained-folder", "N/A", "N/A", "./some-pretrained-folder"),
```
agrski, line 63:
```suggestion
        ("./some-pretrained-folder", True, "N/A", "./some-pretrained-folder"),
        ("./some-pretrained-folder", False, "N/A", "./some-pretrained-folder"),
```
:wrench: The use of "N/A" for `has_model_params` breaks the type hint that this should be a Boolean if provided.

I'm also proposing two cases, so we can clearly see that the pre-trained URI takes precedence, even when the general URI is available.

agrski (commented):
A few suggestions to improve clarity and UX, but nothing major that I can see

runtimes/huggingface/mlserver_huggingface/common.py
```diff
@@ -23,14 +23,16 @@
...
-    # uri = model_parameters.uri
     pipeline = _get_pipeline_class(hf_settings)
 
     batch_size = 1
     if settings.max_batch_size:
         batch_size = settings.max_batch_size
 
+    model = hf_settings.pretrained_model
+    if not model:
+        if settings.parameters is not None:
```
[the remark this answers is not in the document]
vtaskow: In the context of the huggingface runtime, `settings.parameters` is always not empty, otherwise this function [get_huggingface_settings](https://github.com/SeldonIO/MLServer/blob/08e7c86e8052149ac73171894be45ed0303cbc31/runtimes/huggingface/mlserver_huggingface/settings.py#L146) will raise an exception. And also, [settings.parameters.uri is always present](https://github.com/SeldonIO/MLServer/blob/08e7c86e8052149ac73171894be45ed0303cbc31/mlserver/repository/load.py#L7C60-L7C60)(if it's not present in the `model-settings.json`, it will be set to the default model path by the registry load method).
Let me know if that makes sense to you, if not I am open to change :+1:

runtimes/huggingface/README.md
vtaskow, line 1: I have rewritten the docs here. Let me know that you think :pray:

mlserver/repository/factory.py
vtaskow, line 1: Reverted the change to keep the scope small.

runtimes/huggingface/tests/test_common.py
vtaskow: Fixed!

runtimes/huggingface/README.md
```diff
@@ -55,6 +55,19 @@ MLSERVER_MODEL_HUGGINGFACE_OPTIMUM_MODEL=true
 ```
 ````
 
+### Loading models
+#### Local models
+It is possible to load a model from a local path into a HuggingFace pipeline by leaving both `parameters.uri` and `parameters.extra.pretrained_model` empty in the `model-settings.json`.
+In this case, the HuggingFace runtime will populate `parameters.uri` to have the value of the model folder where `rclone` will have
+downloaded the model artefacts folder from the specified in the UI remote location.
```
adriangonz, line 62: I wouldn't reference `rclone` here TBH. Instead, I would just keep it simple and point out that if you want to load a local model you can use the `parameters.uri` field to point to where your local artefact is located.

Main reason for this is that rclone is an implementation detail from Seldon Core - i.e. MLServer doesn't know anything about that. In fact, MLServer can be used outside Seldon Core (in which case, rclone wouldn't be present).

vtaskow: Good point. I will leave it out. Thanks!

adriangonz (approved):
Nice one @vtaskow ! Thanks for making that change - looks great from my side.

runtimes/huggingface/mlserver_huggingface/common.py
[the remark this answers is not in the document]
agrski: Thanks for clarifying.  Makes sense that we need certain parameters and validate this first.  The links are very helpful, thanks!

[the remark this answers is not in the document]
agrski: One follow-up question: if `settings.parameters` is always provided, there should be no need to check `if settings.parameters is not None`, right?

runtimes/huggingface/tests/test_common.py
```diff
@@ -45,3 +47,82 @@ def test_load_pipeline(optimum_model: bool, expected):
...
+        (None, True, "/some/folder/model-artefacts", "/some/folder/model-artefacts"),
+        (None, False, "", None),
+        ("", True, None, None),
+        ("", True, "", ""),
+        ("", True, "/some/folder/model-artefacts", "/some/folder/model-artefacts"),
+        ("", False, "", ""),
+    ],
+)
+@patch("mlserver_huggingface.common._get_pipeline_class")
+def test_pipeline_was_initialised_when_pretrained_model_is_not_supplied(
```
agrski, line 103: :confused: I'm confused by these tests for two reasons.  While the actual logic makes sense, the test names are verbose (what does "was initialised" mean in the context of the tests?) and the test logic is _identical_ between the two tests bar the value for `pretrained_model` as far as I can see.

Is this not clearer as a single, tabulated test?

runtimes/huggingface/tests/test_common.py
agrski: I meant for blocks of imports, rather than imports within a single line.  It's okay if not, but wanted to clarify what I meant

agrski (commented):
Still a couple of questions about structure/format but, as before, functionally this looks simple and correct

runtimes/huggingface/tests/test_common.py
vtaskow: Good point about the test logic, I've refactored it to be more concise. About the test name: pipeline_is_initialised, I mean that when pipeline(name, model, ....) constructor is called, the correct model param is passed.

runtimes/huggingface/tests/test_common.py
vtaskow: Also, `black` settings are configured from pyproject.toml in the root folder under `[tool.black]` section. In this case, the line length is not set so it's using a default length of 88 chars.

runtimes/huggingface/mlserver_huggingface/common.py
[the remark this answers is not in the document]
vtaskow: And to further support the claim that `settings.parameters` is always provided:
In case there are no `model-setting.json` files found from the current path, [there is a fallback to create model settings from the environment, where also it is the case that model params are initialised and therefore not empty](https://github.com/SeldonIO/MLServer/blob/08e7c86e8052149ac73171894be45ed0303cbc31/mlserver/repository/repository.py#L55).

agrski (approved):

Issue #1249, opened 2023-06-26
Huggingface models frequently fail with batch size greater than 0.


ajsalow:
When batch size is greater than 0[ this line ](https://github.com/SeldonIO/MLServer/blob/master/runtimes/huggingface/mlserver_huggingface/common.py#L65C51-L65C63)will often fail with the following error since many huggingface models do not have `eos_token` defined in their config.

```
Traceback (most recent call last):
  File "/opt/conda/lib/python3.8/site-packages/mlserver/parallel/worker.py", line 158, in _process_model_update
    await self._model_registry.load(model_settings)
  File "/opt/conda/lib/python3.8/site-packages/mlserver/registry.py", line 293, in load
    return await self._models[model_settings.name].load(model_settings)
  File "/opt/conda/lib/python3.8/site-packages/mlserver/registry.py", line 148, in load
    await self._load_model(new_model)
  File "/opt/conda/lib/python3.8/site-packages/mlserver/registry.py", line 165, in _load_model
    model.ready = await model.load()
  File "/opt/conda/lib/python3.8/site-packages/mlserver_huggingface/runtime.py", line 27, in load
    await asyncio.get_running_loop().run_in_executor(
  File "/opt/conda/lib/python3.8/concurrent/futures/thread.py", line 57, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/opt/conda/lib/python3.8/site-packages/mlserver_huggingface/common.py", line 65, in load_pipeline_from_settings
    hf_pipeline.tokenizer.pad_token_id = [str(eos_token_id)]  # type: ignore
  File "/opt/conda/lib/python3.8/site-packages/transformers/tokenization_utils_base.py", line 1218, in pad_token_id
    self._pad_token = self.convert_ids_to_tokens(value) if value is not None else None
  File "/opt/conda/lib/python3.8/site-packages/transformers/tokenization_utils_fast.py", line 313, in convert_ids_to_tokens
    index = int(index)
ValueError: invalid literal for int() with base 10: 'None'
```

Examples of Huggingface models that do not have `eos_token` specified in their configs.
https://huggingface.co/dslim/bert-base-NER/tree/main
https://huggingface.co/hf-internal-testing/tiny-bert-for-token-classification

A couple thoughts I have on this code.
```
if settings.max_batch_size:
        model = hf_pipeline.model
        eos_token_id = model.config.eos_token_id
[...]

PR #1336, opened 2023-08-15
Improved huggingface batch logic

ajsalow:
Addresses issues described in https://github.com/SeldonIO/MLServer/issues/1249 where batch size greater than 0 causes some hugging face models to fail loading.

### DIFF
--- a/runtimes/huggingface/mlserver_huggingface/common.py
+++ b/runtimes/huggingface/mlserver_huggingface/common.py
@@ -69,9 +69,9 @@ def load_pipeline_from_settings(
                 hf_pipeline.tokenizer.pad_token_id = [str(eos_token_id)]  # type: ignore
             else:
                 logger.warning(
-                    "Model has neither pad_token or eos_token defined, setting batch size to 1"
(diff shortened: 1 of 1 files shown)


adriangonz (changes requested):
Changes look good! Thanks for taking the time to contribute this patch @ajsalow. 

Before merging, could you add a test covering the new behaviour? Besides that, there also seem to be some conflicts with latest `master`.

ajsalow:
Thanks for taking a look @adriangonz. I've added some test coverage, let me know if anything else is needed.

adriangonz (approved):
Looks good! Thanks for adding that test @ajsalow :+1:

adriangonz:
Hey @ajsalow ,

Just had a look at the test failures. I think from those, the `lint` failure is probably the key one - everything else seems to be just flaky. So once you get a chance to address the linter error, we should be good to merge this one:

```
./runtimes/huggingface/mlserver_huggingface/common.py:72:89: E501 line too long (120 > 88 characters)
```

PR #1337, opened 2023-08-16
feat(huggingface): Merge model settings

### LINKED ISSUE #1329: [huggingface] Merge predictive unit parameters from env with model settings
# Background
## What are PREDICTIVE_UNIT_PARAMETERS
This is a collection of parameters passed though the environment to a HuggingFace model. This is opposed to declaring them in a `model-settings.json` file as it is the case in Seldon Core v2. As of this moment, those parameters are injected via Seldon Core v1 only, where as Seldon Core v2 uses `model-settings.json` to provide the metadata for the model.

PREDICTIVE_UNIT_PARAMETERS are used only  by the HuggingFace runtime and injected only by SCv1.

You can find the code for creating `HuggingFaceSettings` in `./runtimes/huggingface/ml-server-huggingface/settings.py` fle along with functions for parsing those params from env vars or from `model-settings.json`  

# What is the problem
Currently, `HuggingFaceSettings` are created either from parsing the PREDICTIVE_UNIT_PARAMETERS from the environment OR from the `model-settings.json`. Meaning that if there is at least one parameter set in the env var, the `model-settings.json` extra parameters will be ignored. This makes it cumbersome when a deployment is created from the UI because additional params such as `task`, `pretrained_model`, `pretrained_tokenizer`, `framework`, etc. will have to be added one by one in the Wizard. Why they have to be added from the wizard and not just specified in `model-settings.json` - because currently SCv1 always injects `model_uri` param so the PREDICTIVE_UNIT_PARAMETERS env var so it's not empty. Because this var is not empty, the HF settings are initialised from it and the `model-settings.json` is ignored.

# What needs to be done
When creating HuggingFace settings, env vars needs to be merged with params from `model-settings.json`, giving priority to env vars. For example:
If such env var exists:
```
PREDICTIVE_UNIT_PARAMETERS = [{"name":"model_uri","value":"/mnt/models","type":"STRING"}, {"name":"task_suffix","value":"else","type":"STRING"}]
```
and such `model-settings.json` file exists:
```
{
    "name": "transformer",
    "parameters": {
        "extra": {
            "task": "text-generation",
            "task_suffix": "something",
            "framework": "pt"
        }
    }
}
```
The outcome should be that the `task` parameter doesn't need to be specified in the wizard and
The HuggingFace settings should contain values: task = text-generation, task_suffix = else, framework = pt

# Scope
This only relating to `HuggingFace` runtime and when it's used from SCv1 and only valid as long as SCv1 is still operational and related code is present in MLServer.

jesse-c:
If settings for the model are provided in both the `model-settings.json` *and* environment variables, merge them.

Precedence is given to environment variables.

FIXES #1329

### DIFF
--- a/runtimes/huggingface/tests/test_settings.py
+++ b/runtimes/huggingface/tests/test_settings.py
@@ -1,0 +1,168 @@
+import pytest
+
+from typing import Dict
+
+from mlserver.settings import ModelSettings, ModelParameters
+
+from mlserver_huggingface.runtime import HuggingFaceRuntime
+from mlserver_huggingface.settings import (
+    HuggingFaceSettings,
+    PARAMETERS_ENV_NAME,
+    get_huggingface_settings,
+    merge_huggingface_settings_extra,
+)
+from mlserver_huggingface.errors import MissingHuggingFaceSettings
+
+
+@pytest.fixture()
+def model_settings_extra_task():
+    return ModelSettings(
+        name="foo",
+        implementation=HuggingFaceRuntime,
+        parameters=ModelParameters(
+            extra={"task": "text-generation", "pretrained_model": "distilgpt2"}
+        ),
+    )
+
+
+@pytest.fixture()
+def model_settings_extra_none():
+    return ModelSettings(
+        name="foo",
+        implementation=HuggingFaceRuntime,
+        parameters=ModelParameters(extra=None),
+    )
+
+
+@pytest.fixture()
+def model_settings_extra_empty():
+    return ModelSettings(
+        name="foo",
+        implementation=HuggingFaceRuntime,
+        parameters=ModelParameters(extra={}),
+    )
+
+
+@pytest.mark.parametrize(
+    "model_settings,env_params,expected",
+    [
+        (
+            "model_settings_extra_task",
+            {"task": "question-answering"},
+            {"task": "question-answering", "pretrained_model": "distilgpt2"},
+        ),
+        (
+            "model_settings_extra_task",
+            {},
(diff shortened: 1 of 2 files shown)


runtimes/huggingface/tests/test_settings.py
```diff
@@ -0,0 +1,145 @@
...
+            {"task": "question-answering"},
+            {"task": "question-answering"},
+        ),
+    ],
+)
+def test_merge_huggingface_settings_extra(
+    model_settings: ModelSettings, env_params: Dict, expected: Dict, request: Any
+):
+    assert expected == merge_huggingface_settings_extra(
+        request.getfixturevalue(model_settings), env_params
```
adriangonz, line 70: Interesting! I didn't know `request.getfixturevalue` was a thing!

adriangonz (approved):
Nice one! From my side, changes look great :+1:

runtimes/huggingface/tests/test_settings.py
jesse-c: I was having issues with `lazy_fixture` so I went with this alternative! 😅

jesse-c:
Tested manually in clusters locally and in GKE with SCv1 and an image built by myself and pushed to Docker Hub (`docker.io/jsldn/mlserver:1.4.0.dev3-huggingface`).

I used a `model-settings.json` with a `task: text-generation` and had 2 deployments, 1 with no overwrite and 1 with overwrote with an env var for a `task: question-answering`. I used Seldon Deploy to run predictions against them both and received the expected results.

cc: @agrski, @adriangonz

vtaskow (approved):
Great work! Nothing more to say

PR #1346, opened 2023-08-24
Document and test behaviour around NaN

### LINKED ISSUE #1154: Support for NaN in np.ndarray codec?
Sending/receiving a `numpy` array containing NaN values to/from an mlserver with the `NumpyCodec` results in a `ValueError: Out of range float values are not JSON compliant` error. Is this a known limitation, and are there any good workarounds that would still allow me to use a pre-packaged server? I understand that it would probably be bad to assume than `NaN` can be mapped to `null` in the JSON, but maybe that could be an option?

adriangonz:
Fixes #1154

## Changelog

- Add docs around NaN behaviour
- Extend Numpy and Pandas codecs to handle NaNs (only for `FP*` datatypes)

### DIFF
--- a/docs/user-guide/content-type.md
+++ b/docs/user-guide/content-type.md
@@ -202,6 +202,56 @@ raw_response = response.json()
 inference_response = InferenceResponse(**raw_response)
 ```
 
+#### Support for NaN values
+
+The NaN (Not a Number) value is used in Numpy and other scientific libraries to
+describe an invalid or missing value (e.g. a division by zero).
+In some scenarios, it may be desirable to let your models receive and / or
+output NaN values (e.g. these can be useful sometimes with GBTs, like XGBoost
+models).
+This is why MLServer supports encoding NaN values on your request / response
+payloads under some conditions.
+
+In order to send / receive NaN values, you must ensure that:
+
+- You are using the `REST` interface.
+- The input / output entry containing NaN values uses the `FP32` datatype.
+- You are either using the [Pandas codec](#pandas-dataframe) or the [Numpy
+  codec](#numpy-array).
+
+Assuming those conditions are satisfied, any `null` value within your tensor
+payload will be converted to NaN.
+
+For example, if you take the following Numpy array:
+
+```python
+import numpy as np
+
+foo = np.array([[1.2, 2.3], [np.NaN, 4.5]])
+```
+
+We could encode it as:
+
+```{code-block} json
+---
+emphasize-lines: 8
+---
+{
+  "inputs": [
+    {
+      "name": "foo",
+      "parameters": {
+        "content_type": "np"
+      },
+      "data": [1, 2, null, 4]
+      "datatype": "FP32",
+      "shape": [2, 2],
+    }
+  ]
+}
+```
+
+
 ### Model Metadata
 
 Content types can also be defined as part of the [model's
--- a/docs/user-guide/content-type.md
+++ b/docs/user-guide/content-type.md
@@ -215,7 +215,8 @@ payloads under some conditions.
 In order to send / receive NaN values, you must ensure that:
 
 - You are using the `REST` interface.
-- The input / output entry containing NaN values uses the `FP32` datatype.
+- The input / output entry containing NaN values uses either the `FP16`, `FP32`
+  or `FP64` datatypes.
 - You are either using the [Pandas codec](#pandas-dataframe) or the [Numpy
   codec](#numpy-array).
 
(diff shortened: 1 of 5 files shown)


mlserver/codecs/numpy.py
```diff
@@ -92,7 +92,21 @@ def _encode_data(data: np.ndarray, datatype: str) -> list:
...
+
+    # Replace NaN with null
+    has_nan = np.isnan(data).any()
+    if has_nan:
+        flattened_list = list(map(convert_nan, flattened_list))
+
+    return flattened_list
+
+
+def convert_nan(val):
```
vtaskow, line 108: Given that this is used in both pandas and numpy codecs, does it make sense to be move out to the common codecs/util.py file :thinking:

vtaskow: I can see that in the pandas codec files from numpy codec are being imported. Is this correct?

tests/codecs/test_numpy.py
```diff
@@ -110,13 +120,59 @@ def test_encode_output(payload: np.ndarray, expected: ResponseOutput):
             RequestInput(name="foo", shape=[2], data=["foo", "bar"], datatype="BYTES"),
             np.array(["foo", "bar"], dtype=str),
         ),
+        (
+            RequestInput(name="foo", shape=[3], data=[1, 2, None], datatype="FP16"),
+            np.array([1, 2, np.NaN], dtype="float16"),
```
vtaskow, line 135: nit: perhaps using float values here is a bit more appropriate because the dtype is float :thinking: just to make it more obvious

docs/user-guide/content-type.md
```diff
@@ -202,6 +202,57 @@ raw_response = response.json()
...
+emphasize-lines: 8
+---
+{
+  "inputs": [
+    {
+      "name": "foo",
+      "parameters": {
+        "content_type": "np"
+      },
+      "data": [1, 2, null, 4]
```
vtaskow, line 247: Shouldn't those be float numbers

tests/codecs/test_numpy.py
```diff
@@ -110,13 +120,59 @@ def test_encode_output(payload: np.ndarray, expected: ResponseOutput):
...
+        ),
     ],
 )
 def test_decode_input(request_input: RequestInput, expected: np.ndarray):
     decoded = NumpyCodec.decode_input(request_input)
     np.testing.assert_array_equal(decoded, expected)
 
 
+@pytest.mark.parametrize(
```
vtaskow: I am a bit lost about this test compared to the one above :eye: In this one we are ensuring the resulted dtype is float and that the dtype for bytes is removed? Also is the name of the test correct?

tests/codecs/test_pandas.py
```diff
@@ -70,6 +70,16 @@ def test_can_encode(payload: Any, expected: bool):
                 name="bar", shape=[2, 1], data=[[1, 2, 3], [4, 5, 6]], datatype="BYTES"
             ),
         ),
+        (
+            pd.Series(data=[4, np.NaN, 6], name="bar"),
```
vtaskow, line 74: nit: using floats instead of int to better illustrate the idea

tests/codecs/test_pandas.py
```diff
@@ -134,6 +144,23 @@ def test_to_response_output(series, use_bytes, expected):
                 ],
             ),
         ),
+        (
+            pd.DataFrame(
+                {
+                    "a": [1, np.NaN, 3],
```
vtaskow, line 150: nit: same

vtaskow (changes requested):
I am not too sure about some aspects. Other comments are just nits.

mlserver/codecs/numpy.py
adriangonz: It's quite specific to pandas / numpy, so I'd leave it as-is

tests/codecs/test_numpy.py
adriangonz: Well, it's also a good test that what drives the type is the `datatype` field and not the format in which the values come (i.e. `2` vs `2.0`).

docs/user-guide/content-type.md
adriangonz: In JSON both are just "Numbers" - there's no difference between integers / floats / etc.

tests/codecs/test_numpy.py
adriangonz: Good spot - I think I duplicated one test by mistake. It should be fixed in the latest version.

tests/codecs/test_numpy.py
vtaskow: You are right now that I think of it 👍

vtaskow (approved):
Your comments make sense. 👍

PR #1402, opened 2023-09-20
Fix minor typo in `sklearn` README

krishanbhasin-gc:
Fixing a very minor typo (`availabe` -> `available`) in the `sklearn` docs


agrski (approved):
Hi @krishanbhasin-gc, this all looks good.

While you're here, would you mind using a tool like [`typos`](https://github.com/crate-ci/typos) to catch any other minor ones?  No worries if not, but would be helpful if you have a moment!

krishanbhasin-gc:
It looks like quite a few pop up, with many false-positives. If it's okay I'd rather not comb through them right now, but maybe if I get a chance I'll take a look in the future.

PR #1403, opened 2023-09-20
Add catboost classifier support

krishanbhasin-gc:
Reviving https://github.com/SeldonIO/MLServer/pull/529.

From the original PR description
> add catboost support:
>
>  - add a new runtime by heavily copying lightgbm runtime
>   - add test that builds a model, sends an inference request, validates the prediction
>   - add example in docs
>   - list in index/readme

I have only resolved merge conflicts and done a first pass of reshaping some of the configuration files to match the new poetry approach.

-----------
Looking ahead to adding support for the Regressor and Ranker models, given https://github.com/catboost/catboost/issues/2504 we can't use the same approach as XGBoost (using a try-catch: https://github.com/SeldonIO/MLServer/blob/d41f24e8d4e2e45afad8998331bada9edb9cf555/runtimes/xgboost/mlserver_xgboost/xgboost.py#L24-L34).

It might be worth creating 3 separate `MLModel` child classes, one for each of Classifier, Regressor, and Ranker instead. Could this fit the design approach of the project?


adriangonz (approved):
Hey @krishanbhasin-gc , 

Apologies for the delay reviewing this one.

This is great work! Thanks a lot for jumping on top of this one! :rocket: From my point of view, this looks great as-is. Once the tests are all good, we should be able to merge this one. 

On what you mention around classifiers vs regressors, I think we should land this one as a first iteration and then open an issue to add support for extra types of CatBoost models. It's a shame the framework itself doesn't give us any way to differentiate between artefacts - in that case, we may need to add some extra config field to the `model-settings.json` to select between model types.

adriangonz:
Actually, tests are already all green - so we should be good to go with this one. Great work @krishanbhasin-gc ! :+1:

Issue #1344, opened 2023-08-23
Allow huggingface model's from_pretrained kwargs in the model-setting.json?


nanbo-liu:
```python
from transformers import LlamaForCausalLM, AutoTokenizer, TextGenerationPipeline

model = LlamaForCausalLM.from_pretrained("daryl149/llama-2-7b-hf",load_in_8bit=True)
tokenizer = AutoTokenizer.from_pretrained("daryl149/llama-2-7b-hf")
pipeline = TextGenerationPipeline(model, tokenizer)
pipeline("Once upon a time,", max_new_tokens=100)
```
Recent popular LLM like llama-2 takes tons of memory. However, with quantization its memory requirement greatly reduce.
Huggingface allow me to convert llama-2 into quantized 8bit version by simply add an keyword argument ` load_in_8bit=True`
Whish we can specify from_pretrained_kwargs in the model-setting.json
example model-setting.json :
```json
{
    "name": "llama_mlserver",
    "implementation": "mlserver_huggingface.HuggingFaceRuntime",
    "parameters": {
        "extra": {
            "task": "text-generation",
            "pretrained_model": "daryl149/llama-2-7b-hf",
            "model_kwargs":{"load_in_8bit":true}
        }
    }
}

```

nanbo-liu:
Hi, @adriangonz. I just created a PR for this issue.
https://github.com/SeldonIO/MLServer/pull/1417

PR #1417, opened 2023-09-27
added model_kwargs to huggingface model

nanbo-liu:
This is for issue [1344](https://github.com/SeldonIO/MLServer/issues/1344)
This allow `model-settings.json` to take `model_kwargs`:
example below:
```json
{
    "name": "llama",
    "parameters": {
        "extra": {
            "task": "text-generation",
            "pretrained_model": "daryl149/llama-2-7b-hf",
            "model_kwargs": {
                "device_map": "auto",
                "load_in_8bit": true
            }
        }
    }
}

```

### DIFF
--- a/runtimes/huggingface/mlserver_huggingface/common.py
+++ b/runtimes/huggingface/mlserver_huggingface/common.py
@@ -25,7 +25,6 @@ def load_pipeline_from_settings(
     hf_settings: HuggingFaceSettings, settings: ModelSettings
 ) -> Pipeline:
     pipeline = _get_pipeline_class(hf_settings)
-
     batch_size = 1
     if settings.max_batch_size:
         batch_size = settings.max_batch_size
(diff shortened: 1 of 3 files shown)


adriangonz (changes requested):
Nice one @nanbo-liu! This would be a great addition to the HF runtime - thanks for contributing it :rocket: 

Changes look good - although, could you add some tests to validate that the model kwargs get used? Once tests are in, we should be good to land this one. :+1:

nanbo-liu:
@adriangonz ,added unit tests from the latest commit.

adriangonz (approved):
Brilliant! Thanks for the changes @nanbo-liu ! 

This should be good to go now :rocket:

nanbo-liu:
@adriangonz ,I rewrite the unit test a little bit. The previous version depends on package `accelerate`.
I don't want to introduce extra dependency into this repo. Would you take an another look when you have a chance.
Need to rerun workflow

nanbo-liu:
@adriangonz, just fixed the unit test bug. would you try again

nanbo-liu:
@adriangonz ,CICD keeps getting runtime error for `catboost`

adriangonz:
Yuuup, it seems there's an unrelated error in the `catboost` runtime causing some actions to fail. Which means this PR should be good to go as-is :+1:

PR #1438, opened 2023-10-11
Update opentelemetry dependency requirements

wdhorton:
The poetry caret requirements operate differently for version 0.x than for 1.x and above. For ^0.2.0, for example, it would update to 0.2.1 but not 0.3.0. Whereas ^1.2.0 would update to 1.3.0 if it existed.

This has the effect of pinning these opentelemetry deps to a specific version, which I'm not sure was intended, and creates problems for projects that depend on mlserver and also use opentelemetry. Updating to `>=` instead of `^` should still work and offer more flexibility


adriangonz:
Hey @wdhorton , 

Thanks for the contribution!

Before going ahead, could you also update the lock file?

adriangonz:
Hey @wdhorton ,

Closing this one for now, although we're still quite keen to get this merged. Feel free to reopen once you get a chance to address the comments above.

Issue #1023, opened 2023-03-05
Support of caching


saeid93:
This is a feature request for support of caching for hashable data types as implemented in academic serving systems [Clipper: A Low-Latency Online Prediction Serving System](https://www.usenix.org/system/files/conference/nsdi17/nsdi17-crankshaw.pdf) section 4.2 and also similar frameworks like [Triton inference server](https://github.com/triton-inference-server/server/blob/main/docs/user_guide/response_cache.md). This could be very useful for non custom runtimes like HuggingFace runtime.

adriangonz:
Fixed by https://github.com/SeldonIO/MLServer/pull/1440

PR #1440, opened 2023-10-15
Local response cache implementation

SachinVarghese:
This PR introduces basic local response caching as per the issue https://github.com/SeldonIO/MLServer/issues/1023
The implementation is inspired by the triton equivalent local cache https://github.com/triton-inference-server/local_cache

### DIFF
--- a/mlserver/handlers/dataplane.py
+++ b/mlserver/handlers/dataplane.py
@@ -31,7 +31,7 @@ class DataPlane:
         self,
         settings: Settings,
         model_registry: MultiModelRegistry,
-        response_cache: ResponseCache | None,
+        response_cache: Optional[ResponseCache] = None,
     ):
         self._settings = settings
         self._model_registry = model_registry
--- a/mlserver/handlers/dataplane.py
+++ b/mlserver/handlers/dataplane.py
@@ -29,7 +29,7 @@ class DataPlane:
     def __init__(self, settings: Settings, model_registry: MultiModelRegistry):
         self._settings = settings
         self._model_registry = model_registry
-
+        self._response_cache = None
         if settings.cache_enabled:
             self._response_cache = self._create_response_cache()
         self._inference_middleware = InferenceMiddlewares(
@@ -132,3 +132,6 @@ class DataPlane:
 
     def _create_response_cache(self) -> ResponseCache:
         return LocalCache(size=self._settings.cache_size)
+
+    def _get_response_cache(self) -> Optional[ResponseCache]:
+        return self._response_cache
--- a/mlserver/handlers/dataplane.py
+++ b/mlserver/handlers/dataplane.py
@@ -4,7 +4,6 @@ from prometheus_client import (
 )
 from typing import Optional
 
-import asyncio
 from ..errors import ModelNotReady
 from ..context import model_context
 from ..settings import Settings
@@ -18,7 +17,7 @@ from ..types import (
 from ..middleware import InferenceMiddlewares
 from ..cloudevents import CloudEventsMiddleware
 from ..utils import generate_uuid
-from ..cache import ResponseCache
+from ..cache import ResponseCache, LocalCache
 
 
 class DataPlane:
@@ -27,15 +26,11 @@ class DataPlane:
     servers.
     """
 
-    def __init__(
-        self,
-        settings: Settings,
-        model_registry: MultiModelRegistry,
-        response_cache: Optional[ResponseCache] = None,
-    ):
+    def __init__(self, settings: Settings, model_registry: MultiModelRegistry):
         self._settings = settings
         self._model_registry = model_registry
-        self._response_cache = response_cache
+
+        self._response_cache = self._create_response_cache()
         self._inference_middleware = InferenceMiddlewares(
             CloudEventsMiddleware(settings)
         )
@@ -111,7 +106,7 @@ class DataPlane:
             with model_context(model.settings):
                 if (
                     self._settings.cache_enabled
-                    and model.settings.cache_enabled
+                    and model.settings.cache_enabled is not False
                     and self._response_cache is not None
                 ):
                     cache_value = await self._response_cache.lookup(cache_key)
@@ -119,9 +114,8 @@ class DataPlane:
                         prediction = InferenceResponse.parse_raw(cache_value)
                     else:
                         prediction = await model.predict(payload)
-                        asyncio.create_task(
-                            self._response_cache.insert(cache_key, prediction.json())
-                        )
+                        # ignore cache insertion error if any
+                        self._response_cache.insert(cache_key, prediction.json())
                 else:
                     prediction = await model.predict(payload)
 
@@ -133,3 +127,11 @@ class DataPlane:
             self._ModelInferRequestSuccess.labels(model=name, version=version).inc()
 
             return prediction
+
+    def _create_response_cache(self) -> Optional[ResponseCache]:
+        if self._settings.cache_enabled:
+            if self._settings.cache_size is None:
+                # Default cache size if caching is enabled
+                self._settings.cache_size = 100
+            return LocalCache(size=self._settings.cache_size)
+        return None
--- a/mlserver/handlers/dataplane.py
+++ b/mlserver/handlers/dataplane.py
@@ -31,7 +31,7 @@ class DataPlane:
         self,
         settings: Settings,
         model_registry: MultiModelRegistry,
-        response_cache: ResponseCache,
+        response_cache: ResponseCache | None,
     ):
         self._settings = settings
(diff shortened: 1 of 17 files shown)


SachinVarghese:
PTAL @adriangonz :slightly_smiling_face:

mlserver/handlers/dataplane.py
```diff
@@ -25,10 +27,15 @@ class DataPlane:
...
+    def __init__(
+        self,
+        settings: Settings,
+        model_registry: MultiModelRegistry,
+        response_cache: Optional[ResponseCache] = None,
+    ):
         self._settings = settings
         self._model_registry = model_registry
-
+        self._response_cache = response_cache
```
adriangonz, line 38: Given that `response_cache` doesn't need to be shared outside of the `DataPlane` class, should we just create it here from `settings` instead of passing it as an arg?

mlserver/handlers/dataplane.py
```diff
@@ -90,6 +97,7 @@ async def infer(
         ).count_exceptions()
 
         with infer_duration, infer_errors:
+            cache_key = payload.json()
```
adriangonz, line 100: Depending on the request size, this could be quite expensive - would it make sense to only run it when caching enabled?

Alternatively, would it make sense to use the request ID if present?

mlserver/server.py
```diff
@@ -56,6 +60,12 @@ def __init__(self, settings: Settings):
         self._configure_logger()
         self._create_servers()
 
+    def _create_response_cache(self) -> Optional[ResponseCache]:
+        if self._settings.cache_enabled:
+            return LocalCache(size=self._settings.cache_size)
+        else:
+            return None
```
adriangonz, line 67: nit: I personally find `if / return` blocks more readable than `if / else`, e.g. :

```python
def _create_response_cache(self) -> Optional[ResponseCache]:
	if self._settings.cache_enabled:
	    return LocalCache(size=self._settings.cache_size)
	
	return None
```

But this is a completely personal opinion, so feel free to ignore it.

mlserver/settings.py
```diff
@@ -393,3 +399,6 @@ def version(self) -> Optional[str]:
     # However, it's also possible to override them manually.
     parameters: Optional[ModelParameters] = None
     """Extra parameters for each instance of this model."""
+
+    cache_enabled: Optional[bool] = False
```
adriangonz, line 403: I think the logic here should be "whatever the global setting says, unless I override it at the model level". With that in mind, should we set the default to `None` (which would effectively mean "whatever `Settings.cache_enabled` says)?

We could also clarify the comment saying that this is meant to override the global setting and won't enable it if the global is disabled.

mlserver/settings.py
```diff
@@ -241,6 +241,12 @@ class Config:
     _custom_metrics_server_settings: Optional[dict] = None
     _custom_grpc_server_settings: Optional[dict] = None
 
+    cache_enabled: Optional[bool] = False
+    """Enable Caching for the model predictions."""
+
+    cache_size: Optional[int] = 100
+    """Cache size to be used if caching is enabled."""
```
adriangonz, line 248: Given these already get a default non-`None` value, should we remove the `Optional` annotation?

adriangonz (changes requested):
Nice one @SachinVarghese! It's great to see you back contributing some good stuff to MLServer!

I've added a few comments, but would be good to hear your thoughts on those. Besides that, it would also be great if you could add some tests around the caching behaviour.

mlserver/handlers/dataplane.py
SachinVarghese: Thanks for the review. This comment makes a lot of sense to me. I will update the PR soon.

SachinVarghese: Updated in [43dfc1c](https://github.com/SeldonIO/MLServer/pull/1440/commits/43dfc1c7587e738d377d437761fcb2bcade94854)

mlserver/server.py
SachinVarghese: Updated in [43dfc1c](https://github.com/SeldonIO/MLServer/pull/1440/commits/43dfc1c7587e738d377d437761fcb2bcade94854)

mlserver/settings.py
SachinVarghese: Updated in [43dfc1c](https://github.com/SeldonIO/MLServer/pull/1440/commits/43dfc1c7587e738d377d437761fcb2bcade94854)

mlserver/settings.py
SachinVarghese: Updated in [43dfc1c](https://github.com/SeldonIO/MLServer/pull/1440/commits/43dfc1c7587e738d377d437761fcb2bcade94854)

mlserver/handlers/dataplane.py
SachinVarghese: So the idea here is that if the cache key is independent of a request ID and other headers, we can expect more cache hits. No two requests should have the same ID so then we don't benefit from caching. I understand this is expensive, and a different choice of cache key would probably make sense. Let me know your thoughts.

mlserver/handlers/dataplane.py
```diff
@@ -101,7 +109,21 @@ async def infer(
...
+                    and model.settings.cache_enabled
+                    and self._response_cache is not None
+                ):
+                    cache_value = await self._response_cache.lookup(cache_key)
+                    if cache_value != "":
+                        prediction = InferenceResponse.parse_raw(cache_value)
+                    else:
+                        prediction = await model.predict(payload)
+                        asyncio.create_task(
+                            self._response_cache.insert(cache_key, prediction.json())
```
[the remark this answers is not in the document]
SachinVarghese: Updated in [43dfc1c](https://github.com/SeldonIO/MLServer/pull/1440/commits/43dfc1c7587e738d377d437761fcb2bcade94854)
Ignoring the insertion error now

mlserver/handlers/dataplane.py
adriangonz: Ah, I see what you mean - that makes complete sense.

In that case, it may be worth only computing the `payload.json()` key when the response cache is enabled.

mlserver/settings.py
```diff
@@ -241,6 +241,12 @@ class Config:
     _custom_metrics_server_settings: Optional[dict] = None
     _custom_grpc_server_settings: Optional[dict] = None
 
+    cache_enabled: Optional[bool] = None
+    """Enable caching for the model predictions."""
+
+    cache_size: Optional[int] = None
+    """Cache size to be used if caching is enabled."""
```
adriangonz, line 248: I was actually thinking the opposite here! In general, I find nullable types a bit of a pain sometimes (particularly when it comes to type checking - `mypy` complains a lot when you "forget" checking if something like `cache_size` is `None`). 

In this case, I think it would make sense to set them to a default value and make them non-nullable, i.e. something like

```python
  cache_enabled: bool = False
    """Enable caching for the model predictions."""

    cache_size: int = 100
    """Cache size to be used if caching is enabled."""
```

SachinVarghese:
I have made most of the suggested changes now, and all the CI tests should be passing. Please review.
Also, I'm not fully sure how to write the tests for something like this. Any ideas would be good.

mlserver/settings.py
SachinVarghese: Updated in [b30c41f](https://github.com/SeldonIO/MLServer/pull/1440/commits/b30c41f690076eb92d785e4268165ff7a36834aa)

mlserver/handlers/dataplane.py
SachinVarghese: Do you think it is a good idea to compress the cache key string?

adriangonz: Good idea - although I would keep it simple for now, and just use the JSON payloads you have now. We can always come back to it if we see memory spikes.

adriangonz:
I think for this one we probably want to add two types of tests:

- Unit tests for the cache itself, which could go under `tests/cache/...`. These can test the `LocalCache` implementation, i.e. does it rotate values correctly when it's full, what happens when key  is not present, does it save the values, etc.
- Integration tests for the `DataPlane` class, which can go under the `tests/handlers/...`. These can test that the `DataPlane` class create the cache correctly, picks up on cached values, caches new responses, ignores the cache when disabled, etc.

SachinVarghese:
I have added a few tests for the local cache and data plane component now, please review.

adriangonz (approved):
Nice one @SachinVarghese ! Thanks for adding those changes in.

This should be good to go from my side. :+1:

SachinVarghese:
Awesome and thanks for reviewing/ almost co-authoring :smile: Will wait for other reviews.

PR #1467, opened 2023-10-30
fix link to custom runtimes

kretes:


adriangonz (approved):
Good spot @kretes ! Thanks for fixing this one.

agrski (commented):
I don't think this PR fixes anything for how the docs are **meant** to be viewed -- browsing the [runtimes page](https://mlserver.readthedocs.io/en/stable/runtimes/index.html) and clicking the relevant link for custom runtimes works fine for me.

I suspect this is about the docs not linking correctly when browsing in GitHub, but that's a suboptimal experience in any case, as this is a Sphinx project rather than pure Markdown and various things don't render correctly.

With that sald, I'll check this doesn't break anything, as I think Sphinx is _generally_ permissive about `.md` vs. `.html` vs. `` endings for links, although sometimes it's more particular.

agrski (approved):
Running docs locally, this change looks to still work, so fine from my side too

PR #1469, opened 2023-11-01
Improve typing on `Environment` class

krishanbhasin-gc:
Per title; added return type of `Environment` to `__enter__()` as it was missing

### DIFF
--- a/mlserver/env.py
+++ b/mlserver/env.py
@@ -114,7 +114,7 @@ class Environment:
 
         return ""
 
-    def __enter__(self):
+    def __enter__(self) -> "Environment":
         self._prev_sys_path = sys.path
         self._prev_bin_path = os.environ["PATH"]
 


mlserver/env.py
```diff
@@ -114,7 +114,7 @@ def _lib_path(self) -> str:
 
         return ""
 
-    def __enter__(self):
+    def __enter__(self) -> "Environment":
```
agrski, line 117: :question: Shouldn't this be without the quotes, as we're referring to the type not a string?
```suggestion
    def __enter__(self) -> Environment:
```

jklaise: In Python these quoted types are called forward references which are used when you're referring to a type that is currently being defined (as is the case here), otherwise the type-checkers can't process that.

There is an alternative for this, the `typing.Self` type, unfortunately that was only introduced in 3.11. We could use the `Self` type from `typing_extensions` as it backports these new types but it's probably not worth it.

agrski: Thanks for explaining Janis - I hadn't seen this before.  I've just been reading the docs, e.g. [PEP-484](https://peps.python.org/pep-0484/#forward-references).

agrski (approved):

PR #1486, opened 2023-11-15
Fix mlserver_huggingface settings device type

geodavic:
If using a `model-settings.json` of the following form:
```json
{
    "name": "my-model",
    "parameters": {
        "extra": {
            "task": "text-generation",
            "pretrained_model": "model/path",
            "model_kwargs": {
                "load_in_8bit": true
            }
        }
    }
}
```

an error occurs when spinning up the server:
```
The model has been loaded with `accelerate` and therefore cannot be moved to a specific device. Please discard the `device` argument when creating your pipeline object.
```
because the `load_in_8bit` model kwarg makes forces the model to be loaded with Accelerate, and `HuggingFaceSettings.device` default value of -1 is passed. I tried simply passing `"device": null` to the model settings json to prevent this default value, but this wasn't accepted because the schema is typed to `int`.

The solution is to expand the typing of `HuggingFaceSettings.device` to accept null (and string, while we're at it). Therefore I have updated the typing to `Optional[Union[int,str]]`, which is very close to the type hint used in the underlying transformers [pipeline](https://github.com/huggingface/transformers/blob/main/src/transformers/pipelines/__init__.py#L542).

I have tested the above model settings json with this change and the server starts as expected.

### DIFF
--- a/runtimes/huggingface/mlserver_huggingface/settings.py
+++ b/runtimes/huggingface/mlserver_huggingface/settings.py
@@ -83,10 +83,10 @@ class HuggingFaceSettings(BaseSettings):
     runtime.
     """
 
-    device: Optional[Union[int, str]] = -1
+    device: Optional[Union[int, str]] = None
     """
     Device in which this pipeline will be loaded (e.g., "cpu", "cuda:1", "mps",
-    or a GPU ordinal rank like 1).
+    or a GPU ordinal rank like 1). Default value of None becomes cpu.
     """
 
     inter_op_threads: Optional[int] = None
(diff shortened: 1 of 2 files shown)


nanbo-liu:
@adriangonz

adriangonz (changes requested):
Looks great @geodavic! :rocket:  

Would it make sense adding a small test to make sure this behaviour doesn't change in the future? This could be as simple as loading a model with `device` set to empty, `-1`, or `cpu`? And just asserting that the model loads correctly.

geodavic:
@adriangonz Thanks! I added a test that passes `None`, `-1` and `cpu` for the device to ensure they all load into cpu.

geodavic:
@sakoush if you don't mind taking a look, much appreciated!

jyono:
@adriangonz @sakoush 

is there anything holding this back?

sakoush (approved):
Lgtm. Many thanks for your contribution.

PR #1511, opened 2023-12-12
Add Japanese language dependencies

### LINKED ISSUE #1506: Install `fugashi`, `unidic`, `unidic-lite`, and `ipadic` as dependencies to MLServer HuggingFace to support hosting Japanese language models
Because of the fact that Japanese mixes phonetic scripts and Chinese characters, special algorithms and dictionaries are needed to run tokenizers for these these models. A popular example of this is the BERT Japanese model:

https://huggingface.co/transformers/v4.11.3/_modules/transformers/models/bert_japanese/tokenization_bert_japanese.html

Without these dependencies, [mlserver_huggingface/common.py](https://github.com/SeldonIO/MLServer/blob/master/runtimes/huggingface/mlserver_huggingface/common.py) errors when trying to load the tokenizer in the pipeline. 

To reproduce, use any Japanese model. [Here](https://huggingface.co/cl-tohoku/bert-base-japanese) is an example.

jbauer2718:
Resolves #1506


runtimes/huggingface/pyproject.toml
```diff
@@ -13,6 +13,10 @@ mlserver = "*"
 tensorflow = "*"
 pillow = "*"
 optimum = {extras = ["onnxruntime"], version = ">=1.4,<2.0"}
+fugashi = "^1.3.0"
+unidic = "^1.1.0"
+unidic-lite = "^1.0.8"
+ipadic = "^1.0.0"
```
sakoush, line 19: Could we have these libraries as optional?

runtimes/huggingface/tests/test_tasks/test_japanese_mask_fill.py
```diff
@@ -0,0 +1,48 @@
+import pytest
+from mlserver.settings import ModelSettings, ModelParameters
+from mlserver_huggingface import HuggingFaceRuntime
+from mlserver_huggingface.codecs import HuggingfaceRequestCodec
+
+
+test_sentence = "実際に空が[MASK]のか？"
```
sakoush, line 7: Do we need the extra space in the prompt?
Could we document in English the translation?

runtimes/huggingface/tests/test_tasks/test_japanese_mask_fill.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+@pytest.fixture
+def current_model_settings(request) -> ModelSettings:
```
sakoush, line 11: nit: rename fixture to be a bit more descriptive of the fact that the tokenizer is different?

runtimes/huggingface/tests/test_tasks/test_japanese_mask_fill.py
sakoush: Also move it as a test case input and the associated expected output.

runtimes/huggingface/tests/test_tasks/test_japanese_mask_fill.py
```diff
@@ -0,0 +1,48 @@
...
+        use_bytes=False,
+    )
+    return req
+
+
+async def test_infer(current_runtime, japanese_text_request):
+    resp = await current_runtime.predict(japanese_text_request)
+    decoded = HuggingfaceRequestCodec.decode_response(resp)
+    top_result = list(decoded.items())[0][1][0]
+    assert top_result["token_str"] == "見える", "Inference is incorrect!"
```
sakoush, line 48: Inference is correct?

sakoush (changes requested):
Many thanks for your contribution, I left some comments. I suggest that these extra libraries are optional as they are not required for all cases. Also I left some suggestions to tidy up the tests.

runtimes/huggingface/tests/test_tasks/test_japanese_mask_fill.py
jbauer2718: Hey @sakoush , thanks for the feedback. Your suggestions make sense; I'm working on implementing them in the next day or two. One question for clarification: not sure I am seeing an extra space in the prompt. Are you talking about the visual gap between か and ？?

jbauer2718:
@sakoush I believe I have addressed your comments. Let me know if I have misunderstood something!

runtimes/huggingface/tests/test_tasks/test_japanese_mask_fill.py
```diff
@@ -0,0 +1,53 @@
...
+
+    inputs = [test_sentence]
+    req = HuggingfaceRequestCodec.encode_request(
+        {"inputs": inputs},
+        use_bytes=False,
+    )
+    return req
+
+
+async def test_infer(current_runtime, japanese_text_request):
```
sakoush, line 47: I think this test function should be parametrised with test cases [(input_request, output_request),...] even though you just have one at the moment to allow for future extensions.

sakoush (changes requested):
Thanks for acting on the suggestions, I left a minor suggestion to parametrise the test case.

Also could you please update the huggingface runtime docs with to explain how to use your addition?

runtimes/huggingface/tests/test_tasks/test_japanese_mask_fill.py
jbauer2718: Hey @sakoush , I think all of your comments are addressed. I appreciate a review when you get the chance. Enjoy your holidays!

docs/examples/huggingface/README.md
```diff
@@ -199,6 +199,46 @@ inference_request = {
 requests.post("http://localhost:8080/v2/models/transformer/infer", json=inference_request).json()
 ```
 
+### Masked Language Modeling
```
sakoush, line 202: What's the connection between masked language models and Japanese support?

runtimes/huggingface/tests/test_tasks/test_japanese_mask_fill.py
```diff
@@ -0,0 +1,56 @@
...
+    ):
+        pytest.skip()
+
+    return ModelSettings(
+        name="model",
+        implementation=HuggingFaceRuntime,
+        parameters=ModelParameters(
+            extra={
+                "task": "fill-mask",
+                "pretrained_model": "cl-tohoku/bert-base-japanese",
```
sakoush, line 20: How big is this model, if big is there a smaller model that can be used for testing purposes?

runtimes/huggingface/tests/test_tasks/test_japanese_mask_fill.py
```diff
@@ -0,0 +1,56 @@
...
+
+    outputs = [expected_output]
+    reqs = HuggingfaceRequestCodec.encode_request(
+        {"inputs": [test_sentence]},
+        use_bytes=False,
+    )
+    return reqs, outputs
+
+
+async def test_infer(current_runtime, japanese_text_request):
```
sakoush, line 49: I think that this test function should include a parametrised (ie `pytest.mark.parametrize`) list of test cases in the form of [(`input_text`, `expected_response`),...] instead of hardcoding things  in the `japanese_test_request` fixture as it is not readable currently (ie separate out the expected response for it)

sakoush (changes requested):
Thanks for the contribution so far, I left a few minor suggestions.

Apologies for the slow response due to the holiday season.

docs/examples/huggingface/README.md
jbauer2718: There isn't any explicit connection. Apologies if I misinterpreted your earlier comment. I was showcasing this support by example like the other tasks in this section. If there's somewhere else we should explicitly document language support, I'd be happy to add something there.

runtimes/huggingface/tests/test_tasks/test_japanese_mask_fill.py
jbauer2718: I could switch it out for https://huggingface.co/izumi-lab/bert-small-japanese/tree/main which is a smaller model. It's about 70 MB.

runtimes/huggingface/tests/test_tasks/test_japanese_mask_fill.py
jbauer2718: Yeah that makes sense. I'll make this change.

jbauer2718: @sakoush I refactored the test to not make use of fixtures so we can use the `pytest.mark.parameterize` decorator. I have also moved the model config to a parameter. The single test case is documented above where we were using inline comments to describe the test case before. I believe this addresses your comment.

docs/examples/huggingface/README.md
jbauer2718: Also just for clarification, the language support doesn't require anything other than installing the optional dependencies. We're using a mask fill example to showcase this because it's a common way language models are pretrained. The point of the test is that some model (in the case a text fill model) can tokenize and make an inference on Japanese text.

jbauer2718:
@sakoush Hey! Hope your holidays went well. Just bumping this to your attention because I think it's ready if not close to ready to merge. Let me know if I misunderstood some of your comments. Looking forward to your feedback.

docs/examples/huggingface/README.md
sakoush: Maybe the heading should reflect that it is an example for using an optional library that is required for tokenizers?

docs/examples/huggingface/README.md
```diff
@@ -199,6 +199,46 @@ inference_request = {
...
+
+We can also serve a masked language model. In the following example, we also build the `huggingface` runtime with the `-E japanese` flag to enable support for Japanese tokenizers.
```
sakoush, line 204: maybe provide a cmd how to build exactly?

sakoush (commented):
Many thanks for the contribution, I left a minor doc change.

docs/examples/huggingface/README.md
jbauer2718: Hey @sakoush, provided a detailed command in my last commit and updated the header. Let me know if anything else needs to change!

jyono (approved):

jbauer2718:
Hey @sakoush , the pipeline is failing on TF v2.14 dropping support for Python 3.8. Would you rather drop support for Python 3.8 for this runtime or have me deprecate TF to v2.13?

jbauer2718:
@sakoush Looked into this a bit more. I am pretty sure we're going to have to bump to 3.9 unless we deprecate tensorflow-io too. In my personal opinion, I think it just makes sense to go this route rather than freeze all of our tensorflow dependencies forever. Let me know what you think.

sakoush:
@jbauer2718 could you please rebase your changes on top of latest master as they are now merge conflicts.

jbauer2718:
@sakoush I synced with the main branch. Looks like some packages were updated in the runtime lockfiles. Let me know if you want me to explicitly revert these, but should be GTG.

sakoush:
@jbauer2718 it looks like that merge conflicts have not been resolved correctly, at least by looking at the changes in `docs/examples/huggingface/README.md`

jbauer2718:
@sakoush Sorry for missing that. Should be good now.

PR #1591, opened 2024-02-28
 docs: List supported Python versions 

jesse-c:
> Python 3.8 is due to reach its scheduled upstream end-of-life in October 2024.

In preparation for this, we're marking support for this version as deprecated. It's still available, but will be removed in a future version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,25 @@ Out of the box, MLServer provides support for:
 | Alibi-Explain | ✅        | [MLServer Alibi Explain](./runtimes/alibi-explain)               |
 | HuggingFace   | ✅        | [MLServer HuggingFace](./runtimes/huggingface)                   |
 
+## Supported Python Versions
+
+🔴 Unsupported
+
+🟠 Deprecated: To be removed in a future version
+
+🟢 Supported
+
+🔵 Untested
+
+| Python Version | Status |
+| -------------- | ------ |
+| 3.7            | 🔴     |
+| 3.8            | 🔴     |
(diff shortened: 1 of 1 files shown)


sakoush (commented):
We should probably establish whether mlserver supports py11 and py12

jesse-c:
@sakoush: I was thinking about that. I did some brief testing, and had some dependency installation failures on Python `3.11.8`. Happy to update it to say that it's "untested" or something like that.

wdyt?

sakoush:
> @sakoush: I was thinking about that. I did some brief testing, and had some dependency installation failures on Python `3.11.8`. Happy to update it to say that it's "untested" or something like that.
> 
> wdyt?

yes we need to reflect status and perhaps also add an issue to investigate why mlserver doesnt work on py3.11.

RafalSkolasinski:
As far as I understand https://github.com/SeldonIO/MLServer/pull/1603 removed support for Python 3.8. 
Because of 
https://github.com/SeldonIO/MLServer/blob/775fa4f9d238f2e41c2aec282b25d5c819bfa09a/pyproject.toml#L46
it will not be able to use it with Python 3.8 so it is not deprecated but removed

jesse-c:
@RafalSkolasinski: This PR was opened before that, so I'll update this now.

jesse-c:
@RafalSkolasinski: Done! Thank you for the reminder. 🙏🏻

sakoush (approved):

PR #1602, opened 2024-03-01
build: Bump version and improve release doc

jesse-c:
Bumps the version to match the latest release and improves the release doc, to mention this.

FIXES INFRA-838


sakoush (commented):
I think the pattern is to upgrade the version to the next dev version, so `1.5.0.dev1`  in the master branch. For the official releases (on the release branches), the CI will tag the version on them properly.

jesse-c:
Ah, that makes sense. Since it may be `1.4.1.dev` too, is it best to leave it as is?

jesse-c:
Changed to `1.5.0.dev1`. Thank you!

sakoush (approved):
lgtm

PR #1651, opened 2024-03-27
Pydantic v2 support

privatedumbo:
Making `MLServer` compatible with `pydantic v1` **AND** `pydantic v2`.

Things added in this PR:
- removing `pydantic` dependency from runtimes when not required.
- redefining classes so that they use an interface compatible with versions 1 and 2 of pydantic.
- How to use with `pydantic v1`: install MLServer as usual, but define in the calling project an upper boundary for `pydantic`.
- How to use with `pydantic v2`: install MLServer with `pip install MLServer[pydantic-v2]` as an extra.


mlserver/pydantic_migration.py
```diff
@@ -0,0 +1,19 @@
...
+if TYPE_CHECKING:
+    from pydantic.types import ImportString as PyObject
+    from pydantic_settings import BaseSettings
+else:
+    if is_pydantic_v1():
+        from pydantic import PyObject
+        from pydantic import BaseSettings
+    else:
+        from pydantic.types import ImportString as PyObject
+        from pydantic_settings import BaseSettings
```
privatedumbo, line 19: This seems ugly, but it seems to be the only way to trick MyPy (link to a related issue [here](https://github.com/pydantic/pydantic/issues/6022#issuecomment-1671630257)).

How this goes:
1) If `TYPE_CHECKING`, import the objects from the pydantic version installed in the environment. As this is the import that runs first, `mypy` will use these imported objects when linting.
2) `else` imports as used during runtime.

mlserver/parallel/messages.py
```diff
@@ -26,13 +26,10 @@ class ModelRequestMessage(Message):
...
 
 
-class ModelResponseMessage(Message):
-    class Config:
-        # This is to allow having an Exception field
-        arbitrary_types_allowed = True
-
-    return_value: Optional[Any]
-    exception: Optional[Union[Exception, CancelledError]]
+# `arbitrary_types_allowed` to allow having an Exception field
```
jesse-c, line 29: I appreciate this comment.

jesse-c (commented):
Thank you for this PR, @privatedumbo! Overall, the change to support both v1 and v2, as decided by the user makes sense. It gives a nice way to drop v1 support.

I've left some minor comments. Could you please do a rebase to get the latest changes around dependency updates? I'll approve workflows being able to be run.

Would you be able to add some tests for the v2 path?

sakoush:
@privatedumbo @franco-bocci many thanks for your contribution to provide support for `pydantic` v2 on MLServer. This is a  something that is high on our priority list at the moment.

We have been discussing your change internally and we think that it might be simpler to move directly to `pydantic` v2 and drop support for `pydantic` v1 in the next release of MLServer. Our reasoning is:
- `pydantic` v1 EOL is June 30, 2024 as described [here](https://docs.pydantic.dev/latest/version-policy/#pydantic-v1)
- Given that `pydantic` EOL is very soon, we think that it will be a simpler PR and for ongoing management work from our side to move directly for `pydantic` v2 and drop support for v1, as otherwise we have to:
    - Add docs to explain how users can still use both versions for cases such as `pip`, `poetry` and docker images.
    - Increase test coverage to include `pydantic` v1 and v2 tests.
    - Remove pydantic v1 support at some point in the future.
    

Alternatively the plan is to release MLServer `1.6.0` with **just** `pydantic` v2 support. You can keep using MLServer `1.5.x` that supports `pydanic` v1 for any downstream projects that still require it. We hope that this is possible in your case. Please let us know.



cc: @jesse-c @ramonpzg

jesse-c:
**follow-up:** We've opened https://github.com/SeldonIO/MLServer/pull/1748 to supplant this. We really do appreciate the work in this PR as a step towards that work being possible.

jesse-c:
We've merged in Pydantic v2 support. Again, thank you for your proposed solution and the time you spent on it! 🙇🏻

PR #1677, opened 2024-04-11
fix: Missing f-string directives

jesse-c:
This is a regression from presumably the line being split up.


mauicv (approved):
LGTM! Good catch!

Issue #1461, opened 2023-10-27
Python version mismatch after loading a Conda env


GDegrove:
Hello, 

I'm trying to run a MLFlow saved pyfunc model. This model is fairly simple, it's a linear regression with a custom identity transformer for the pipeline. The idea is to demonstrate how to load custom code at runtime. 

The upload/download of the artifact works perfectly. 

Now comes the weird part. 

My model run with python 3.11.2, this version is saved by MLFlow as conda environment. I packed the conda env using `conda-pack` and it's being uploaded as an artifact.

This artifact is downloaded with the model, and I see that the conda env is being loaded by MLServer. 

In the logs of the mlserver docker container, I see: 

```
--> Unpacking environment at /mnt/models/environment.tar.gz...
--> Sourcing new environment at ./envs/environment...
--> Calling conda-unpack...
--> Disabling user-installed packages...
```

Then, when I start mlserver using, then I see this in the logs:

```
2023/10/27 14:57:47 WARNING mlflow.pyfunc: Detected one or more mismatches between the model's dependencies and the current Python environment:
 - mlflow (current: 2.4.1, required: mlflow==2.5.0)
 - numpy (current: 1.24.3, required: numpy==1.24.4)
 - scikit-learn (current: 1.3.0, required: scikit-learn==1.2.2)
 - scipy (current: 1.10.1, required: scipy==1.9.3)
To fix the mismatches, call `mlflow.pyfunc.get_model_dependencies(model_uri)` to fetch the model's environment and install dependencies using the resulting environment file.
2023/10/27 14:57:47 WARNING mlflow.pyfunc: The version of Python that the model was saved in, `Python 3.11.2`, differs from the version of Python that is currently running, `Python 3.8.16`, and may be incompatible
```

This seems odd as the conda env should be correctly setup. 

Here is the MLModel attached to this particular model:

```
artifact_path: model
flavors:
  python_function:
    cloudpickle_version: 2.2.1
    code: code
    env:
      conda: conda.yaml
      virtualenv: python_env.yaml
    loader_module: mlflow.pyfunc.model
    python_model: python_model.pkl
    python_version: 3.11.2
mlflow_version: 2.5.0
model_uuid: 8f5c6ad749ad48f6ada69b82b2daa1c0
run_id: d6cef80c75c24b55ab90982a9cfe41b5
signature:
  inputs: '[{"type": "double", "name": "acc_rate"}, {"type": "long", "name": "avg_daily_trips"},
[...]

PR #1739, opened 2024-05-06
Spawn worker in custom environment

lhnwrk:
As per #1461, MLServer currently spawns worker with the same Python version as the main process, so user-provided environment is forced to match the Python version of the server. However, `multiprocessing` allows setting the worker's executable path, this PR extends support for spawning workers in custom environments.

### DIFF
--- a/docs/user-guide/custom.md
+++ b/docs/user-guide/custom.md
@@ -224,10 +224,12 @@ To load a custom environment, [parallel inference](./parallel-inference)
 ```
 
 ```{warning}
-When loading custom environments, MLServer will always use the same Python
-interpreter that is used to run the main process.
-In other words, all custom environments will use the same version of Python
-than the main MLServer process.
+The main MLServer process communicates with workers in custom environments via
+[`multiprocessing.Queue`](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Queue)
+using pickled objects. Custom environments therefore **must** use the same
+version of MLServer and a compatible version of Python with the same [default
+pickle protocol](https://docs.python.org/3/library/pickle.html#pickle.DEFAULT_PROTOCOL)
+as the main process.
 ```
 
 If we take the [previous example](#loading-a-custom-mlserver-runtime) above as
--- a/docs/user-guide/custom.md
+++ b/docs/user-guide/custom.md
@@ -229,8 +229,20 @@ The main MLServer process communicates with workers in custom environments via
 using pickled objects. Custom environments therefore **must** use the same
 version of MLServer and a compatible version of Python with the same [default
 pickle protocol](https://docs.python.org/3/library/pickle.html#pickle.DEFAULT_PROTOCOL)
-as the main process.
+as the main process. Consult the tables below for environment compatibility.
 ```
+
+| Status | Description  |
+| ------ | ------------ |
+| 🔴     | Unsupported  |
+| 🟢     | Supported    |
+| 🔵     | Untested     |
+
+| Worker Python \ Server Python | 3.9 | 3.10 | 3.11 |
+| ----------------------------- | --- | ---- | ---- |
+| 3.9                           | 🟢  | 🟢   | 🔵   |
+| 3.10                          | 🟢  | 🟢   | 🔵   |
+| 3.11                          | 🔵  | 🔵   | 🔵   |
 
 If we take the [previous example](#loading-a-custom-mlserver-runtime) above as
 a reference, we could extend it to include our custom environment as:
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -4,7 +4,6 @@ import shutil
 import asyncio
 import glob
 import json
-import sys
 
 from filelock import FileLock
 from typing import Dict, Any, Tuple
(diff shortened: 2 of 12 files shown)


sakoush:
@lhnwrk many thanks for your contribution to add the ability to run mlserver workers on different python versions. 

Internally mlserver communicates with these parallel workers using `multiprocessing.Queue` (check [here](https://github.com/SeldonIO/MLServer/blob/410cf6f838782443267c88bd410cabf95c760c54/mlserver/parallel/worker.py#L43-L44)) and therefore our implementation requires that the main process and the workers to use  the same python version (or at least compatible versions) in terms of communication and serialisation. We have not tested actually what versions of python can interplay. It is unlikely though that workers can be on python `2.x` for example.

Having said that, the change you propose is slightly orthogonal to the above point as there is an argument to be made that we should set `multiprocessing.set_executable` from the custom env anyway.

We suggest the following for us to be able to accept this change:
- Extend the unit tests so that workers can be created using different python versions to back up your change.
- Provide a list of cases as docs where this change can / cannot be applicable. 

We can provide more pointers if you are happy to address the above.

cc: @jesse-c, @lc525

lhnwrk:
AFAIK we only need the worker process to have the same pickle protocol as the main process for `multiprocessing.Queue` to pass serialized objects from one process to another. Unfortunately we don't get to control what protocol `multiprocessing` uses, its [`ForkingPickler`](https://github.com/python/cpython/blob/35c436186b849f8f2f9fb866c59015c9d034d448/Lib/multiprocessing/reduction.py#L33-L54) simply falls back to the default pickle protocol. Version 4 has been the default since python 3.8, so for now any later versions should be compatible. 
 
I've updated the [docs](https://github.com/lhnwrk/MLServer/blob/610cce80f731eafd6c20af615c473963124cd067/docs/user-guide/custom.md?plain=1#L226-L233) to point out the caveats for users, and extended the test suites to include custom environments with different python versions. For now they include the main process's python, the minimum tested python (3.9), and the maximum tested python (3.10) so we cover all cases of running a worker process with the same, lower, and higher python versions. 

Let me know if there's anything else!

tests/conftest.py
```diff
@@ -59,17 +63,43 @@ def testdata_cache_path() -> str:
...
 
 
+@pytest.fixture(
+    params=[
+        PYTHON_VERSION,
+        pytest.param(
+            MIN_PYTHON_VERSION,
+            marks=pytest.mark.skipif(
+                MIN_PYTHON_VERSION >= PYTHON_VERSION,
+                reason="requires lower Python version",
```
sakoush, line 73: I am not sure what does the reason mean here? for example if the current system python version is `3.8` then the parameter `(3, 9)` is not going to be used? and therefore the `reason` is probably misleading. Could you clarify please?

tests/conftest.py
```diff
@@ -59,17 +63,43 @@ def testdata_cache_path() -> str:
...
+            ),
+        ),
+        pytest.param(
+            MAX_PYTHON_VERSION,
+            marks=pytest.mark.skipif(
+                MAX_PYTHON_VERSION <= PYTHON_VERSION,
+                reason="requires higher Python version",
```
sakoush, line 80: this will skip testing mlserver main process on py 3.9 and worker on py 3.10. this is probably safer, just curious if you spotted any issues though with this combination?

docs/user-guide/custom.md
```diff
@@ -224,10 +224,12 @@ To load a custom environment, [parallel inference](./parallel-inference)
...
-When loading custom environments, MLServer will always use the same Python
-interpreter that is used to run the main process.
-In other words, all custom environments will use the same version of Python
-than the main MLServer process.
+The main MLServer process communicates with workers in custom environments via
+[`multiprocessing.Queue`](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Queue)
+using pickled objects. Custom environments therefore **must** use the same
+version of MLServer and a compatible version of Python with the same [default
+pickle protocol](https://docs.python.org/3/library/pickle.html#pickle.DEFAULT_PROTOCOL)
+as the main process.
```
sakoush, line 232: could you also add a note about the specific python versions that are currently supported. i.e. a table showcasing main process and parallel workers python versions?

tests/testdata/environment.yml
```diff
@@ -2,7 +2,7 @@ name: custom-runtime-environment
 channels:
   - conda-forge
 dependencies:
-  - python == 3.8
+  - python == 3.9
```
sakoush, line 5: good spot

sakoush (commented):
@lhnwrk many thanks for addressing the comments. I left minor clarification questions.

sakoush:
@lhnwrk could you rebase on top of master to make sure that tests are passing on this PR?

tests/conftest.py
lhnwrk: My original thought was to test three cases of a worker environment with the same, lower, or higher Python version than the main process, so `MIN_PYTHON_VERSION` is only tested when it's lower than the main Python for example. This is updated now to just test all python versions between `MIN_PYTHON_VERSION` and `MAX_PYTHON_VERSION`.

lhnwrk:
@sakoush Rebased on `master`!

docs/user-guide/custom.md
lhnwrk: Added a table here to clarify supported/tested worker environment!

tests/testdata/environment.yml
```diff
@@ -2,7 +2,7 @@ name: custom-runtime-environment
...
   - scikit-learn == 1.0.2
   - pip:
-      - mlserver == 1.3.0.dev2
+      - git+${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}.git@${GITHUB_REF}
```
lhnwrk, line 8: I switched to installing `mlserver` from git here since worker environment has to match main process for this PR. GitHub Actions should [set these environment variables to the fork/branch](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables) so the worker environment installs the same `mlserver`, but default to `SeldonIO/MLServer`'s `master` branch in `tox.ini` otherwise.

sakoush: @lhnwrk is it possible to install from the local mlserver directory so it is easier logic and also locally we might want to be testing from changes done locally?

sakoush (commented):
@lhnwrk many thanks for the changes, i did add a couple more minor comments but it looks great otherwise.

I will also restart the failing CI tests.

tests/testdata/environment.yml
lhnwrk: I did go with this option first, but as it turns out this breaks docker build tests since the local mlserver directory is not available inside the container and the template Dockerfile only copies environment file. It has been a hassle to test locally though, what do you think if we use a separate yaml with a pinned version of mlserver for the CLI build test and install local directory for others?

sakoush (approved):
lgtm! nice one @lhnwrk.

PR #1729, opened 2024-04-30
test: Detect generate failures

jesse-c:
Currently, since we don't exit on a command failure, it's been missed that generation is currently broken for the dataplane.

You can see an example in these logs from a previous workflow (https://github.com/SeldonIO/MLServer/actions/runs/8880776962/job/24381621013).

```
Writing mypy to dataplane_pb2.pyi
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.10.14/x64/lib/python3.10/site-packages/datamodel_code_generator/__main__.py", line 428, in main
    generate(
  File "/opt/hostedtoolcache/Python/3.10.14/x64/lib/python3.10/site-packages/datamodel_code_generator/__init__.py", line 462, in generate
    results = parser.parse()
  File "/opt/hostedtoolcache/Python/3.10.14/x64/lib/python3.10/site-packages/datamodel_code_generator/parser/base.py", line 1153, in parse
    self.parse_raw()
  File "/opt/hostedtoolcache/Python/3.10.14/x64/lib/python3.10/site-packages/datamodel_code_generator/parser/openapi.py", line 571, in parse_raw
    specification: Dict[str, Any] = load_yaml(source.text)
  File "/opt/hostedtoolcache/Python/3.10.14/x64/lib/python3.10/site-packages/datamodel_code_generator/__init__.py", line 51, in load_yaml
    return yaml.load(stream, Loader=SafeLoader)
  File "/opt/hostedtoolcache/Python/3.10.14/x64/lib/python3.10/site-packages/yaml/__init__.py", line 81, in load
    return loader.get_single_data()
  File "/opt/hostedtoolcache/Python/3.10.14/x64/lib/python3.10/site-packages/yaml/constructor.py", line 49, in get_single_data
    node = self.get_single_node()
  File "yaml/_yaml.pyx", line 673, in yaml._yaml.CParser.get_single_node
  File "yaml/_yaml.pyx", line 687, in yaml._yaml.CParser._compose_document
  File "yaml/_yaml.pyx", line 731, in yaml._yaml.CParser._compose_node
  File "yaml/_yaml.pyx", line 845, in yaml._yaml.CParser._compose_mapping_node
  File "yaml/_yaml.pyx", line 731, in yaml._yaml.CParser._compose_node
  File "yaml/_yaml.pyx", line 845, in yaml._yaml.CParser._compose_mapping_node
  File "yaml/_yaml.pyx", line 731, in yaml._yaml.CParser._compose_node
[...]

PR #1749, opened 2024-05-08
build: Add granularity in types generation

jesse-c:
This makes it easier to know which types generation is failing. It's a follow-on from https://github.com/SeldonIO/MLServer/pull/1729.


mauicv (approved):
LGTM

Issue #1625, opened 2024-03-06
Schema enforcement error when using `PandasCodec` inference request


mfueller:
Hi!

My goal is to serve a mlflow model with a signature via mlserver and observed some issue with the signature enforcement and the request generated by `PandasCodec`.

I followed the example from https://mlserver.readthedocs.io/en/latest/examples/mlflow/README.html

The model signature in the example is inferred by
```python
model_signature = infer_signature(train_x, train_y)
```
and logged to mlflow via:
```python
mlflow.sklearn.log_model(
                lr,
                "model",
                registered_model_name="ElasticnetWineModel",
                signature=model_signature,
            )
```

model serving is done with mlserver
```
mlserver start .
```

I can test the inference via the given plain json example
```python
import requests

inference_request = {
    "inputs": [
        {
          "name": "fixed acidity",
          "shape": [1],
          "datatype": "FP32",
          "data": [7.4],
        },
        {
          "name": "volatile acidity",
          "shape": [1],
          "datatype": "FP32",
          "data": [0.7000],
        },
       ...
    ]
}

endpoint = "http://localhost:8080/v2/models/wine-classifier/infer"
response = requests.post(endpoint, json=inference_request)

response.json()
```
```
{'model_name': 'ElasticnetWineModel',
 'id': 'c0dbba4c-ac18-43ba-a408-3cf9536931bd',
 'parameters': {'content_type': 'np'},
 'outputs': [{'name': 'output-1',
   'shape': [1, 1],
   'datatype': 'FP64',
   'parameters': {'content_type': 'np'},
   'data': [5.576883936610762]}]}
```

However, if I create the inference request using `PandasCodec` like this

```python
inference_request = PandasCodec.encode_request(test_x.head(1))
endpoint = "http://localhost:8080/v2/models/wine-classifier/infer"
response = requests.post(endpoint, json=inference_request.dict())
response.json()
```

I get the following error response:

```
[...]

PR #1751, opened 2024-05-09
Fix Pandas codec decoding from numpy arrays

lhnwrk:
Pandas codec currently breaks mlflow runtime's schema enforcement (#1625). This is because numeric mlflow datatypes are assigned `NumpyCodec` content type [by default](https://github.com/SeldonIO/MLServer/blob/410cf6f838782443267c88bd410cabf95c760c54/runtimes/mlflow/mlserver_mlflow/metadata.py#L24-L28), which decodes `PandasCodec`'s explicit batch size into a column vector:
```
input = RequestInput(
	name="foo",
	shape=[3, 1],
	data=[1,2,3],
	datatype="INT64",
)
NumpyCodec.decode_input(input)  # np.array([[1], [2], [3]])
```

However, `pd.Series` expects one-dimensional data, and when the `_decoded_payload` is converted to a `pd.Series` [here](https://github.com/SeldonIO/MLServer/blob/410cf6f838782443267c88bd410cabf95c760c54/mlserver/codecs/pandas.py#L28-L33) returns unexpected results:
```
# Expect pd.Series([1,2,3]), got pd.Series([(1,), (2,), (3,)])
payload = np.array([[1], [2], [3]])
payload = list(payload)
pd.Series(payload, dtype="int64")
```

This seems like a bug from pandas side, `dtype="int64"` should have thrown an error (and in fact it does for higher dimension, but for some reason is okay with `[np.array([1]), np.array([2]), np.array([3])]` and weirdly casts them into tuples), but either way mlserver should have dropped the trailing axis as `pd.Series` already implies column vector. 

From this [test](https://github.com/SeldonIO/MLServer/blob/410cf6f838782443267c88bd410cabf95c760c54/tests/codecs/test_pandas.py#L319-L339), it seems like the conversion to `list` in the codec was to accommodate the use case of `pd.Series` of tensors:
```
# Expect decoded pd.Series([np.array([1,2,3])]) in test
RequestInput(
	name="a",
	data=[1, 2, 3],
	datatype="FP32",
	shape=[1, 3],
	parameters=Parameters(_decoded_payload=np.array([[1, 2, 3]])),
)
```

I'd argue in the context of Pandas codec this should be made explicit with shape `[1, 1]` and datatype `BYTES` for array elements. In fact, a [test](https://github.com/SeldonIO/MLServer/blob/410cf6f838782443267c88bd410cabf95c760c54/tests/codecs/test_pandas.py#L66-L72) above does so, specifying the intended shape is a `[2,1]` column vector with datatype `BYTES` for `list` elements.
```
# Expect encoded ResponseOutput(name="bar", shape=[2, 1], data=[[1, 2, 3], [4, 5, 6]], datatype="BYTES") in test
pd.Series(data=[[1, 2, 3], [4, 5, 6]], name="bar")
```

This PR explicitly reshapes `np.array` payload to expected shapes before conversion to `pd.Series`.

### DIFF
--- a/mlserver/codecs/pandas.py
+++ b/mlserver/codecs/pandas.py
@@ -26,8 +26,7 @@ def _to_series(input_or_output: InputOrOutput) -> pd.Series:
         return pd.Series(payload)
 
     if isinstance(payload, np.ndarray):
-        # Necessary so that it's compatible with pd.Series
-        payload = list(payload)
+        payload = payload.reshape(-1, 1).squeeze(axis=1)
 
     dtype = to_dtype(input_or_output)
     return pd.Series(payload, dtype=dtype)
--- a/mlserver/codecs/pandas.py
+++ b/mlserver/codecs/pandas.py
@@ -26,7 +26,7 @@ def _to_series(input_or_output: InputOrOutput) -> pd.Series:
         return pd.Series(payload)
 
     if isinstance(payload, np.ndarray):
-        payload = payload.reshape(-1, 1).squeeze(axis=1)
+        payload = payload.reshape(input_or_output.shape[0], 1).squeeze(axis=1)
 
     dtype = to_dtype(input_or_output)
     return pd.Series(payload, dtype=dtype)
--- a/runtimes/mlflow/mlserver_mlflow/metadata.py
+++ b/runtimes/mlflow/mlserver_mlflow/metadata.py
@@ -46,7 +46,7 @@ def _get_shape(input_spec: InputSpec) -> List[int]:
     if isinstance(input_spec, TensorSpec):
         return list(input_spec.shape)
 
-    return [-1]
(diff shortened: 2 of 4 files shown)


jesse-c (approved):
LGTM

We're resolving some issues around testing and then I plan on getting this in.

jesse-c:
We've gotten that fix in. @lhnwrk: could you please rebase this branch on top of `master`?

lhnwrk:
@jesse-c Rebased on `master`!

jesse-c:
@lhnwrk: Thank you! I've had to re-approve it and run the checks, so they're going now. As soon as it passes, I'll merge it in.

jesse-c:
Merged in! Thank you again, @lhnwrk. 🙏🏻

PR #1752, opened 2024-05-10
Fixed openapi dataplane.yaml

RobertSamoilescu:
This PR fixes a syntax error in the `dataplane.yaml`.

PR #1759, opened 2024-05-15
Revert "Fixed openapi dataplane.yaml"

jesse-c:
Reverts SeldonIO/MLServer#1752

This has introduced some failures around a type becoming an object instead of a list.


jesse-c:
Superseded by https://github.com/SeldonIO/MLServer/pull/1760.

PR #1761, opened 2024-05-15
build: Bump versions for Read the Docs

jesse-c:
We don't support Python `3.8` anymore.

Poetry `1.8.1` matches what's currently in our Poetry lockfile.

Building the docs is currently failing [1], and this ideally will fix that.

[1] From readthedocs.org:
```
poetry install -vvv --without dev --with docs --with all-runtimes --with all-runtimes-dev
Loading configuration file /home/docs/.config/pypoetry/config.toml
The currently activated Python version 3.8.18 is not supported by the project (>=3.9,<3.12).
Trying to find and use a compatible version.
Trying python3
Trying python3.9
```

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -16,14 +16,14 @@ formats:
 build:
   os: ubuntu-22.04
   tools:
-    python: "3.8"
+    python: "3.10"
   jobs:
     # Set up Poetry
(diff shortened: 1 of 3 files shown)


sakoush (approved):
lgtm, thanks for fixing it.

PR #1785, opened 2024-05-30
 build: Install dependencies only in Tox environments 

jesse-c:
Follow-up to: https://github.com/SeldonIO/MLServer/pull/1765#discussion_r1618957160

Additionally, I've given the jobs more informative names.


mauicv (approved):
LGTM!

PR #1794, opened 2024-06-06
Fixes for all-runtimes

RobertSamoilescu:
This PR includes some minor fixes for the `all-runtimes` tests.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -167,11 +167,13 @@ jobs:
 
   # Ensure that having all the runtimes installed together works
   all-runtimes:
+    if: github.event_name == 'push'
     strategy:
       fail-fast: false
       matrix:
         os:
           - ubuntu-22.04
+          - macos-13
         python-version:
           - "3.9"
           - "3.10"
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -167,13 +167,11 @@ jobs:
 
   # Ensure that having all the runtimes installed together works
   all-runtimes:
-    if: github.event_name == 'push'
     strategy:
       fail-fast: false
       matrix:
         os:
           - ubuntu-22.04
-          - macos-13
         python-version:
           - "3.9"
           - "3.10"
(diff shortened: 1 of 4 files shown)


mlserver/env.py
```diff
@@ -117,6 +117,9 @@ def _lib_path(self) -> str:
         matches = glob.glob(pattern)
 
         for match in matches:
+            if match.endswith("3.1"):
```
sakoush, line 120: why is this required? add a note in the code.

mlserver/metrics/server.py
```diff
@@ -72,5 +73,8 @@ def _get_config(self):
         return uvicorn.Config(self._app, **kwargs)
 
     async def stop(self, sig: Optional[int] = None):
+        if sig is None:
```
sakoush, line 76: why is this required? add a note in the code

tox.ini
```diff
@@ -42,6 +42,10 @@ commands =
         {toxinidir}/runtimes/lightgbm \
         {toxinidir}/runtimes/mlflow \
         {toxinidir}/runtimes/huggingface
+set_env =
```
sakoush, line 45: ditto

sakoush (approved):
lgtm. please clarify in the code why these changes are added.

tox.ini
RobertSamoilescu: See the `mlserver` env which requires that for some env test. Feels a bit weird to add some comments there.

sakoush: oh I see, we missed this change from [here](https://github.com/SeldonIO/MLServer/blob/b848836872a215cc781075afdbcac345bea49ab6/tox.ini#L24) to be added to `all-runtimes`

.github/workflows/tests.yml
```diff
@@ -167,13 +167,11 @@ jobs:
 
   # Ensure that having all the runtimes installed together works
   all-runtimes:
-    if: github.event_name == 'push'
```
sakoush, line 170: assuming this is for testing, revert back before merging.

PR #1796, opened 2024-06-07
Removed macos from github workflows

RobertSamoilescu:
This PR removes macos from the github worflow.


sakoush (approved):
lgtm. Add notes in the description why we are not doing the macos tests anymore.
Also check in the docs if there is a reference that says we support macos, if so then adjust accordingly.

PR #1805, opened 2024-06-14
Fix CI/CD macos-huggingface

RobertSamoilescu:
This PR fixes `macos - huggingface` CI. `torch` no longer supports `macos x86` (see issue [here](https://github.com/pytorch/pytorch/issues/114602))


sakoush (approved):
lgtm

PR #1858, opened 2024-07-12
Included Prometheus interceptor support for gRPC streaming

RobertSamoilescu:
This PR includes Prometheus interceptor support for gRPC streaming. Currently for gRPC streaming, we have to set `"metrics_endpoint": null`, thus Prometheus logs cannot be scraped.  It also updates docs and test for Prometheus interceptor.

### DIFF
--- a/mlserver/grpc/interceptors.py
+++ b/mlserver/grpc/interceptors.py
@@ -1,9 +1,12 @@
-from typing import Awaitable, Callable, Tuple
+from typing import Awaitable, AsyncIterator, Callable, Tuple, Optional
 from functools import partial
 from timeit import default_timer
 
+from mlserver.grpc import dataplane_pb2 as pb
 from grpc.aio import ServerInterceptor, ServicerContext
 from grpc import HandlerCallDetails, RpcMethodHandler, RpcError, StatusCode
+
+from prometheus_client import Counter
 from py_grpc_prometheus.prometheus_server_interceptor import (
     grpc_utils,
     PromServerInterceptor as _PromServerInterceptor,
@@ -66,7 +69,9 @@ class PromServerInterceptor(ServerInterceptor):
         """
         grpc_service_name, grpc_method_name, _ = method_call
 
-        async def new_behavior(request, servicer_context):
+        async def new_behavior(
+            request: pb.ModelMetadataRequest, servicer_context: ServicerContext
+        ) -> Optional[pb.ModelMetadataRequest]:
             response = None
             try:
                 start = default_timer()
@@ -139,8 +144,9 @@ class PromServerInterceptor(ServerInterceptor):
                 raise e
 
         async def new_behavior_stream(
-            request_async_iterator, servicer_context: ServicerContext
-        ):
+            request_async_iterator: AsyncIterator[pb.ModelInferRequest],
+            servicer_context: ServicerContext,
+        ) -> AsyncIterator[pb.ModelInferRequest]:
             response_async_iterator = None
             try:
                 grpc_type = grpc_utils.get_method_type(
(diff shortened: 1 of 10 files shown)


tests/grpc/test_interceptor.py
```diff
@@ -0,0 +1,129 @@
+import pytest
```
sakoush, line 1: thanks for adding these tests. ideally we also should test other metrics / errors but can happen as a follow up PR.

tests/grpc/test_interceptor.py
```diff
@@ -0,0 +1,129 @@
...
+    inference_service_stub: AsyncGenerator[GRPCInferenceServiceStub, None],
+    model_generate_request: pb.ModelInferRequest,
+    model_name: str,
+):
+    model_generate_request.model_name = model_name
+
+    async def get_stream_request(request):
+        yield request
+
+    # send 10 requests
```
sakoush, line 78:
```suggestion
    # send 1 requests
```

mlserver/grpc/interceptors.py
```diff
@@ -192,9 +188,62 @@ async def _new_handler(request_or_iterator, servicer_context: ServicerContext):
...
+        if isinstance(code, int):
+            if code not in self._status_codes:
+                return StatusCode.UNKNOWN
+
+            return self._status_codes[code]
+
+        return code
+
+
+async def wrap_async_iterator_inc_counter(
```
sakoush, line 238: add type hints

mlserver/grpc/interceptors.py
```diff
@@ -99,60 +66,111 @@ def _metrics_wrapper(
...
-                        ]
-                        response_or_iterator = grpc_utils.wrap_iterator_inc_counter(
-                            response_or_iterator,
-                            sent_metric,
-                            grpc_type,
-                            grpc_service_name,
-                            grpc_method_name,
-                        )
+                    # wrap the original behavior with the metrics
+                    sent_metric = self._interceptor._metrics[
```
sakoush, line 159: nit: the above code uses `self._interceptor._metrics["grpc_server_stream_msg_received"]` directly, while the below code doesnt. Is this to get around line width? perhaps unify style?

tests/grpc/test_interceptor.py
```diff
@@ -0,0 +1,129 @@
...
+        model_generate_request
+    )
+
+    # we count the number of words because
+    # each word is gonna be streamed back
+    request_text = StringCodec.decode_input(inference_request_g.inputs[0])[0]
+    num_words = len(request_text.split())
+
+    assert int(counted_requests) == num_requests
+    assert int(counted_requests) * num_words == int(counted_responses)
```
sakoush, line 129: question: how are we actually counting the actual words in the responses?

sakoush (approved):
LGTM. I left some minor comments.

mlserver/grpc/server.py
```diff
@@ -38,14 +38,14 @@ def _create_server(self):
             self._model_repository_handlers
         )
 
-        interceptors = []
+        self._interceptors = []
```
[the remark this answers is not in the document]
RobertSamoilescu: Was not a bug. Just needed a way to access the interceptors list for testing.

tests/grpc/test_interceptor.py
RobertSamoilescu: That should be 10, but I forgot to update the `num_request` var.

tests/grpc/test_interceptor.py
RobertSamoilescu: The model used is a dummy text one which split words by white space. Each word is going to be streamed back by the model.

PR #1901, opened 2024-09-09
Fix tensorflow upperbound macos

RobertSamoilescu:
Limited `tensorflow`  upperbound to `2.17` for macos.
It seems that `2.17` is not compiled for `macos` (see package location [here](https://www.tensorflow.org/install/pip#package_location))

### DIFF
--- a/runtimes/huggingface/pyproject.toml
+++ b/runtimes/huggingface/pyproject.toml
@@ -23,8 +23,11 @@ torch = [
     {markers = "sys_platform == 'darwin' and platform_machine == 'x86_64'", version = "<2.3.0"},
     {markers = "sys_platform != 'darwin' or platform_machine != 'x86_64'", version = "<2.4.0"}
 ]
+tensorflow = [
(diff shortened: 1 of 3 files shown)


runtimes/huggingface/pyproject.toml
```diff
@@ -23,8 +23,13 @@ torch = [
     {markers = "sys_platform == 'darwin' and platform_machine == 'x86_64'", version = "<2.3.0"},
     {markers = "sys_platform != 'darwin' or platform_machine != 'x86_64'", version = "<2.4.0"}
 ]
+# tensorflow 2.17.0 not compiled for MacOS
+# see: https://www.tensorflow.org/install/pip#package_location
+tensorflow = [
+    {markers = "sys_platform == 'darwin' and platform_machine == 'x86_64'", version = ">=2.12,<2.17"},
+    {markers = "sys_platform != 'darwin' or platform_machine != 'x86_64'", version = ">=2.12,<2.18"}
```
sakoush, line 30: It is probably fine, although I wondered a bit if we should be explicitly about what we include (as opposed to exclude).

sakoush (approved):

PR #1918, opened 2024-10-03
GitBook Docs Migration

ramonpzg:
This PR contains the changes necessary to migrate our MLServer docs to GitBook. There is one missing component though and that is the API references, this can be added down the line though.


sakoush:
done as part of #1919

Issue #1660, opened 2024-04-03
Inference parameters are not passed to the predict function in MLFlow runtime


Okamille:
## How to reproduce

Define a MLFlow model, using a custom `params` parameter. Example from the [MLFlow documentation](https://mlflow.org/docs/latest/traditional-ml/creating-custom-pyfunc/notebooks/override-predict.html#Defining-our-custom-PythonModel):

```
class ModelWrapper(PythonModel):
    def __init__(self):
        self.model = None

    def load_context(self, context):
        from joblib import load

        self.model = load(context.artifacts["model_path"])

    def predict(self, context, model_input, params=None):
        params = params or {"predict_method": "predict"}
        predict_method = params.get("predict_method")

        if predict_method == "predict":
            return self.model.predict(model_input)
        elif predict_method == "predict_proba":
            return self.model.predict_proba(model_input)
        elif predict_method == "predict_log_proba":
            return self.model.predict_log_proba(model_input)
        else:
            raise ValueError(f"The prediction method '{predict_method}' is not supported.")
```

Log that model into MLFlow, and try to serve it using MLServer. Create an inference request using a specific parameter. The parameter value is not taken into account.

## Proposed fix

Currently, the predict function in the MLFlow runtime does not pass the `params` parameter in the prediction. We can add it, like it was done for the custom invocation endpoint.

PR #1921, opened 2024-10-08
Forward rest parameters to model

idlefella:
# Pull Request

Fixes SeldonIO/MLServer#1660

## Description

The parameters sent over REST to the mlserver are not forwarded to the predict function of the model. This PR fixes that issue.

## Changes Made
Properly forward the params to the predict function of the model.

## Related Issues
* #1660

## Screenshots (if applicable)
-

## Checklist

- [x] Code follows the project's style guidelines
- [x] All tests related to the changes pass successfully
- [ ] Documentation is updated (if necessary)
- [ ] Code is reviewed by at least one other team member
- [x] Any breaking changes are communicated and documented

## Additional Notes
-

### DIFF
--- a/runtimes/mlflow/mlserver_mlflow/runtime.py
+++ b/runtimes/mlflow/mlserver_mlflow/runtime.py
@@ -196,5 +196,8 @@ class MLflowRuntime(MLModel):
 
     async def predict(self, payload: InferenceRequest) -> InferenceResponse:
         decoded_payload = self.decode_request(payload)
-        model_output = self._model.predict(decoded_payload)
+        params = None
+        if payload.parameters and payload.parameters.model_extra:
+            params = payload.parameters.model_extra
+        model_output = self._model.predict(decoded_payload, params=params)
         return self.encode_response(model_output, default_codec=TensorDictCodec)
--- a/runtimes/mlflow/tests/test_runtime.py
+++ b/runtimes/mlflow/tests/test_runtime.py
@@ -4,7 +4,6 @@ import numpy as np
 import pandas as pd
 
 from typing import Any
-
 from mlserver.codecs import NumpyCodec, PandasCodec, StringCodec
 from mlserver.types import (
     InferenceRequest,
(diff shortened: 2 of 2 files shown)


runtimes/mlflow/mlserver_mlflow/runtime.py
```diff
@@ -196,5 +196,8 @@ def _sync_metadata(self) -> None:
 
     async def predict(self, payload: InferenceRequest) -> InferenceResponse:
         decoded_payload = self.decode_request(payload)
-        model_output = self._model.predict(decoded_payload)
+        params = None
+        if payload.parameters and payload.parameters.model_extra:
+            params = payload.parameters.model_extra
+        model_output = self._model.predict(decoded_payload, params=params)
```
sakoush, line 202: is there a test that can be added for this?

sakoush: @idlefella I think `params` is only supported for class-based pyfunc models and therefore this proposed changed will not work for function-based models.
I think we need to find a way to check the type of the model loaded and not pass `params` for function-based models.
check [here](https://mlflow.org/docs/latest/python_api/mlflow.pyfunc.html#function-based-model-vs-class-based-model)

idlefella: Hi @sakoush 
Thanks for the comment. Ok then I will try to write a test for `class-based` and `function-based` models and see how or if these parameters can be forwarded. I hope to find time in the next couple of days.
Thanks!

idlefella: Hi @sakoush, when the underlying model is a python function, then the function is wrapped into a `_FunctionPythonModel`:
https://github.com/mlflow/mlflow/blob/7d5dcf7e9b59c0ed8799024cae9c054634c616cd/mlflow/pyfunc/model.py#L142
When params are set, the class prints a warning. So this doesn't cause any issues.
I also added some unittests for the forwarded params.

sakoush (approved):
LGTM. Thanks for your contribution @idlefella!

Issue #1920, opened 2024-10-04
Strict Version Dependency of Pydantic


lemonhead94:
Hi there,

I was trying to build a custom Runtime for serving `MLflow Langchain Models` and I noticed that you guys have a strict dependency on pydantic version `2.7.1`. Is there any practical reason not to relax it a bit more, for example `pydantic = ">=2.7.1"` and `pydantic-settings = ">=2.2.1"`.

Reference: [pydantic = "2.7.1"](https://github.com/SeldonIO/MLServer/blob/master/pyproject.toml#L65)

sakoush:
Hi @lemonhead94 yes relaxing pydantic bounds is ok, we welcome contributions if you are willing to submit a PR.

PR #1928, opened 2024-10-19
Relaxing Pydantic dependencies

lemonhead94:
This should allow for having a custom Langchain Runtime in MLServer [Issue 1920](https://github.com/SeldonIO/MLServer/issues/1920).


pyproject.toml
```diff
@@ -62,8 +62,8 @@ gevent = "*"
 aiofiles = "*"
 orjson = "*"
 uvloop = {version = "*", markers = "sys_platform != 'win32' and (sys_platform != 'cygwin' and platform_python_implementation != 'PyPy')"}
-pydantic = "2.7.1"
-pydantic-settings = "2.2.1"
+pydantic = "^2.7.1"
```
sakoush, line 65: Just to note that I dont think there is anything specific to pydantic ver 2.7.1 in the codebase, I think that we support pydantic v2 in general and probably this specific version was pinned at the time of testing an upgrade. It is fine to keep it as such though I think just to be on the safe side.

sakoush (approved):

pyproject.toml
lemonhead94: Can you rerun the xgboost test, it seems to be fluke

sakoush: sure, they pass now.

sakoush:
@lemonhead94 thanks for your contribution, there are related changes to move to ver 2.9 for pydantic done in https://github.com/SeldonIO/MLServer/pull/1934

PR #1934, opened 2024-10-21
build(deps): Upgrade fastapi and starlette

sakoush:
This change upgrade fastapi to `0.115.2` and starlette to `0.40.0`. As part of the upgrade other packages are also required to be upgraded because some transitive dependencies clashes.


RobertSamoilescu (approved):
LGTM

PR #2006, opened 2024-12-24
Add token metrics: model_input_tokens and model_output_tokens

luohua13:


RobertSamoilescu:
Hi @luohua13, 

Thank you so much for opening this PR. It is indeed a useful metric to expose when working with LLMs. What I would like to suggest to abstract away the use of Prometheus and use the custom metrics as presented in the docs [here](https://mlserver.readthedocs.io/en/stable/user-guide/metrics.html#custom-metrics). The only issue with the custom metrics approach is that MLServer currently supports only histogram type of metrics as seen [here](https://github.com/SeldonIO/MLServer/blob/master/mlserver/metrics/context.py).  If you think that a histogram might not be appropriate, it would be great if you consider adding support for `Counters` as well.

Other things to consider, although those might be already addressed, are:
* ensure that this PR does not affect or generalizes across other pipeline tasks (e.g., classification, qa, etc)
* padding - when performing batch inference, some inputs might be padded (how do we count the tokens in this case)

Please also add a brief description of what this PR is trying to solve, what is your approach, etc.

Thank you again for your contribution!

PR #2040, opened 2025-02-03
Included separate inference pool

RobertSamoilescu:
Currently MLServer can't deal well with models that are loaded on the same inference pool with other models that are heavy used. In this case there is a risk of starvation and we want to allow the user to be able to create models on separate processes (different inference pool).

MLServer does this but only if the model uses a specific custom environment tarball.

**Proposed solution**:  Introduce `inference_pool_gid` to the `ModelParameters` class. In this way, we give the user the option to add the model to a dedicated inference pool group. Note that this allows to add either a single model or multiple models to the same group. I also included the `autogenerate_inference_pool_gid` boolean flag. When set to `True` and no `inference_pool_gid` is provided, a gid is generated using `uuid4`. This will add the model to a dedicated group - this way the user does not have to specify the gid themself.

### DIFF
--- a/mlserver/parallel/registry.py
+++ b/mlserver/parallel/registry.py
@@ -38,6 +38,14 @@ def _get_env_tarball(model: MLModel) -> Optional[str]:
     return to_absolute_path(model_settings, env_tarball)
 
 
+def _get_environment_hash_gid(
+    env_hash: str, inference_pool_gid: Optional[str] = None
+) -> str:
+    if inference_pool_gid:
+        return f"{env_hash}-{inference_pool_gid}"
+    return env_hash
+
+
 class InferencePoolRegistry:
     """
     Keeps track of the different inference pools loaded in the server.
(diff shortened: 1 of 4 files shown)


mlserver/parallel/registry.py
```diff
@@ -145,8 +174,18 @@ def _get_env_path(self, env_hash: str) -> str:
...
         env_hash = _get_environment_hash(model)
+        inference_pool_gid = (
+            model.settings.parameters.inference_pool_gid
+            if model.settings.parameters
+            else None
+        )
+
         if not env_hash:
-            return self._default_pool
+            return (
```
sakoush, line 184: nit (style): I think it might be more legible to have if condition and multiple return as opposed to inline if-return.

mlserver/settings.py
```diff
@@ -313,6 +316,12 @@ class ModelParameters(BaseSettings):
     """Path to the environment tarball which should be used to load this
     model."""
 
+    inference_pool_gid: Optional[str] = None
```
sakoush, line 319: I think that docs should follow this PR.

sakoush (approved):
LGTM. Docs should follow this PR to describe the new fields.

mlserver/settings.py
RobertSamoilescu: I will do another PR to update the docs.

Issue #2112, opened 2025-04-16
PandasCodec behavior regression with 1.7.0


ReveStobinson:
Since the v1.7.0 update, `PandasCodec.encode_request()` is no longer able to encode columns that contain bytestring data. This issue was not present in v1.6.1.

## Steps to reproduce

The following code:

```python
import pandas as pd
from mlserver.codecs import PandasCodec

test_data = {
    "timestamp": 12,
    "uuid": "test",
    "data": bytes("testing", encoding="utf-8"),
}

payload = pd.DataFrame(test_data, index=[0])

request = PandasCodec.encode_request(payload)
```

Will result in this error:

```
Traceback (most recent call last):
  File "/home/username/file.py", line 12, in <module>
    request = PandasCodec.encode_request(payload)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/var/opt/miniforge/envs/mlserver/lib/python3.11/site-packages/mlserver/codecs/pandas.py", line 150, in encode_request
    outputs = cls.encode_outputs(payload, use_bytes=use_bytes)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/var/opt/miniforge/envs/mlserver/lib/python3.11/site-packages/mlserver/codecs/pandas.py", line 142, in encode_outputs
    return [
           ^
  File "/var/opt/miniforge/envs/mlserver/lib/python3.11/site-packages/mlserver/codecs/pandas.py", line 143, in <listcomp>
    _to_response_output(payload[col], use_bytes=use_bytes) for col in payload
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/var/opt/miniforge/envs/mlserver/lib/python3.11/site-packages/mlserver/codecs/pandas.py", line 54, in _to_response_output
    data = [encode_to_json(elem, use_bytes) for elem in data]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/var/opt/miniforge/envs/mlserver/lib/python3.11/site-packages/mlserver/codecs/pandas.py", line 54, in <listcomp>
    data = [encode_to_json(elem, use_bytes) for elem in data]
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #2117, opened 2025-04-21
Fixed byte encoding in the PandasCodec

RobertSamoilescu:
This PR fixes the byte encoding issue in the PandasCodec. Addresses #2112.


lc525 (approved):
lgtm

Joelius300:
Any estimate when this will be released (1.7.1)? We cannot upgrade without this fix.

PR #2178, opened 2025-06-04
ci: Merge change for release 1.7.1

RobertSamoilescu:


lc525 (approved):
lgtm; would you also be able to look at the security scans failing? perhaps we can improve things a bit wrt critical vulnerabilities

